PR #1, opened 2024-03-29
Add dotfiles and initial repository configuration

Krusty93:
### List of changes

Add dotfiles such as editorconfig, gitignore, pre-commit hook

### Motivation and context

Base configuration for any pagopa repo

### Type of changes

- [X] Add new resources

- [X] No


.pre-commit-config.yaml
```diff
@@ -0,0 +1,10 @@
+repos:
+  - repo: https://github.com/antonbabenko/pre-commit-terraform
```
gunzip, line 2: is this a reliable source?

PR #2, opened 2024-04-02
[DEVEX-50] Implement identity configuration with a new simplified module

Krusty93:
### List of changes

Bring changes of: https://github.com/pagopa/dx-typescript/pull/7

### Motivation and context

A new simplified module creates managed identities on Azure and federates them with GitHub repositories.
This module is available to (and should be used by) any other PagoPA repository.

Of course, this repository identity configuration uses this module as well.

### Type of changes

- [X] Add new resources

- [X] No


infra/github_federated_identity/prod/westeurope/locals.tf
```diff
@@ -0,0 +1,18 @@
+locals {
```
gunzip, line 1: isn't this repo supposed to contain only modules?

Krusty93: It's true, but it might need to log in to Azure for some automation process.
However, ff preferred we can defer the identity creation to the actual need

gunzip: it's ok

PR #4, opened 2024-04-15
[DEVEX-101] Update @pagopa/eslint-config

lucacavallaro:
### List of changes
1. add `code-review.yaml` pipeline
2. configure yarn 4
3. add `@pagopa/eslint-config` workspace that exports an `eslint flat configuration`

### ESlint plugins configured

* `@eslint/js` - default JS rules from `eslint`
* [typescript-eslint](https://typescript-eslint.io/) - `strict` and `stylistic` rules for TypeScript
* [eslint-plugin-vitest](https://github.com/veritem/eslint-plugin-vitest) - common best practices for `vitest`
* [eslint-plugin-perfectionist](https://github.com/azat-io/eslint-plugin-perfectionist) - non-blocking, auto-fixable ordering rules (`import`, `classes`, `objects`)
* `eslint-plugin-prettier` to avoid conflicts with prettier formatting


turbo.json
```diff
@@ -0,0 +1,17 @@
...
+      "inputs": ["src/**/*.ts"],
+      "outputs": ["dist/**"]
+    },
+    "typecheck": {
+      "inputs": ["src/**/*.ts"]
+    },
+    "lint:check": {
+      "inputs": ["src/**/*.ts"]
+    },
+    "format:check": {}
```
gunzip, line 15: why this doesn't have inputs?

lucacavallaro: Because the input for prettier is "all files"

gunzip:
What about having two sets? ie. lite vs nazi (we can settle on "lite" for this pr)

packages/eslint-config/index.js
```diff
@@ -0,0 +1,26 @@
...
+export default [
+  js.configs.recommended,
+  ...ts.configs.strict,
+  ...ts.configs.stylistic,
+  prettier,
+  perfectionist,
+  comments.configs.recommended,
+  {
+    ...vitest.configs.recommended,
+    rules: {
```
BurnedMarshal, line 17: Evaluate some rule change not included into the default reccomendend or strict configurations like:
- `"max-lines-per-function": ["error", 200]`
-  `"@typescript-eslint/no-unused-expressions": ["error"]`
-  `"@typescript-eslint/no-unused-vars": ["error", { argsIgnorePattern: "^_" }]`
- `"arrow-body-style": "error"`
- `complexity: "error"` whit default 20 value or a little bit more
- `"guard-for-in": "error"` Less useful if class are forbidden
- `eqeqeq: ["error", "smart"]`
- `"no-bitwise": "error"`
- `"no-eval": "error"`
-  `"no-console": "error"`  A logger must be used instead of standard console.log
- `"no-new-wrappers": "error"`
-  `"no-undef-init": "error"` Stylish but less usefull when immutability will be used
- `"no-param-reassign": "error"`
- `"no-var": "error"`
- `radix: "error"`
- `"prefer-const": "error"`

Other stilystic rules can be added using the plugin [eslint-stylistic](https://github.com/eslint-community/eslint-stylistic) when it will support ESLint 9, formatting rules will be deprecated in `typescript-eslint` like `member-delimiter-style` .

BurnedMarshal:
> What about having two sets? ie. lite vs nazi (we can settle on "lite" for this pr)

I agree on having multiple base rulset, but some principles must be guarantees even in lite configuration.

gunzip:
I'd like to have only fixable items in lite. Which principles are you referring to @BurnedMarshal ?

BurnedMarshal:
> I'd like to have only auto-fixable items in lite. Which principles are you referring to @BurnedMarshal ?

Have **only** auto-fixable rules is impossible, because even the raccomended ts-eslint and raccomended eslint rules have rules that don't support autofix or have reported errors on autofix feature.
Some of the principles that I would like to see enforced are:
- functional programming
- immutability (reducing side effects)
- complex functions obtained composing simple functions
- reduce dangerous JS patterns

gunzip:
I've doubts about enforcing functional programming everywhere. Think about a frontend application or infrastructure layers (sdk) where it's not mandatory, neither strongly suggested, to adopt `fp-ts` or something alike.

BurnedMarshal:
> I've doubts about enforcing functional programming everywhere. Think about a frontend application or infrastructure layers (sdk) where it's not mandatory, neither strongly suggested, to adopt `fp-ts` or something alike.

I agree, functional programming enforcing can be enabled on "nazy" ruleset or specific rule with functional paradigm.

lucacavallaro:
> > I've doubts about enforcing functional programming everywhere. Think about a frontend application or infrastructure layers (sdk) where it's not mandatory, neither strongly suggested, to adopt `fp-ts` or something alike.
> 
> 
> 
> I agree, functional programming enforcing can be enabled on "nazy" ruleset or specific rule with functional paradigm.

I think that functional programming and immutability should be enforced at project-level, for specific workspace/folders. 

With this common rules (that will be used by the whole PagoPA org) we should enforce writing good and modern TypeScript code.

packages/eslint-config/index.js
lucacavallaro: > * `"max-lines-per-function": ["error", 200]`

what is the rationale behind `200` ? imho we can restrict to 50 or 100

> * "@typescript-eslint/no-unused-vars": ["error", { argsIgnorePattern: "^_" }]

I would remove `argsIgnorePattern` option, and set `args: "after-used"` as it explained [here](https://eslint.org/docs/latest/rules/no-unused-vars#args). In this way we can write 

```js
app.get((req, res) => {
  res.json("hello!");
}
```

instead the ugly 

```js
app.get((_req, res) => {
  res.json("hello!");
}
```

without getting any warning. If also `res` is not used, the linter will raise an `error` as expected. 
Are you ok with this?

> * "arrow-body-style": "error"

I don't understand what you want to achieve with this rule. Would you disable arrow function with `() => {}`  shape or only when it is not needed?

> * eqeqeq: ["error", "smart"]

Why you prefer `smart` over `always`?

>

BurnedMarshal: > what is the rationale behind 200 ? imho we can restrict to 50 or 100
In my experience this rule can be hard to follow when the max line number is small. 50 or 100 lines per function produce a good quality and readable code, 200 avoid to write almost unreadable functions. Any team can force a smaller value at project level.

> I would remove argsIgnorePattern option, and set args: "after-used" 
I'll prefer the more strict version to avoid dangerous patter when the parameters share some prop o methods. Is more evident an error when the developer use the wrong argument.

```javascript
(function(foo, bar) {
   // Both foo and bar have a do method
    return bar.do();
})();
```
> - arrow-body-style": "error"

"as-neded". Avoid the `return` when you have only one instruction
```javascript
// bad
const foo = () => {
    return 0;
};
// ok
let foo = () => 0;

// bad
let bar = () => {
    return {
       bar: {
            foo: 1,
            bar: 2,
        }
    };
};
// ok
let bar = () => ({
       bar: {
            foo: 1,
            bar: 2,
        }
    });
```
> - eqeqeq
Why you prefer smart over always?

Is only a less more strict, this rules protect from bad comparison, smart is enough to reach that goal. [ref](https://eslint.org/docs/latest/rules/eqeqeq#smart)

BurnedMarshal:
> > > I've doubts about enforcing functional programming everywhere. Think about a frontend application or infrastructure layers (sdk) where it's not mandatory, neither strongly suggested, to adopt `fp-ts` or something alike.
> > 
> > 
> > I agree, functional programming enforcing can be enabled on "nazy" ruleset or specific rule with functional paradigm.
> 
> I think that functional programming and immutability should be enforced at project-level, for specific workspace/folders.
> 
> With this common rules (that will be used by the whole PagoPA org) we should enforce writing good and modern TypeScript code.
> 
> Functional programming is not "nazi", it's just a different paradigm that can be used in some workspaces.

The functional paradigm is the suggested way for Azure Functions and App Service template in the `dx-typescript` monorepo, and will be provided as template. Do you mean that linter rules that enforce functional constraint must be defined inside the `dx-typescript`? I'll provide a ruleset specific in here anyway:
- recommended
- functional

packages/eslint-config/index.js
BurnedMarshal: style vs safe

Should be enough the highlighting provided by the IDE, it's a good compromise 👍

gunzip (approved):
do this requires more work?

PR #12, opened 2024-05-07
[DEVEX-50] Fix azure_federated_identity_with_github module

Krusty93:
### List of changes

Remove output precondition
Made roles optional

### Motivation and context

Client couldn't create one identity only (either ci or cd) because of the output precondition
If an identity is disabled, is no longer necessary to use `null` as value for `roles`

### Type of changes

- [ ] Add new resources

- [ ] No


balanza:
What's the scenario of an identity to be created if CI is not enabled?

Krusty93:
> What's the scenario of an identity to be created and CI is not enabled?

In CI, often you only need to run linter, tests, a build, etc. without interacting with Azure
Usually, you only need to interact with Azure to perform `terraform plan` in CIs

gunzip (approved):

PR #13, opened 2024-05-07
[DEVEX-117] Add js_code_review reusable workflow

lucacavallaro:
### List of changes
1. Add a new github reusable workflow: `js_code_review` that runs the `code-review` script in javascript monorepo. This workflow is configured to cache `turbo` output and sends (only when is enabled) test coverage reports to https://about.codecov.io/

### How to use this workflow

Create a file named `code-review.yaml` in `.github/workflows` with the following content

```yaml
name: Code Review

on:
  workflow_dispatch:
  pull_request:
    types: [opened, synchronize]
    paths:
      - apps/**
      - packages/**

jobs:
  js_code_review:
    uses: pagopa/dx/.github/workflows/js_code_review.yaml@c461016f071d813b2012c88674df41c753f56b16
    name: Code Review
    secrets: inherit
```

### DIFF
--- a/.github/workflows/js_code_review.yaml
+++ b/.github/workflows/js_code_review.yaml
@@ -1,0 +1,51 @@
+name: "Code Review"
+
+on:
+  workflow_call: {}
+
+jobs:
+  js_code_review:
+    runs-on: ubuntu-22.04
+
+    steps:
+      - name: Check out code
+        uses: actions/checkout@0ad4b8fadaa221de15dcec353f45205ec38ea70b # v4.1.4
+
+      # Corepack is an official tool by Node.js that manages package managers versions
+      - name: Setup yarn
+        run: corepack enable
+
+      - name: Setup Node.js
+        uses: actions/setup-node@60edb5dd545a775178f52524783378180af0d1f8 # v4.0.2
+        with:
+          node-version-file: ".node-version"
+          cache: "yarn"
+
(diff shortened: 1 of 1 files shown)


gunzip (approved):

.github/workflows/js_code_review.yaml
```diff
@@ -0,0 +1,59 @@
...
+
+      # Codecov provides reports and metrics about test coverage data.
+      # To enable set CODECOV_TOKEN secret at repo level and make sure
+      # that your "code-review" script creates a "coverage/" folder in
+      # the root workspace.
+      - name: Check if CODECOV_TOKEN is set
+        id: codecov
+        shell: bash
+        run: |
+          if [ "${{ secrets.CODECOV_TOKEN }}" != '' ]; then
```
Krusty93, line 49: You could santize the secret value with:
```yaml
env:
  CODECOV_TOKEN: ${{ secrets.CODECOV_TOKEN }}
run: |
  if [ "${{ env.CODECOV_TOKEN }}" != '' ]; then
```

lucacavallaro: I saw it in some stackoverflow threads, but it's safe? I don't know how Github Workflow handles this step-scoped environments.

If it seems safe to you I will refactor it to remove that extra step!

PR #14, opened 2024-05-09
[DEVEX-122] Add Function App deploy workflow template for Node workloads

Krusty93:
### List of changes

Add a workflow template to automate Function App deployments for Node workloads

### Motivation and context

Make easier the deployment of Function App

### Type of changes

- [X] Add new resources

- [X] No

### DIFF
--- a/.github/workflows/function_app_deploy.yaml
+++ b/.github/workflows/function_app_deploy.yaml
@@ -1,0 +1,147 @@
+on:
+  workflow_call:
+    inputs:
+      workspace_name:
+        type: string
+        required: true
+        description: The name of the workspace to create the artifact for.
+      environment:
+        description: Environment where the artifact will be deployed.
+        type: string
+        required: true
+      resource_group_name:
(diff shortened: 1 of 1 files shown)


.github/workflows/function_app_deploy.yaml
```diff
@@ -0,0 +1,147 @@
...
+
+      - name: Build
+        run: yarn build
+        working-directory: ./out
+
+      - name: Build the Function App Artifact
+        id: make-function-app-artifact
+        run: |
+          npm pkg set --json "bundledDependencies"=true
+          npm pkg set --json "files"='["**/function.json", "dist", "host.json","extensions.csproj"]'
```
gunzip, line 80: do we need to copy `package.json` as well? @lucacavallaro

lucacavallaro: `package.json` is always included by default :)

gunzip (approved):

PR #15, opened 2024-05-14
[DEVEX-121] New TF module to create an Azure Function App

Krusty93:
### List of changes

A new module to create an Azure Function App.

This module creates:
- subnet for the function app
- storage account
- production slot
- staging slot
- private endpoints
- required rbac roles

This module DOES NOT use vnet routing rules, but instead uses private endpoints for both connecting to the storage account and regulating access to the function app. Access is actually disabled from outside the vnet (without the vpn).
Also, the Function App connects to the Storage Account using the managed identity rather than the connection string.

Naming convention is also forced

### Motivation and context

Actual module is too complex and does not force to use certain patterns like private endpoints

### Type of changes

- [X] Add new resources

- [X] No


infra/modules/azure_function_app/variables.tf
```diff
@@ -0,0 +1,86 @@
...
+  type      = string
+  sensitive = true
+}
+
+variable "health_check_path" {
+  type = string
+}
+
+variable "node_version" {
+  type = number
```
gunzip, line 56: can we force an enum? moreover, I'm wondering if this module should be called 

`azure_function_app_node` since a function supports generic runtimes (ie. java). for example self-care uses java functions. if the moving parts (variables to setup rintime) aren't that much, we may consider to collapse both into one module.

infra/modules/azure_function_app/variables.tf
```diff
@@ -0,0 +1,86 @@
...
+  description = "(Optional) A list of app_setting names that the Linux Function App will not swap between Slots when a swap operation is triggered"
+  default     = []
+}
+
+variable "subnet_cidr" {
+  type = string
+}
+
+variable "subnet_pep_id" {
+  type = string
```
gunzip, line 78: what's this?

infra/modules/azure_function_app/locals.tf
```diff
@@ -0,0 +1,6 @@
+locals {
+  project = "${var.prefix}-${var.env_short}-${var.location_short}"
+
+  sku_name               = var.env_short == "p" ? "P1v3" : "B1"
```
gunzip, line 4: consider that the development sku should have the same networking capabilities of the production one otherwise it might be unfeasible to connect the function with other services.

infra/modules/azure_function_app/function_app.tf
```diff
@@ -0,0 +1,177 @@
...
+  https_only                    = true
+  public_network_access_enabled = false
+  virtual_network_subnet_id     = azurerm_subnet.this.id
+
+  identity {
+    type = "SystemAssigned"
+  }
+
+  site_config {
+    http2_enabled                          = true
```
gunzip, line 31: are we sure we want to enable http2?

infra/modules/azure_function_app/function_app.tf
```diff
@@ -0,0 +1,177 @@
...
```
gunzip, line 101: ditto

infra/modules/azure_function_app/networking.tf
```diff
@@ -0,0 +1,105 @@
...
+
+    service_delegation {
+      name    = "Microsoft.Web/serverFarms"
+      actions = ["Microsoft.Network/virtualNetworks/subnets/action"]
+    }
+  }
+
+  service_endpoints = [
+    "Microsoft.Web",
+    # "Microsoft.Storage"
```
gunzip, line 18: why it's commented out?

infra/modules/azure_function_app/networking.tf
```diff
@@ -0,0 +1,105 @@
...
+
+  private_dns_zone_group {
+    name                 = "private-dns-zone-group"
+    private_dns_zone_ids = [data.azurerm_private_dns_zone.storage_account_file.id]
+  }
+
+  tags = var.tags
+}
+
+resource "azurerm_private_endpoint" "staging_function_sites" {
```
gunzip, line 85: just notice that there's an hard limit of 64 private endpoints for subscription

infra/modules/azure_function_app/variables.tf
Krusty93: yes we can, although enums in HCL are quite awkward

it's a question I asked to myself as well. We should add something like [this](https://github.com/pagopa/terraform-azurerm-v3/blob/c5e3aade01153d9dbd0bfb4f32bc6834fbded3c2/function_app/main.tf#L257), including node, java and docker only

infra/modules/azure_function_app/variables.tf
Krusty93: the id of the subnet in charge of hosting private endpoints

infra/modules/azure_function_app/locals.tf
Krusty93: Yesterday I created the resource with the "dev" sku and I didn't notice any issue. I'm still waiting to deploy an application on that to ensure everything run smoothly

infra/modules/azure_function_app/function_app.tf
Krusty93: Actual module does that

infra/modules/azure_function_app/function_app.tf
Krusty93: As above

infra/modules/azure_function_app/networking.tf
Krusty93: I'm not sure we need it

infra/modules/azure_function_app/outputs.tf
```diff
@@ -0,0 +1,30 @@
...
+
+output "storage_account" {
+  value = {
+    id   = azurerm_storage_account.this.id
+    name = azurerm_storage_account.this.name
+  }
+}
+
+output "function_app" {
+  value = {
```
christian-calabrese, line 16: If you agree, I would add the id of the function app's identity. With this it would be easy to add additional role assignments in the root module

infra/modules/azure_function_app/variables.tf
lucacavallaro: I would remove this variable, to me we should support only one node version: the latest LTS available (currently 20.x)

Krusty93: @lucacavallaro I'm afraid this approach would be a little too tight. What if you need to make a change and suddenly the module support the new version only? We should use tags to version the module, but I'd avoid that approach as long as we can (see Slack)

infra/modules/azure_function_app/outputs.tf
Krusty93: Nice catch

infra/modules/azure_function_app/variables.tf
gunzip: > it's a question I asked myself as well. We should add something like this, including node, java and docker only

If i remember well there were some issues using docker on app services @lucacavallaro

infra/modules/azure_function_app/locals.tf
Krusty93: I realized the staging slot is not supported out of premium plans. Is it an issue for dev resources? @lucacavallaro

infra/modules/azure_function_app/networking.tf
Krusty93: Not needed

infra/modules/azure_function_app/locals.tf
Krusty93: As discussed, we plan to introduce a variable where combination among sku scope and slots are tighten together and exposed as enum to consumer

infra/modules/azure_function_app/variables.tf
Krusty93: As discussed, Docker had issue with lifecycle and log management. We are going to include node and java only

infra/modules/azure_function_app/main.tf
```diff
@@ -0,0 +1,14 @@
+terraform {
+  required_providers {
+    azurerm = {
+      source  = "hashicorp/azurerm"
+      version = ">= 3.103.1"
```
christian-calabrese, line 5:
```suggestion
      version = "~> 3.103.1"
```
Do you think it may be beneficiary to employ the same kind of constraint used in azurermv3?

christian-calabrese, line 5: Is there any feature that forces to have the azurerm provider constrained to a very recent version?

infra/modules/azure_function_app/networking.tf
christian-calabrese, line 1: Are all the private endpoints necessary? If not all of them are and you agree, it would be nice to make them "flaggable" to reduce costs

Krusty93: We took the decision to remove both table and file

infra/modules/azure_function_app/main.tf
Krusty93: Accorded

infra/modules/azure_function_app/function_app.tf
```diff
@@ -0,0 +1,87 @@
...
+
+  tags = var.tags
+}
+
+resource "azurerm_linux_function_app" "this" {
+  name                = "${local.project}-${var.domain}-${var.app_name}-func-${var.instance_number}"
+  location            = var.location
+  resource_group_name = var.resource_group_name
+
+  service_plan_id = azurerm_service_plan.this.id
```
lucacavallaro, line 17: we should enable module consumers to optionally specify a different `service_plan_id` to allow `asp` reuse

lucacavallaro: (should we split `asp` module and `function_app` module?)

Krusty93: Changes done

infra/modules/azure_function_app/function_app.tf
```diff
@@ -1,20 +1,9 @@
...
-  tags = var.tags
-}
-
 resource "azurerm_linux_function_app" "this" {
   name                = "${local.project}-${var.domain}-${var.app_name}-func-${var.instance_number}"
   location            = var.location
   resource_group_name = var.resource_group_name
 
-  service_plan_id = azurerm_service_plan.this.id
+  service_plan_id = local.app_service_plan.enable ? azurerm_service_plan.this[0].id : var.app_service_plan_id
```
lucacavallaro, line 6: why `this[0]`?

Krusty93: The App Service Plan is now an array of string due to the `count` property, useful to implement deployment conditions (a fancy HCL thing)

infra/modules/azure_function_app/variables.tf
```diff
@@ -0,0 +1,148 @@
...
+  description = "Resource group to deploy resources to"
+}
+
+variable "app_service_plan_id" {
+  type        = string
+  default     = null
+  description = "(Optional) Set the AppService Id where you want to host the Function App"
+}
+
+variable "application_insights_connection_string" {
```
gunzip, line 59: should be optional in case you want to use only the sdk integration

Krusty93: Updated

christian-calabrese (approved):
LGTM

lucacavallaro (approved):
lgtm

PR #19, opened 2024-05-26
[DEVEX-130] Add documentation static website initial structure

gunzip:
### List of changes

Added initial structure for Docusaurus GithHub pages.

### Motivation and context

This will let dx team to document tools
and build a website with the documentation.

### Type of changes

- [x] Add new resources

- [X] No


.github/workflows/docs_website_test.yaml
```diff
@@ -0,0 +1,34 @@
+name: Test deployment
```
Krusty93, line 1:
```suggestion
name: Validate GitHub Pages Website
```

.github/workflows/docs_website_publish.yaml
```diff
@@ -0,0 +1,59 @@
...
+
+jobs:
+  build:
+    name: Build Docusaurus
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+        with:
+          fetch-depth: 0
+      - uses: actions/setup-node@v4
```
lucacavallaro, line 25: since we are using `packageManager` property on `package.json` we have to enable corepack before `actions/setup-node`

see https://github.com/pagopa/dx/blob/main/.github/workflows/js_code_review.yaml#L14

.github/workflows/docs_website_publish.yaml
```diff
@@ -0,0 +1,59 @@
...
+defaults:
+  run:
+    working-directory: ./website
+
+jobs:
+  build:
+    name: Build Docusaurus
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
```
lucacavallaro, line 22:
```suggestion
      - uses: actions/checkout@a5ac7e51b41094c92402da3b24376905380afc29 # v4.1.6
```

(all the action used must be referenced by the commit sha)

lucacavallaro (changes requested):

lucacavallaro (approved):

PR #26, opened 2024-05-27
[DEVEX-131] New TF module to create an Azure App Service

Krusty93:
### List of changes

A new module to create an Azure App Service.

This module creates:

subnet for the app service
production slot
staging slot
private endpoints
This module DOES NOT use vnet routing rules, but instead uses private endpoints to regulate app service access. Then, access is disabled from outside the vnet (without the vpn).

Naming convention is also forced.

### Motivation and context

Actual module is too complex and does not force to use certain patterns like private endpoints

### Type of changes

- [X] Add new resources

- [X] No

### DIFF
--- a/infra/modules/azure_app_service/app_service_plan.tf
+++ b/infra/modules/azure_app_service/app_service_plan.tf
@@ -1,0 +1,12 @@
+resource "azurerm_service_plan" "this" {
+  count = local.app_service_plan.enable ? 1 : 0
+
+  name                   = "${local.project}-${var.environment.domain}-${var.environment.app_name}-asp-${var.environment.instance_number}"
+  location               = var.environment.location
+  resource_group_name    = var.resource_group_name
+  os_type                = "Linux"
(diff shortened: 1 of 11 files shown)


infra/modules/azure_app_service/app_service_plan.tf
```diff
@@ -0,0 +1,12 @@
+resource "azurerm_service_plan" "this" {
+  count = local.app_service_plan.enable ? 1 : 0
```
gunzip, line 2: may you add a comment on why we can disable service plan?

Krusty93: I think it's described as variable documentation: https://github.com/pagopa/dx/pull/26/files#diff-8f8180c01ecb10d9bb9bed1f97e62b618a83b7b232d939e01a8cdd95182b458dR24

gunzip (approved):

PR #30, opened 2024-05-29
[DEVEX-122] Update Function App workflow to support Function App with a single slot

Krusty93:
### List of changes

Add support to deploy zip on the single slot

### Motivation and context

Not all Function App have a staging slot, this scenario should be supported

### Type of changes

- [ ] Add new resources

- [ ] No

### DIFF
--- a/.github/workflows/function_app_deploy.yaml
+++ b/.github/workflows/function_app_deploy.yaml
@@ -131,7 +131,7 @@ jobs:
 
       - name: Deploy to Staging Slot
         uses: azure/webapps-deploy@v2
-        if: ${{ env.USE_STAGING_SLOT }}
+        if: ${{ env.USE_STAGING_SLOT == 'true' }}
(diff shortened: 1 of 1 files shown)


.github/workflows/function_app_deploy.yaml
```diff
@@ -123,15 +129,25 @@ jobs:
           tenant-id: ${{ env.ARM_TENANT_ID }}
           subscription-id: ${{ env.ARM_SUBSCRIPTION_ID }}
 
+      - name: Deploy
+        uses: azure/webapps-deploy@v2
+        if: ${{ env.USE_STAGING_SLOT == 'false' }}
```
gunzip, line 134: isn't it

```suggestion
        if: ${{ env.USE_STAGING_SLOT == false }}
```

without quotes?

gunzip (approved):

PR #31, opened 2024-05-30
[DEVEX-133] Fix the Storage Account name generation in Azure Functions App module

Krusty93:
### List of changes

The `-` character replace is performed for the entire string and not for the prefix only

### Motivation and context

If the `app_name` contains a `-`, it must be removed too

### Type of changes

- [ ] Add new resources

- [X] No


christian-calabrese (approved):
LGTM

PR #35, opened 2024-06-03
[DEVEX-138] Add `fn` suffix to Function App Storage Account name

Krusty93:
### List of changes

Add `fn` suffix to the name of Storage Account created in Azure Function App module

### Motivation and context

If a workload is composed by a Function App and a Storage Account, the name of the Storage Account would be exactly the same as the one created as a backend for the Function App itself.

Example:
Function App: `io-p-services-cms-webapp-fn-01`
Function App Storage Account: `io-p-services-cms-webapp-st-01`
Storage Account with application-specific business logic: `io-p-services-cms-webapp-st-01`

Applying this pr they would be:
Function App: `io-p-services-cms-webapp-fn-01`
Function App Storage Account: `io-p-services-cms-webapp-stfn-01`
Storage Account with application-specific business logic: `io-p-services-cms-webapp-st-01`

The abbreviation `stfn` has been deducted from [Microsoft abbreviation list](https://learn.microsoft.com/en-us/azure/cloud-adoption-framework/ready/azure-best-practices/resource-abbreviations#compute-and-web), in particular from `VM storage account` entry (`stvm`)

N.B. Storage Account names don't have `-` in names, I've included them in for readiness ease

### Type of changes

- [ ] Add new resources

- [X] Yes, users may be impacted applying this change
- [ ] No

### DIFF
--- a/infra/modules/azure_function_app/locals.tf
+++ b/infra/modules/azure_function_app/locals.tf
@@ -18,7 +18,7 @@ locals {
 
   storage_account = {
     replication_type = var.tier == "test" ? "LRS" : "ZRS"
-    name             = replace("${local.project}${var.environment.domain}${var.environment.app_name}st${var.environment.instance_number}", "-", "")
+    name             = replace("${local.project}${var.environment.domain}${var.environment.app_name}stfn${var.environment.instance_number}", "-", "")
   }
 
   private_dns_zone = {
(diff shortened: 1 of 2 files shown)


gunzip (approved):

gunzip:
how to migrate?

Krusty93:
> how to migrate?

I do expect Terraform will destroy and create the new resource, updating the reference in Function App settings. Keys should be saved there then.
However, we do not have any production product yet

PR #41, opened 2024-06-07
Upgrade Terraform version constraints

Krusty93:
### List of changes

Upgrade Terraform version constraints in `azure_federated_identity_with_github` and `github_selfhosted_runner_on_container_app_jobs` modules

### Motivation and context

Dependency doesn't have strictly requirements anymore, so it is possible to do the same here

### Type of changes

- [ ] Add new resources

- [ ] No


christian-calabrese (approved):
LGTM

PR #42, opened 2024-06-10
[DEVEX-134] Add Autoscaler module for App Service and Function App

Krusty93:
### List of changes

Add autoscaler module for both App Service and Function App hosted by App Service Plan

### Motivation and context

Simplify autoscaler management

Minimal usage:
```hcl
module "autoscaler" {
  source = "../azure_app_service_plan_autoscaler"

  environment = {
    prefix          = "io"
    env_short       = "d"
    location        = "italynorth"
    domain          = "test"
    app_name        = "autoscaler"
    instance_number = "01"
  }

  resource_group_name = "dev-andreag"
  app_service_plan_id = "/subscriptions/a4e96bcd-59dc-4d66-b2f7-5547ad157c12/resourceGroups/dev-andreag/providers/Microsoft.Web/serverfarms/io-d-test-autoscaler-asp-01"
  app_service_id      = "/subscriptions/a4e96bcd-59dc-4d66-b2f7-5547ad157c12/resourcegroups/dev-andreag/providers/Microsoft.Web/sites/io-d-test-autoscaler-app-01"

  scheduler = {
    normal_load = {
      default = 2
      minimum = 1
      maximum = 5
    }
  }

  tags = {
    Test = "autoscale"
  }
}
```

Set one recurrent profile for low load
```hcl
module "autoscaler" {
  source = "../azure_app_service_plan_autoscaler"

  scheduler = {
    normal_load = {
      default = 2
      minimum = 1
      maximum = 5
    }
    low_load = {
      default = 1
      minimum = 1
      maximum = 2
      start = {
        hour = 18
        minutes = 30
      }
      end = {
        hour = 19
        minutes = 45
      }
      name = "low"
    }
  }
}

```

Set two profiles for low and high load
```hcl
module "autoscaler" {
  source = "../azure_app_service_plan_autoscaler"

  scheduler = {
    normal_load = {
      default = 2
      minimum = 1
      maximum = 5
    }
    low_load = {
      default = 1
      minimum = 1
      maximum = 2
      start = {
        hour = 18
        minutes = 30
      }
      end = {
        hour = 19
        minutes = 45
      }
      name = "low"
    }
    high_load = {
      default = 3
      minimum = 3
      maximum = 8
      start = {
        hour = 04
        minutes = 30
      }
      end = {
        hour = 05
        minutes = 45
      }
      name = "high"
    }
  }
}
```

Set custom metrics over requests
```hcl
module "autoscaler" {
  source = "../azure_app_service_plan_autoscaler"

  scheduler = {
    normal_load = {
      default = 2
      minimum = 1
      maximum = 5
    }
  }

  scale_metrics = {
    requests = {
      upper_threshold = 1000
      lower_threshold = 100
      increase_by     = 1
      decrease_by     = 1
    }
  }
}
```

Set custom metrics overriding default values (CPU and memory are mandatory, if not set default values are used)
```hcl
module "autoscaler" {
  source = "../azure_app_service_plan_autoscaler"

  scheduler = {
    normal_load = {
      default = 2
      minimum = 1
      maximum = 5
    }
  }

  scale_metrics = {
    cpu = {
      upper_threshold = 50
      lower_threshold = 5
      increase_by     = 1
      decrease_by     = 1
    }
  }
}
```

### Type of changes

- [X] Add new resources

- [X] No


gunzip:
1. can we have some presets with default values (ie. low, standard, high) ? I mean, other modules that use this one but setup some defaults
2. since the autoscaler it's tied to one app service, can we populate all the environment values automatically from the app service resource one? ie. accepting some parameters for a data resource

Krusty93:
> 1. can we have some presets with default values (ie. low, standard, high) ? I mean, other modules that use this one but setup some defaults

Not sure, it's difficult to estimate the number of requests a workload should support and their thresholds. Values can vary a lot. It's the same reason why I set defaults on CPU and memory (_percentage_ values) but not on number of requests.

Any suggestion?

According to this logic, I tried to set a maximum of 3 recurrent schedules (normal, low and high) trying to abstract lot of details

> 2. since the autoscaler is tied to one app service, can we populate all the environment values automatically from the app service resource ones? ie. accepting some parameters for a data resource. same for `resource_group_name` (or viceveversa, computing `app_service_plan_id` and `app_service_id` from environment values?)

No because we don't know if the id is about an App Service or a Function App, so we don't know which resource type use for the `data` block

infra/modules/azure_naming_convention/locals.tf
```diff
@@ -0,0 +1,5 @@
+locals {
+  location_short       = var.environment.location == "italynorth" ? "itn" : var.environment.location == "westeurope" ? "weu" : var.environment.location == "germanywestcentral" ? "gwc" : "neu"
```
gunzip, line 2: I'd rather not default and use a map, something alike:

```
locals {
  location_map = {
    "italynorth" : "itn",
    "westeurope" : "weu",
    "germanywestcentral" : "gwc"
  }
  location_short = lookup(location_map, var.environment.location, "Error: Invalid location")

  project              = "${var.environment.prefix}-${var.environment.env_short}-${local.location_short}"
  resource_name_prefix = "${local.project}-${var.environment.domain}-${var.environment.app_name}"
}

```

gunzip: we may centralize this map someway

gunzip:
About

> No because we don't know if the id is about an App Service or a Function App, so we don't know which resource type use for the data block

Can we provide a variable for the type (either function or app service) or use two different parameters?

The goal is to avoid to specify the whole environment section here which can be extracted from appservice / function module outputs:

```
environment = {
    prefix          = "io"
    env_short       = "d"
    location        = "italynorth"
    domain          = "test"
    app_name        = "autoscaler"
    instance_number = "01"
  }

  resource_group_name = "dev-andreag"
  app_service_plan_id = "/subscriptions/a4e96bcd-59dc-4d66-b2f7-5547ad157c12/resourceGroups/dev-andreag/providers/Microsoft.Web/serverfarms/io-d-test-autoscaler-asp-01"
  app_service_id      = "/subscriptions/a4e96bcd-59dc-4d66-b2f7-5547ad157c12/resourcegroups/dev-andreag/providers/Microsoft.Web/sites/io-d-test-autoscaler-app-01"

```

Ideally this configuration should be reduced to only one parameter, either:
 
`app_service_id  = module.some_app_service.outputs.id` or  `function_app_id   = module.some_function_app.outputs.id` and everything else computed from some data source.

I don't know if it's possibile.

gunzip:
Example.

```
data "azurerm_app_service_plan" "example" {
  name                = "my-app-service-plan"
}
```

```
locals {
  environment = {
    instance_number = 1  # You can set this to a fixed value or derive it from another logic
    location        = data.azurerm_app_service_plan.example.location
  }
}
```

```
module "autoscale_setting" {
  source = "../azure_app_service_plan_autoscaler"

  app_service_plan_id = var.app_service_plan_id
  tags                = var.tags

  autoscale_settings = {
    default = 2
    minimum = 1
    maximum = 5
  }

  environment = local.environment
}
```

```
resource "azurerm_monitor_autoscale_setting" "this" {
  name                = "autoscale-${local.environment.instance_number}"
  location            = local.environment.location
  target_resource_id  = var.app_service_plan_id

  profile {
    name = "default-autoscale-profile"

    capacity {
      default = var.autoscale_settings.default
      minimum = var.autoscale_settings.minimum
      maximum = var.autoscale_settings.maximum
    }
  }

  tags = var.tags
}
```

infra/modules/azure_naming_convention/locals.tf
Krusty93: Indeed a better solution. The code is already centralized in the module `azure_naming_convention`

infra/modules/azure_app_service_plan_autoscaler/variables.tf
```diff
@@ -0,0 +1,122 @@
...
+  description = "Set name of the App Service to monitor"
+  default     = null
+}
+
+variable "function_app_name" {
+  type        = string
+  description = "Set the name of the Function App to monitor"
+  default     = null
+}
+
```
gunzip, line 22: what about adding some condition (exactly one between function_app_name OR app_service_name must be set):

```hcl
variable "app_service_name" {
  description = "The name of the App Service"
  type        = string
  default     = null
}

variable "function_app_name" {
  description = "The name of the Function App"
  type        = string
  default     = null
}

# Validate that exactly one of the two variables is set
variable "validate_one_set" {
  description = "Validation variable"
  default     = null

  validation {
    condition     = (var.app_service_name != null) != (var.function_app_name != null)
    error_message = "Exactly one of 'app_service_name' or 'function_app_name' must be set."
  }
}
```

gunzip (commented):
better!

infra/modules/azure_app_service_plan_autoscaler/variables.tf
```diff
@@ -0,0 +1,122 @@
...
+    high_load = optional(object({
+      start = object({
+        hour    = number
+        minutes = number
+      })
+      end = object({
+        hour    = number
+        minutes = number
+      })
+      name    = string
```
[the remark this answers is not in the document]
gunzip: something like

```hcl
default = {
    high_load = {
      start = {
        hour = 9
        minutes = 0
      }
      end = {
        hour = 17
        minutes = 0
      }
      name = "high_load_schedule"
      default = 5
      minimum = 2
      maximum = 10
    }
    low_load = {
      start = {
        hour = 0
        minutes = 0
      }
      end = {
        hour = 6
        minutes = 0
      }
      name = "low_load_schedule"
      default = 1
      minimum = 1
      maximum = 3
    }
    normal_load = {
      default = 3
      minimum = 2
      maximum = 5
    }
  }
```

infra/modules/azure_app_service_plan_autoscaler/locals.tf
```diff
@@ -0,0 +1,148 @@
...
+
+  is_app_service  = var.app_service_name != null
+  is_function_app = var.function_app_name != null
+
+  base_name = local.is_app_service ? data.azurerm_linux_web_app.this[0].name : data.azurerm_linux_function_app.this[0].name
+  # base_name = regex("[^-]+-[^-]+-[^-]+-[^-]+-[^-]+", local.is_app_service ? data.azurerm_linux_web_app.this[0].name : data.azurerm_linux_function_app.this[0].name)
+  # number    = regex("[0-9]+$", local.is_app_service ? data.azurerm_linux_web_app.this[0].name : data.azurerm_linux_function_app.this[0].name)
+
+  # autoscale_name      = "${local.base_name}-as-${local.number}"
+  autoscale_name      = replace(replace(replace(local.base_name, "fn", "as"), "func", "as"), "app", "as")
```
gunzip, line 11: I'd keep this is let us replace the current autoscaler that are cut and pasted everywhere.

infra/modules/azure_app_service_plan_autoscaler/variables.tf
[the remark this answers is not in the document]
BurnedMarshal: What is the motivation behind having different maximum values ​​in different time slots? Instances will be allocated only if the autoscale rules will be triggered. IMHO the maximum should be an unique value in input.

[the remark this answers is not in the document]
Krusty93: @BurnedMarshal I agree with your idea, I'm not sure about the variable structure consistency. However, I'll bring it in

@gunzip I would avoid any default value to force consumers to rationally think about values to use and avoid any surprise as they could lead to disruptions. Thoughts?

infra/modules/azure_app_service_plan_autoscaler/locals.tf
Krusty93: The procedure to remove the existing autoscaler from the state and import it into the new one is long and useless, and often it tries to recreate it even if the target resource id stays the same for some reason (I did it several times during TF refactoring tasks)

In my opinion, the quickest way to do it is to replace the existing definition with this module and apply it ignoring the resource recreation/update

infra/modules/azure_app_service_plan_autoscaler/variables.tf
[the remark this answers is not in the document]
gunzip: I'd try to share some defaults for common use cases we already are using.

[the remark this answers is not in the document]
Krusty93: done!

infra/modules/azure_app_service_plan_autoscaler/variables.tf
```diff
@@ -0,0 +1,157 @@
...
+    normal_load = object({
+      default = number
+      minimum = number
+    })
+    maximum = optional(number, 30)
+  })
+
+  default = {
+    high_load = {
+      name = "high_load_profile"
```
gunzip, line 63: this format needs some documentation. for example, what's the difference between 'default' and 'minimum'? what `start` / `end` means?

gunzip (commented):
starts looking good :-)

infra/modules/azure_app_service_plan_autoscaler/variables.tf
Krusty93: In Docusaurus or where else?

gunzip: here in the module itself (ie. in the variable description)

infra/modules/azure_app_service_plan_autoscaler/variables.tf
```diff
@@ -0,0 +1,157 @@
...
+  type = object({
+    app_service_name  = optional(string)
+    function_app_name = optional(string)
+  })
+
+  validation {
+    condition = (
+      (var.target_service.app_service_name != null && var.target_service.function_app_name == null) ||
+      (var.target_service.app_service_name == null && var.target_service.function_app_name != null)
+    )
```
gunzip, line 21:
```suggestion
validation {
  condition = (var.target_service.app_service_name != null) != (var.target_service.function_app_name != null)
}

```

infra/modules/azure_app_service_plan_autoscaler/variables.tf
```diff
@@ -0,0 +1,157 @@
...
+      minimum = 2
+    }
+    normal_load = {
+      default = 11
+      minimum = 3
+    }
+    maximum = 30
+  }
+
+  description = "Set the recurrent autoscaling profiles, including start and end time ([hh]:[mm]), the minimum and maximum number of instances and the default value (used when metrics are not available for some technical issue). Each default value can be overridden"
```
gunzip, line 95: I'd rename default to `fallback`. Moreover, can we make it clear how the specified interval works against the min/max values? it's not clear what happens outside / inside  that interval

gunzip (changes requested):

infra/modules/azure_app_service_plan_autoscaler/variables.tf
```diff
@@ -0,0 +1,154 @@
...
+      minimum = 2
+    }
+    normal_load = {
+      default = 11
+      minimum = 3
+    }
+    maximum = 30
+  }
+
+  description = "Set the recurrent autoscaling profiles, including start and end time ([hh]:[mm]), the minimum and maximum number of instances and the fallback value (used when metrics are not available for some technical issue). Outside of low/high load profile time span, \"normal\" load values are used. Each default value can be overridden."
```
gunzip, line 92:
```suggestion
  description = "Set the recurrent autoscaling profiles, including start and end time ([hh]:[mm]), the minimum and maximum number of instances and the fallback ("default") value (used when metrics are not available for some technical issue). Outside of low/high load profile time span, \"normal\" load values are used. Each default value can be overridden."
```

gunzip (approved):

PR #44, opened 2024-06-14
[DEVEX-91] azure_role_assignments module implemented

christian-calabrese:
### List of changes

Implemented the terraform module to allow principals (mostly managed identities and user assigned identities) to read, write or manage the most commonly used resources.

The supported roles for each role assignment are:
- reader
- writer
- owner

As of the time of writing the following resources are supported as scopes of the role assignments:
- Cosmos DB:
    - Account
    - Database
    - Collections
- Event Hubs:
    - Namespace
    - Event Hub
- Key Vault (both RBAC or Access Policies)
    - Secrets
    - Keys
    - Certificates
- Redis
- Storage Accounts
    - Queue (all in a storage account or single ones)
    - Table (all in a storage account or single ones)
    - Blob Container (all in a storage account or single ones)

More services that will be implemented in the next future are:
- Service Bus
- Function App
- App Service
- Notification Hub

### Motivation and context

The aim of this module is to simplify the segmented management of RBAC assignments in Azure.

### Other information

To request more supported services, please contact the DevEx team


infra/modules/azure_role_assignments/variables.tf
```diff
@@ -0,0 +1,87 @@
...
+}
+
+variable "cosmos" {
+  description = "A list of CosmosDB role assignments"
+  type = list(object({
+    account_name        = string
+    resource_group_name = string
+    role                = string
+    database            = optional(string, "*")
+    collection          = optional(string, "*")
```
Krusty93, line 13: Should we use a `list(string)` for both databases and collections?

infra/modules/azure_role_assignments/variables.tf
```diff
@@ -0,0 +1,87 @@
...
+  description = "The ID of the principal to which assign roles. It can be a managed identity."
+  type        = string
+}
+
+variable "cosmos" {
+  description = "A list of CosmosDB role assignments"
+  type = list(object({
+    account_name        = string
+    resource_group_name = string
+    role                = string
```
Krusty93, line 11: Is this value supposed to be validated to consider only `writer` and `reader` as valid values?

infra/modules/azure_role_assignments/main.tf
```diff
@@ -0,0 +1,14 @@
...
+      source  = "hashicorp/azurerm"
+      version = ">= 3.100.0"
+    }
+  }
+}
+
+provider "azurerm" {
+  features {}
+
+  storage_use_azuread = true
```
Krusty93, line 13: Do we need this?

Krusty93:
Very nice work!

infra/modules/azure_role_assignments/variables.tf
christian-calabrese: Good point! Initially, I actually started with a list of strings. However, the code became a little more intricate, so I thought that, if the caller needs to add permissions for different databases and collections, would do something like this:

```
module "roles" {
  source       = "../../"
  principal_id = module.function.function_app.function_app.principal_id

  cosmos = [
    {
      account_name        = "io-p-cosmos-example-cms"
      resource_group_name = "io-p-example-cms-rg"
      role                = "writer"
      database            = "a"
      collection          = "x"
    },
    {
      account_name        = "io-p-cosmos-example-cms"
      resource_group_name = "io-p-example-cms-rg"
      role                = "writer"
      database            = "a"
      collection          = "y"
    }
  ]
}
```

I see that it can become large quickly, so the alternative can be to make the collection parameter as a list(string). I wouldn't make the database parameter a list(string) because we would lose the database => collection relationship.
In this case, it would look like this which looks better indeed:

```
module "roles" {
  source       = "../../"
  principal_id = module.function.function_app.function_app.principal_id

  cosmos = [
    {
      account_name        = "io-p-cosmos-example-cms"
      resource_group_name = "io-p-example-cms-rg"
      role                = "writer"
      database            = "a"
      collection          = ["x", "y"]
    }
  ]
}
```

infra/modules/azure_role_assignments/variables.tf
christian-calabrese: Right! Validation will be put in place asap

infra/modules/azure_role_assignments/examples/function_app/main.tf
```diff
@@ -0,0 +1,68 @@
...
+  source       = "../../"
+  principal_id = module.function.function_app.function_app.principal_id
+
+  cosmos = [
+    {
+      account_name        = "io-p-cosmos-example-cms"
+      resource_group_name = "io-p-example-cms-rg"
+      role                = "writer"
+      database            = "a"
+      collection          = ["x", "y"]
```
gunzip, line 40:
```suggestion
      collections          = ["x", "y"]
```

Krusty93 (approved):

PR #47, opened 2024-06-17
Add web_app_deploy workflow

lucacavallaro:
### List of changes
1. Add `web_app_deploy.yaml` workflow

#### Changes from `function_app_deploy.yaml`
1. This workflow builds the artifact using `esbuild` and `zip` instead of `npm-pack-zip` and `bundledDependencies`
2. The artifacts created by this workflow contains a "plain old javascript project" with `index.js`, `package.json` and `node_modules`
3. This workflow can create artifacts for all kind of workspaces (function apps, app services) as long they have an entrypoint specified in the `main` field of their `package.json` file

### Motivation and context
This workflow is required to use `Yarn Plug'n'Play` in production workflow and to unify the artifact building for node projects.

### Type of changes

- [ ] Add new resources

- [ ] No


.github/workflows/web_app_deploy.yaml
```diff
@@ -0,0 +1,159 @@
...
+        uses: actions/upload-artifact@694cdabd8bdb0f10b2cea11669e1bf5453eed0a6 # v4.2.0
+        with:
+          name: ${{ env.BUNDLE_NAME }}
+          path: ${{ steps.make-artifact.outputs.artifact-path }}
+          if-no-files-found: error
+          retention-days: 7
+
+  deploy:
+    name: Deploy
+    if: ${{ !github.event.act }}
```
gunzip, line 96:
```suggestion
if: ${{ github.event_name != 'workflow_call' }}
```

?

.github/workflows/web_app_deploy.yaml
```diff
@@ -0,0 +1,159 @@
...
+
+      - name: Azure Login
+        uses: azure/login@v2 # v2.0.0
+        with:
+          client-id: ${{ env.ARM_CLIENT_ID }}
+          tenant-id: ${{ env.ARM_TENANT_ID }}
+          subscription-id: ${{ env.ARM_SUBSCRIPTION_ID }}
+
+      - name: Deploy
+        uses: azure/webapps-deploy@v2
```
Krusty93, line 127: I did some deployment test with AZ CLI to remove the `curl` ping due to recent issues and applying a more reliable approach.

Here the test:
scenario:
- app with a single endpoint that responds 200
- deploy from local via az cli
- 3 instances per instance

steps:
1. application deployment that responds 200 to the root with the `az webapp deploy --src-path app.zip` command, which seems to be (almost) synchronous
2. staging slot creation:
    - the same application that runs on prod slot responds both on `<app>-staging.azurewebsites.net` and `<app>.azurewebsites.net?x-ms-routing-name=staging`
3. enabled health checks
    - all instances are green on both slots
4. changed the code to reply 400 instead of 200
5. deployment of the “broken” app on staging slot
    - the deployment “never” finishes, in the sense that it tries again until the timeout of about 15 minutes expires (timeout is customizable)
    - as expected, the working app continues to respond 200 to the url `<app>.azurewebsites.net`
    - the broken app responds with 400 to the url `<app>-staging.azurewebsites.net`
    - unexpectedly, the url `<app>.azurewebsites.net?x-ms-routing-name=staging` responds with 200 instead of 400
    - prod shows all healthy instances, staging shows them all red
6. slot swap
    - as expected, `<app>.azurewebsites.net` responds 400 and `<app>-staging.azurewebsites.net` responds 200
    - <app>.azurewebsites.net?x-ms-routing-name=staging responds 400
    - as expected, healths are reversed (prod red, staging green)
7. new slot swap
    - green prod, red staging
8. set of these variables in both slots:
    - WEBSITE_WARMUP_PATH: / (`A relative path on the site that should be pinged whenever the site restarts (not only during slot swaps)
    - WEBSITE_SWAP_WARMUP_PING_STATUSES: 200 (Valid HTTP response codes for the warm-up operation. Add this app setting with a comma-separated list of HTTP codes. An example is 200,202 . If the returned status code isn't in the list, the warmup and swap operations are stopped. By default, all response codes are valid.)
9. attempted new swap, got an error:
   - Failed to swap slots 'staging' and 'production': Cannot swap slots for site 'io-d-itn-test-autoscaler-app-02' because the application initialization in 'staging' slot either took too long or failed. Please check AppInit module configuration or try using swap with preview if application initialization time is very long.
    - as expected, the app in production continues to respond as before

Given this test, I would set these environment variables in all services (including our module update), remove curl from the pipeline and run the swap command.
I would also deploy the package using with az cli because the action [uses the `async=true` query string](https://github.com/microsoft/pipelines-appservice-lib/blob/0af7ef8d67738e5249f8f5e3c5e8a036b940ecf5/packages/appservice-rest/src/Utilities/KuduServiceUtility.ts#L108). Through az cli we could use a synchronous approach and set the timeout to 1 minute or little more.
I would also set the variable `WEBSITE_SWAP_WARMUP_PING_PATH` (`The path to ping over HTTP to warm up your site. Add this app setting by specifying a custom path that begins with a slash as the value`), which from what I understand is the same as WEBSITE_WARMUP_PATH but is only called for swap operations (I think in case you want to use different endpoints for swap and restart, but you might as well set them both)

.github/workflows/web_app_deploy.yaml
```diff
@@ -0,0 +1,184 @@
...
+        working-directory: ./out
+
+      - name: Build
+        run: yarn build
+        working-directory: ./out
+
+      - name: Build the artifact
+        id: make-artifact
+        env:
+          YARN_NODE_LINKER: node-modules
```
gunzip, line 79:
```suggestion
         # we fall back to node-moules, even in case pnp is configured, in order to avoid bundling dependendencies
          YARN_NODE_LINKER: node-modules
```

.github/workflows/web_app_deploy.yaml
```diff
@@ -0,0 +1,184 @@
...
+          YARN_NM_HOISTING_LIMITS: workspaces
+        run: |
+          set -e
+
+          ENTRY_POINT=$(jq -r ${{ inputs.target_name == 'main' && '.main' || format('.exports.{0}', inputs.target_name) }} package.json)
```
gunzip, line 84:
```suggestion
          # determine the entry point file name from package.json main property
          ENTRY_POINT=$(jq -r ${{ inputs.target_name == 'main' && '.main' || format('.exports.{0}', inputs.target_name) }} package.json)
```

.github/workflows/web_app_deploy.yaml
```diff
@@ -0,0 +1,184 @@
+on:
+  workflow_call:
+    inputs:
+      workspace_name:
+        description: The name of the workspace to create the artifact for.
+        type: string
+        required: true
+      target_name:
+        description: "(Optional) The name of the target to create artifact for (default: main)"
```
gunzip, line 9: what do you mean here with 'target'?

lucacavallaro: Workspaces can have multiple entry points, `target_name` is the name to the application to deploy. I borrowed the name `target` from other build tools such as `make`, `xmake`, `target`

`package.json` has two ways to specify an entry point
1. `main` field - legacy and simpler, used by old CJS-only projects. It supports only one `entry point` (the "default one")
2. `exports` field - more advanced, that supports **multiple** entries.

`target_name` - in our context - is the name of the `package.json` entry that holds the path of the entry point file.

### Example 1 (single entry point, basic setup)

```json
{
  "name": "my-app",
  "version": "1.0.0",
  "main": "./dist/index.js"
}
```

`target_name` should be omitted or set to `"main"`

### Example 2 (mutliple entry points)

```json
{
  "name": "my-app",
  "version": "1.0.0",
  "exports": {
    "web": "./dist/web.js",
    "func": "./dist/func.js"
  }
}
```

`target_name` should be either `"web"` or `"func"`

---

This workflow will reduce all workspaces in a node.js project with a single `main` field to enhance compatibility (Azure Functions, for example, relies on `main` field to find the entry point)

I chose not to call this input "entry_point" to avoid confusion. Also because it is not the entry point of the code (`./src/index.ts`) but of the application (`./dist/index.js`)

gunzip: this should be documented someway otherwise it's not clear to the user what 'target_name' means in this context

lucacavallaro: I decided to simplify it using only `main` as target, so I removed this input.

.github/workflows/web_app_deploy.yaml
lucacavallaro: I updated the workflow to use `az webapp deploy` instead the GitHub Action. I think we should update our web apps tf modules to include `WEBSITE_WARMUP_PATH` by default (pointing to the health check path)

Krusty93: I like your first change, I'm going to approve the PR on my side
Instead, I don't think adding the `WEBSITE_WARMUP_PATH` variable is correct as we've seen some issues while spinning up the container (probably things are not ready and it is restarted multiple times), while I would add `WEBSITE_SWAP_WARMUP_PING_PATH` with the health ping

gunzip (approved):

Krusty93 (approved):

PR #51, opened 2024-06-25
[DEVEX-152] Add CI workflow to validate changes in Terraform code

Krusty93:
### List of changes

Add CI workflow to run `pre-commit` hook at PR creation/modification time
Validation is performed only on modified files

### Motivation and context

Be sure to validate each Terraform change (formatting, security, etc.) to do not break the code.

### Type of changes

- [X] Add new resources

- [X] No


.github/workflows/static_analysis.yaml
```diff
@@ -0,0 +1,62 @@
...
+          git fetch --no-tags --prune --depth=1 origin +refs/heads/*:refs/remotes/origin/*
+
+      - name: Get changed files
+        id: file_changes
+        run: |
+          set -eu
+
+          export DIFF=$(git diff --name-only origin/${{ github.base_ref }} ${{ github.sha }})
+          echo "Diff between ${{ github.base_ref }} and ${{ github.sha }}"
+          echo "files=$( echo "$DIFF" | xargs echo )" >> $GITHUB_OUTPUT
```
gunzip, line 47: shouldn't we escape `$DIFF` someway?

.github/workflows/static_analysis.yaml
```diff
@@ -0,0 +1,62 @@
+name: Terraform Analysis
```
gunzip, line 1: can we have a comment / description about what this action is supposed to do?

gunzip (changes requested):

.github/workflows/static_analysis.yaml
Krusty93: Added

.github/workflows/static_analysis.yaml
Krusty93: Added

.github/workflows/static_analysis.yaml
```diff
@@ -0,0 +1,73 @@
+# Perform validation using pre-commit hook on .tf files
```
gunzip, line 1:
```suggestion
# Perform validation using pre-commit hook on .tf files
```
```suggestion
# This pipeline assumes the presence of a .pre-commit-config.yaml file at 
# the top level of the repository, which contains Terraform validation tasks.
# For more details, refer to the pre-commit configuration file:
# https://github.com/pagopa/dx/blob/main/.pre-commit-config.yaml.
```

gunzip (approved):

PR #54, opened 2024-07-03
[EC-350] Update Drift Detection workflow with Failure notification

mamu0:
### List of changes

Updated Drift Detection by removing unused variables and introducing a new step for Slack notification in case of each failure.
Removed webook_url input to avoid clear value in logs.
Changed Slack notification action with `slackapi` official.
### Motivation and context

We have added a notification step to receive alerts for issues beyond drift detection related to pipeline execution.

### Type of changes

- [ ] Add new resources

- [x] No

### DIFF
--- a/.github/workflows/infra_drift_detection.yml
+++ b/.github/workflows/infra_drift_detection.yml
@@ -13,10 +13,6 @@ on:
         type: string
         required: true
         default: infra/resources/
-      env_vars:
-        description: List of environment variables to set up, given in env=value format.
-        type: string
-        required: false
       use_private_agent:
         description: Use a private agent to run the Terraform plan.
         type: boolean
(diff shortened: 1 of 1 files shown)


.github/workflows/infra_drift_detection.yml
```diff
@@ -182,5 +166,22 @@ jobs:
             :wavy_dash: Resource to change: ${{ env.TO_CHANGE }}
             :heavy_minus_sign: Resource to destroy: ${{ env.TO_DESTROY }}
           footer: "Linked Repo <{repo_url}|{repo}>"
+        env:
+          SLACK_WEBHOOK_URL: ${{ env.SLACK_WEBHOOK_URL }}
```
gunzip, line 170: why do we need this aside `webhook_url`?

.github/workflows/infra_drift_detection.yml
```diff
@@ -182,5 +166,22 @@ jobs:
...
+
+      - name: Failure Notification
+        if: ${{ always() && (env.SLACK_WEBHOOK_URL || inputs.webhook_url) }}
+        uses: ravsamhq/notify-slack-action@2.3.0
```
gunzip, line 174: needs to be pinned to hash, you can use https://github.com/mheap/pin-github-action

gunzip: I'd use https://github.com/slackapi/slack-github-action instead which comes from a verified vendor

.github/workflows/infra_drift_detection.yml
gunzip: moreover, this must be a secret and not show up in action logs

gunzip (changes requested):

gunzip (approved):
lgtm, just ensure that slack webhook does not appear in action logs

PR #56, opened 2024-07-05
[CES-4] Add new module for azure app service that can be exposed

mamu0:
### List of changes

- A new module called `azure_app_service_exposed` has been created. Functionally, it can generate an app service like the `azure_app_service` module, but without the networking configuration (subnets, private endpoints, etc.) and is exposed.
- All `README.md` files for the various modules have been updated with a `pre-commit`.
- A bug was fixed in `azure_app_service` related to the creation of the slot during the insertion of the node version.

### Motivation and context

The reason for creating this new module was to develop a cleaner and more generic solution for `azure_app_service`.

### Type of changes

- [x] Add new resources

- [x] No


gunzip:
can't we reuse the existing app service module someway? in this way it might be complex to keep things in sync between two different modules cc @Krusty93

infra/modules/azure_app_service/app_service_slot.tf
```diff
@@ -22,7 +22,7 @@ resource "azurerm_linux_web_app_slot" "this" {
     ip_restriction_default_action     = "Deny"
 
     application_stack {
-      node_version = var.stack == "node" ? var.node_version : null
+      node_version = var.stack == "node" ? "${var.node_version}-lts" : null
```
Krusty93, line 25: Why this change?

infra/modules/azure_app_service_exposed/variables.tf
```diff
@@ -0,0 +1,113 @@
...
+
+variable "sticky_app_setting_names" {
+  type        = list(string)
+  description = "(Optional) A list of application setting names that are not swapped between slots"
+  default     = []
+}
+
+# ------------ NETWORKING ------------ #
+
+variable "enable_public_access" {
```
Krusty93, line 103: What's the purpose of this variable? In which cases it'd be false?

infra/modules/azure_app_service_exposed/variables.tf
```diff
@@ -0,0 +1,113 @@
...
+  type        = bool
+  description = "(Optional) Enable public access to the App Service"
+  default     = true
+}
+
+variable "azurerm_subnet_id" {
```
Krusty93, line 109: ditto

infra/modules/azure_app_service_exposed/app_service.tf
```diff
@@ -0,0 +1,66 @@
...
+  dynamic "sticky_settings" {
+    for_each = length(var.sticky_app_setting_names) == 0 ? [] : [1]
+    content {
+      app_setting_names = var.sticky_app_setting_names
+    }
+  }
+
+  lifecycle {
+    ignore_changes = [
+      app_settings["DOCKER_CUSTOM_IMAGE_NAME"],
```
Krusty93, line 57: I think we don't need this as Docker is not supported by the module, on purpose

Krusty93:
> can't we reuse the existing app service module someway? in this way it might be complex to keep things in sync between two different modules cc @Krusty93

In my opinion, having two modules is much less complex rather than:
  - having a variable to control the creation of all underlying resources
  - having lot of variables with no effect when the var `is_public` is enabled. To me, it would create confusion

This should be a ready to go module to create a public app service without any fancy things around, to get an app up and running in a matter of minutes

infra/modules/azure_app_service/app_service_slot.tf
mamu0: After a plan i noticed that it gave an error because it wanted a string like `"20-tls"` like in `app_service.tf`, but instead it had a number.

infra/modules/azure_app_service_exposed/variables.tf
mamu0: I thought that since it is similar to `azure_app_service`, it might be useful to give a choice, but actually if we want this module to only do the exposed one it is not necessary.

infra/modules/azure_app_service_exposed/app_service.tf
mamu0: ok, i will remove this also in the `azure_app_service` module?

infra/modules/azure_app_service/app_service_slot.tf
Krusty93: @lucacavallaro did you notice this issue when applied [this](https://github.com/pagopa/io-fims/blob/cba305f01699061ad0e69fb1bb2034bac372b353/infra/resources/_modules/web_apps/linux_function_app_op.tf#L12)?

infra/modules/azure_app_service_exposed/app_service.tf
Krusty93: Makes sense

infra/modules/azure_app_service/app_service_slot.tf
Krusty93: @mamu0 go ahead with `-lts` in both modules
Close this conv when you read it

PR #57, opened 2024-07-09
[CES-3] Resources migrated from Prod to Dev Environment

mamu0:
### List of changes

All references from **Prod** have been migrated to **Dev** as the current subscription in use is `DEV-ENGINEERING`.
The **Prod** folders in _identity_ and _resources_ have been deleted.
The terraform configurations for _identity_, _resources_ and _repository_ have been applied.
Changed the _Drift Detection_ Workflow as a template, and added a new one that call it.

### Motivation and context

Since we are now using `DEV-ENGINEERING` as the primary subscription for this repository, all old references to `PROD-IO` and `DEV-IO` have been updated accordingly.

### Type of changes

- [x] Add new resources
- [x] Remove existing resources

- [ ] No


.github/workflows/infra_drift_detection.yml
```diff
@@ -7,7 +7,7 @@ on:
         description: Environment where the resources will be deployed.
```
Krusty93, line 7: I'd keep `prod` in this file as primary consumers are other repos

infra/resources/dev/locals.tf
```diff
@@ -1,12 +1,12 @@
...
   location  = "westeurope"
 
   tags = {
     CostCenter  = "TS310 - PAGAMENTI & SERVIZI"
     CreatedBy   = "Terraform"
-    Environment = "Prod"
+    Environment = "Dev"
     Owner       = "DevEx"
-    Source      = "https://github.com/pagopa/dx/blob/main/infra/resources/prod/westeurope"
+    Source      = "https://github.com/pagopa/dx/blob/main/infra/resources/dev/westeurope"
```
Krusty93, line 10:
```suggestion
    Source      = "https://github.com/pagopa/dx/blob/main/infra/resources/dev"
```

Krusty93 (changes requested):
Few comments

Shouldn't `resources/prod` moved to `resources/dev` as well?

.github/workflows/infra_drift_detection.yml
mamu0: I will disable the schedule? Because with prod the pipe will not work in this repo

infra/resources/dev/locals.tf
```diff
@@ -1,12 +1,12 @@
 locals {
-  env_short = "p"
+  env_short = "d"
   location  = "westeurope"
 
   tags = {
     CostCenter  = "TS310 - PAGAMENTI & SERVIZI"
```
christian-calabrese, line 6: Shall we use the Engineering cost center? cc @Krusty93 
```suggestion
    CostCenter  = "TS700 - ENGINEERING"
```

christian-calabrese: Discard this suggestion. As of now, we don't have certainity about the correctness of this costcenter

Krusty93 (approved):

PR #59, opened 2024-07-09
Set the correct path for engineering-team-cloud-eng in CODEOWNERS

Krusty93:
### List of changes

Replace `docs` with `infra`

### Type of changes

- [ ] Add new resources

- [ ] No


christian-calabrese (approved):
LGTM

PR #70, opened 2024-08-02
fix: App service staging slot name is too long

christian-calabrese:
### List of changes

Set app service staging slot name to "staging"

### Motivation and context
The slot staging name is automatically concatenated to the name of the app service, therefore it is not necessary to do it via terraform to prevent extremely long names

### Type of changes

- [ ] Add new resources

- [X] No

### DIFF
--- a/infra/modules/azure_app_service/app_service_slot.tf
+++ b/infra/modules/azure_app_service/app_service_slot.tf
@@ -1,7 +1,7 @@
 resource "azurerm_linux_web_app_slot" "this" {
   count = local.app_service.is_slot_enabled
 
-  name = "${local.project}-${var.environment.domain}-${var.environment.app_name}-staging-app-${var.environment.instance_number}"
+  name = local.app_service_slot.name
 
   app_service_id = azurerm_linux_web_app.this.id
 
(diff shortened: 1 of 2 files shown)


gunzip:
isn't this a breaking chage for all of those who already use the old name? I mean in case, they're pointing to main

Krusty93 (approved):

gunzip:
is it possible that the deploy pipeline points to the old name?

christian-calabrese:
> isn't this a breaking chage for all of those who already use the old name? I mean in case, they're pointing to main

Yes, it is, but luckily nobody is using the module with the slot enabled yet. The only project is io-services-cms, which developers suggested the fix

Krusty93:
> > isn't this a breaking chage for all of those who already use the old name? I mean in case, they're pointing to main
> 
> Yes, it is, but luckily nobody is using the module with the slot enabled yet. The only project is io-services-cms, which developers suggested the fix

Exactly, only reference found is https://github.com/pagopa/io-services-cms/blob/1e19669026c7014f82df3a57b1d60b01d544d569/infra/_modules/backoffice/main.tf#L6

christian-calabrese:
> is it possible that the deploy pipeline points to the old name?

In the module from terraform-azurerm-v3, the name of the slot is chosen by the user, so it shouldn't be an issue

PR #71, opened 2024-08-05
Fix Function App module issue against host ID collision

Krusty93:
### List of changes

Fix the issue by adding `AzureFunctionsWebHost__hostid` a sticky setting.

Add validation to avoid double entry for `AzureFunctionsWebHost__hostid` key in case of is already passed by the client.
Set `APPINSIGHTS_SAMPLING_PERCENTAGE` as sticky setting and set it to 100% for staging slot

Update azurerm provider and state file name

### Motivation and context

An issue with host id collision was found. Steps to reproduce it:
- Create a function so that staging function name is longer than 32 chars (i.e. `ts-d-itn-consumers-01`
- The Function App will use the default value for `AzureFunctionsWebHost__hostid`, while the staging slot overrides it using `staging` as it is longer than 32 chars
- When swapping the slots, the main one takes `staging` as value as well, so the two host ids collide and staging slot loses the app setting

Adding the key name in sticky setting list fixes the issue

### Type of changes

- [ ] Add new resources

- [X] No

### DIFF
--- a/infra/modules/azure_function_app/README.md
+++ b/infra/modules/azure_function_app/README.md
@@ -12,7 +12,7 @@
 
 | Name | Version |
 |------|---------|
-| <a name="provider_azurerm"></a> [azurerm](#provider\_azurerm) | >= 3.100.0 |
+| <a name="provider_azurerm"></a> [azurerm](#provider\_azurerm) | 3.113.0 |
 
 ## Modules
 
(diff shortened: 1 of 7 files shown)


mamu0 (approved):
LGTM

christian-calabrese (approved):

PR #73, opened 2024-08-06
fix: Set cancel-in-progress to false in infra_apply

christian-calabrese:
### Type of changes

- [ ] Add new resources

- [ ] No

### DIFF
--- a/.github/workflows/infra_apply.yaml
+++ b/.github/workflows/infra_apply.yaml
@@ -28,7 +28,7 @@ env:
 
 concurrency:
   group: ${{ github.workflow }}-cd
-  cancel-in-progress: true
+  cancel-in-progress: false
 
 jobs:
 


Krusty93 (approved):

mamu0 (approved):

gunzip:
can we enqueue further runs instead of running them in parallel?

christian-calabrese:
> can we enqueue further runs instead of running them in parallel?

I think that's already the behavior due to the concurrency group. However, [as the doc states](https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/using-concurrency#using-concurrency-in-different-scenarios), only one job can be enqueued

When a concurrent job or workflow is queued, if another job or workflow using the same concurrency group in the repository is in progress, the queued job or workflow will be pending. Any pending job or workflow in the concurrency group will be canceled. This means that there can be at most one running and one pending job in a concurrency group at any time.

PR #77, opened 2024-08-08
Add optional slot swap

BurnedMarshal:
### List of changes
Add some improvements in deploy webapp pipeline:
- Add ping health check endpoint when a slot is present
- Add optional swap step configuration for selective deploy in production

### Motivation and context
The deploy pipeline has not a step that verify that the health check endpoint is alive after the deployment. A broken web app could be deployed causing production outages.
For some kind of deploy the team should have the possibility to deploy only on staging slots testing the application before the production switch. Moreover some deploy could be handled with progressive traffic routing between stage and productions slots.

### Type of changes

- [ ] Add new resources

- [X] No

### DIFF
--- a/.github/workflows/web_app_deploy.yaml
+++ b/.github/workflows/web_app_deploy.yaml
@@ -127,8 +127,8 @@ jobs:
 
   deploy:
     name: Deploy
-    if: ${{ !github.event.act }}
-    needs: [build]
+    if: ${{ always() && !github.event.act }}
+    needs:  [build]
     runs-on: ${{ inputs.use_private_agent == true && 'self-hosted' || 'ubuntu-22-04' }}
     environment: ${{ inputs.environment }}-cd
     permissions:
--- a/.github/workflows/web_app_deploy.yaml
+++ b/.github/workflows/web_app_deploy.yaml
@@ -32,6 +32,11 @@ on:
         type: boolean
         required: false
         default: true
+      skip_artifact_creation:
+        description: Skip Artifact creation when is already created by another step
+        type: boolean
+        required: false
+        default: false
 
 concurrency:
   group: ${{ github.workflow }}-cd
@@ -46,6 +51,7 @@ jobs:
     runs-on: ubuntu-22.04
     env:
       WORKSPACE: ${{ inputs.workspace_name }}
+    if: ${{ !inputs.skip_artifact_creation }}
 
     steps:
       - uses: actions/checkout@b4ffde65f46336ab88eb53be808477a3936bae11 # v4.1.1
--- a/.github/workflows/web_app_deploy.yaml
+++ b/.github/workflows/web_app_deploy.yaml
@@ -22,6 +22,16 @@ on:
         type: boolean
         required: false
         default: true
+      health_check_path:
+        description: The health probe path exposed by the Web App.
+        type: string
+        default: ""
+        required: false
+      skip_swap_step:
+        description: Skip slot swap for selective deployment and testing.
+        type: boolean
+        default: false
+        required: false
       use_private_agent:
         description: Use a private agent to deploy the built artifact.
         type: boolean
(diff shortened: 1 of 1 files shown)


.github/workflows/web_app_deploy.yaml
```diff
@@ -22,6 +22,16 @@ on:
         type: boolean
         required: false
         default: true
+      health_check_path:
+        description: The health probe path exposed by the Web App.
+        type: string
+        default: ""
+        required: false
+      skip_swap_step:
```
gunzip, line 30: what about an opt-intinstead of opt-out?

.github/workflows/web_app_deploy.yaml
```diff
@@ -160,8 +170,17 @@ jobs:
            --type zip \
            --async false
 
+      - name: Ping Staging Health
+        if: ${{ inputs.use_staging_slot == 'true' && input.health_check_path != ''}}
```
arcogabbo, line 174:
```suggestion
        if: ${{ inputs.use_staging_slot == 'true' && inputs.health_check_path != ''}}
```

.github/workflows/web_app_deploy.yaml
```diff
@@ -160,8 +170,17 @@ jobs:
...
+        run: |
+          curl \
+            --retry 5 \
+            --retry-max-time 120 \
+            --retry-all-errors \
+            -f 'https://${{ inputs.web_app_name }}-staging.azurewebsites.net${{ inputs.health_check_path }}'
+
       - name: Swap Staging and Production Slots
-        if: ${{ inputs.use_staging_slot == true }}
+        if: ${{ inputs.use_staging_slot == true && input.skip_swap_step == false }}
```
arcogabbo, line 183:
```suggestion
        if: ${{ inputs.use_staging_slot == true && inputs.skip_swap_step == false }}
```

arcogabbo (changes requested):

lucacavallaro:
We intentionally removed the ping from the health check because, with the right environment variables set (`WEBSITE_SWAP_WARMUP_PING_PATH` etc.), it becomes unnecessary.

See https://github.com/pagopa/dx/pull/47#discussion_r1645955762

BurnedMarshal:
> We intentionally removed the ping from the health check because, with the right environment variables set (`WEBSITE_SWAP_WARMUP_PING_PATH` etc.), it becomes unnecessary.
> 
> See [#47 (comment)](https://github.com/pagopa/dx/pull/47#discussion_r1645955762)

Thanks for the information, I miss it. Unfortunately this pipeline has an hidden requirement to work properly and safely. If I'm not wrong only 2 function has the correct settings in their configuration `fims` and `cgn-onboarding-portal-infra`. We should think a way to inform clearly the final users about all the requirements.
For now I'm removing the ping step and the related variable, adding a small comment with the Azure documentation reference.

.github/workflows/web_app_deploy.yaml
BurnedMarshal: Fixed in [65c6818](https://github.com/pagopa/dx/pull/77/commits/65c6818bd914ae9bb60cc32ceada88b62996a63b)

lucacavallaro:
> > We intentionally removed the ping from the health check because, with the right environment variables set (`WEBSITE_SWAP_WARMUP_PING_PATH` etc.), it becomes unnecessary.
> > See [#47 (comment)](https://github.com/pagopa/dx/pull/47#discussion_r1645955762)
> 
> Thanks for the information, I miss it. Unfortunately this pipeline has an hidden requirement to work properly and safely. If I'm not wrong only 2 function has the correct settings in their configuration `fims` and `cgn-onboarding-portal-infra`. We should think a way to inform clearly the final users about all the requirements. For now I'm removing the ping step and the related variable, adding a small comment with the Azure documentation reference.

maybe we can update the function/app services modules to include these variables by default

.github/workflows/web_app_deploy.yaml
```diff
@@ -22,6 +22,11 @@ on:
         type: boolean
         required: false
         default: true
+      swap_slot_step:
+        description: Skip slot swap for selective deployment and testing.
```
gunzip, line 26:
```suggestion
        description: Swap production and staging slot.
```

gunzip:
During our latest meeting, we discussed optimizing our code testing process using slots. We decided to experiment with preview slots—assigning one to each pull request—before finalizing our approach. This should help us determine the most effective strategy moving forward. What do you think about this @BurnedMarshal ?

PR #78, opened 2024-08-21
[DEVEX-165] Added GitHub Action template for CDN Deploy

mamu0:
### List of changes

Added `.github/actions/code_build/action.yaml` reusable action.
Added `.github/actions/cdn_code_deploy/action.yaml` reusable action.
Added `.github/actions/github_modified_files/action.yaml` reusable action.

### Motivation and context

The Actions for the github action that will deploy the **CDN** has been added.
There is the possibility of listing the paths of the files to purge if it is not necessary to do it on `*`, just insert the value `'true' ` ​​in `content_paths`, and the `github_modified_files` action will start and return a list of all files changed in that commit.

A Workflow example can be:
```
name: NodeJS CDN - Template

on:
  workflow_dispatch:

concurrency:
  group: ${{ github.workflow }}-cd
  cancel-in-progress: false

jobs:
  build:
    name: Build
    runs-on: ubuntu-22.04

    steps:
      - uses: actions/checkout@692973e3d937129bcbf40652eb9f2f61becf3332 # v4.1.7
        name: Checkout
      
      - uses: ./.github/actions/code_build
        name: Build Artifact
        with:
          base_workdir: '.'
          zip_artifact: 'false'

  deploy:
    name: Deploy
    if: ${{ !github.event.act }}
    environment: dev-cd
    needs: [build]
    runs-on: ubuntu-22.04
    permissions:
      id-token: write
      contents: read
      actions: read
    env:
      ARM_CLIENT_ID: ${{ secrets.ARM_CLIENT_ID }}
      ARM_TENANT_ID: ${{ secrets.ARM_TENANT_ID }}
      ARM_SUBSCRIPTION_ID: ${{ secrets.ARM_SUBSCRIPTION_ID }}

    steps:
      - uses: actions/checkout@692973e3d937129bcbf40652eb9f2f61becf3332 # v4.1.7
        name: Checkout

      - uses: ./.github/actions/cdn_code_deploy
        name: Deploy Artifact
        with:
          arm_client_id: ${{ env.ARM_CLIENT_ID }}
          arm_tenant_id: ${{ env.ARM_TENANT_ID }}
          arm_subscription_id: ${{ env.ARM_SUBSCRIPTION_ID }}
          resource_group_name: 'io-test-rg-01'
          storage_account_name: 'io-test-sa-01'
          profile_cdn_name: 'io-test-dev-cdn-01'
          endpoint_name: 'io-test-dev-cdn-01'
          blob_container_name: 'io-test-dev-cdn-01'
          content_paths: 'true'
```

### Type of changes

- [x] Add new resources

- [x] No

### Other information

This template need to be tested in a real environment, a possible point of failure can be the `Update Artifact` of the out folder and the initial Prune.


gunzip:
It looks like this isn't a general purpose action to deploy to a CDN, but makes some opinionated guess about the structure of repository (typescript, yarn ,...). If this is the case, we should update module name and PR description to something like "NextJS deploy to CDN"

.github/workflows/js_cdn_deploy.yaml
```diff
@@ -0,0 +1,176 @@
...
+      workspace_name:
+        description: The name of the workspace to create the artifact for.
+        type: string
+        required: true
+      environment:
+        description: Environment where the artifact will be deployed.
+        type: string
+        required: true
+      resource_group_name:
+        description: Web App resource group name.
```
gunzip, line 13: isn't this the CDN resource group?

.github/workflows/js_cdn_deploy.yaml
```diff
@@ -0,0 +1,176 @@
...
+        type: string
+        required: true
+      storage_account_name:
+        description: Storage account name.
+        type: string
+        required: true
+      profile_cdn_name:
+        description: Profile CDN name.
```
gunzip, line 21:
```suggestion
        description: CDN Profile name.
```

gunzip (changes requested):

.github/actions/github_modified_files/action.yaml
```diff
@@ -0,0 +1,43 @@
+name: Get Files Diff
```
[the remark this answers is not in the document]
mamu0: The Action is necessary, because otherwise it cannot be used as a step of a workflow.
The workflow is an addition, if it will never be used we can remove, I left it with the purpose of evaluating in this PR if can be useful to call it individually.

.github/workflows/js_cdn_deploy.yaml
```diff
@@ -0,0 +1,176 @@
...
+
+      - name: Install dependencies
+        run: yarn install --immutable
+        working-directory: ./out
+      
+      # This variable have the form "key=value" e.g.
+      # VAR_1=VALUE_1
+      # VAR_2=VALUE_2
+      # VAR_N=VALUE_N
+      - name: Define env-config.js
```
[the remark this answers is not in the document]
mamu0: It is not necessary, in fact if the variable `custom_env_config_variables` is not specified the step is not executed; it is useful in case you want to dynamically define the `env-config.js` file

.github/actions/github_modified_files/action.yaml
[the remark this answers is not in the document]
gunzip: that's right, but it looks like the workflow does not use the action atm but re-implements the same code

.github/workflows/js_cdn_deploy.yaml
mamu0: yes it is

.github/actions/github_modified_files/action.yaml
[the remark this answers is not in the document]
mamu0: yes you are right, my fault

christian-calabrese (approved):
LGTM

.github/actions/cdn_code_build/action.yaml
```diff
@@ -0,0 +1,87 @@
+name: NodeJS CDN Code Build
```
gunzip, line 1: it looks like this is a generic build action for yarn based project (being them websites or backends). the only peculiar part is the "Define env-config.js" step, which, imho should be delegated to the project build task someway (though npm tasks). 

I'm not convinced that this action adds value, what do you think @lucacavallaro ?

.github/actions/cdn_code_deploy/action.yaml
```diff
@@ -0,0 +1,100 @@
+name: NodeJS CDN Code Deploy
```
gunzip, line 1: is there something in this action that is node specific?

mamu0: no actually it's just a typo brought from the old configuration

.github/actions/cdn_code_build/action.yaml
lucacavallaro: I agree. Usually the ""environment variables"" (in frontend there aren't proper env variables 😛) are managed by the framework - in Next project for example we can commit different `.env.SOMETHING` and pass them as build argument.

mamu0: So do you think it's better to remove this specificity and make the `code_build` part generic?

lucacavallaro: yep (and we can share the `code build` part between backend and frontend projects)

mamu0: maybe we can think about this in a second phase

.github/actions/cdn_code_deploy/action.yaml
```diff
@@ -0,0 +1,100 @@
+name: CDN Code Deploy 
+description: An action Deploy the artifact to the CDN
```
gunzip, line 2:
```suggestion
description: An action Deploy an artifact to a CDN
```

can you add some details about this artifact? ie. must be a zip archive?

.github/actions/cdn_code_deploy/action.yaml
```diff
@@ -0,0 +1,100 @@
...
+    storage_account_name:
+      description: Storage account name.
+      type: string
+      required: true
+    profile_cdn_name:
+      description: CDN Profile name.
+      type: string
+      required: true
+    endpoint_name:
+      description: CDN endpoint name.
```
gunzip, line 19: is this a URL? I'd replace "name" with "url" in case

.github/actions/cdn_code_deploy/action.yaml
```diff
@@ -0,0 +1,100 @@
...
+      type: string
+      required: true
+    blob_container_name:
+      description: Blob container name.
+      type: string
+      required: true
+    content_paths:
+      description: If you want to include only specific file set 'true' else 'false'
```
gunzip, line 27: What do you mean with "specific file" here?

.github/actions/cdn_code_deploy/action.yaml
```diff
@@ -0,0 +1,100 @@
...
+      type: string
+      required: false
+    bundle_name:
+      description: Name of the bundle
```
gunzip, line 31: with or without any (zip) extension?

.github/actions/cdn_code_deploy/action.yaml
```diff
@@ -0,0 +1,100 @@
...
+
+    - name: Sync Storage
+      shell: bash
+      run: |
+        az storage blob sync \
+          -c ${{ inputs.blob_container_name }} \
+          --account-name ${{ inputs.storage_account_name }} \
+          -s "./${{ inputs.bundle_name || 'bundle' }}"
+
+    - name: Found modified files
```
gunzip, line 75: what do you mean with 'modified' here?

.github/actions/github_modified_files/action.yaml
gunzip, line 1: I'd remove the 'github' prefix. Better name: get_changed_files

.github/actions/github_modified_files/action.yaml
```diff
@@ -0,0 +1,43 @@
...
+description: An action to get the list of modified files between the base branch and the current commit.
+
+inputs:
+  base_branch:
+    description: The base branch
```
gunzip, line 6: I'd try to be more descriptive (base branch of what?)

.github/workflows/github_modified_files.yaml
```diff
@@ -0,0 +1,29 @@
+on:
+  workflow_call:
+    inputs:
+      base_branch:
+        description: The base branch
```
gunzip, line 5: ditto

.github/workflows/github_modified_files.yaml
```diff
@@ -0,0 +1,29 @@
...
+        type: string
+        required: false
+        default: main
+
+jobs:
+  get-files-diff:
+    name: Found GitHub files diff
```
gunzip, line 12:
```suggestion
    name: Get changed files
```

gunzip (changes requested):

.github/actions/cdn_code_deploy/action.yaml
mamu0: no it's not a URL, it's a string value for the `az cdn` command called [name](https://learn.microsoft.com/en-us/cli/azure/cdn/endpoint?view=azure-cli-latest#az-cdn-endpoint-purge-optional-parameters)

.github/actions/cdn_code_deploy/action.yaml
mamu0: You can decide to purge the entire folder or only the files that have been added and/or modified

.github/actions/cdn_code_deploy/action.yaml
mamu0: without, but in this case is not a zip

.github/actions/cdn_code_deploy/action.yaml
mamu0: that compared to the previous commit the files are the same, but they have some changes

.github/actions/cdn_code_deploy/action.yaml
```diff
@@ -0,0 +1,107 @@
+name: CDN Code Deploy 
+description: Action that Deploy an artifact to the CDN
+
+# This action allows you to deploy the passed artifact to the CDN.
+# The artifact have to be a directory that contains all the files you want to sync.
```
gunzip, line 5:
```suggestion
# The artifact must be the name of a directory that contains all the files you want to sync.
```

.github/actions/cdn_code_deploy/action.yaml
```diff
@@ -0,0 +1,107 @@
...
+    content_paths:
+      description: If you want to include only specific files set 'true' else 'false'
+      type: string
+      required: false
+    base_branch:
+      description: The branch to compare against the current branch. Defaults to "main" (Not necessary if 'content_paths' is 'false').
+      type: string
+      required: false
+    bundle_name:
+      description: Name of the bundle
```
gunzip, line 38:
```suggestion
      description: Name of the directory containing the files to be synchronized
```

.github/actions/cdn_code_deploy/action.yaml
```diff
@@ -0,0 +1,107 @@
...
+  using: composite
+  steps:
+    - name: Checkout
+      uses: actions/checkout@692973e3d937129bcbf40652eb9f2f61becf3332 # v4.1.7
+
+    - name: Download Artifact
+      uses: actions/download-artifact@6b208ae046db98c579e8a3aa621ab581ff575935 # v4.1.1
+      with:
+        name: ${{ inputs.bundle_name || 'bundle' }}
+        path: "./${{ inputs.bundle_name || 'bundle' }}"
```
gunzip, line 63: what if i want to deploy the root directory?

.github/actions/cdn_code_deploy/action.yaml
```diff
@@ -0,0 +1,107 @@
...
+      with:
+        client-id: ${{ inputs.arm_client_id }}
+        tenant-id: ${{ inputs.arm_tenant_id }}
+        subscription-id: ${{ inputs.arm_subscription_id }}
+
+    - name: Sync Storage
+      shell: bash
+      run: |
+        az storage blob sync \
+          -c ${{ inputs.blob_container_name }} \
```
gunzip, line 78: shouldn't inputs be escaped to avoid command injection?

.github/actions/cdn_code_deploy/action.yaml
```diff
@@ -0,0 +1,107 @@
...
+    endpoint_name:
+      description: CDN endpoint name.
+      type: string
+      required: true
+    blob_container_name:
+      description: Blob container name.
+      type: string
+      required: true
+    content_paths:
+      description: If you want to include only specific files set 'true' else 'false'
```
gunzip, line 30: what do you mean here with "specific files"?

.github/actions/cdn_code_deploy/action.yaml
gunzip: what do you mean with "previous (commit)" here?

.github/actions/cdn_code_deploy/action.yaml
```diff
@@ -0,0 +1,107 @@
...
+
+    - name: Sync Storage
+      shell: bash
+      run: |
+        az storage blob sync \
+          -c ${{ inputs.blob_container_name }} \
+          --account-name ${{ inputs.storage_account_name }} \
+          -s "./${{ inputs.bundle_name || 'bundle' }}"
+
+    - name: Found modified files
```
gunzip, line 82:
```suggestion
    - name: Get changed file names (from previous commit)
```

.github/actions/code_build/action.yaml
```diff
@@ -0,0 +1,128 @@
+name: Code Build
```
gunzip, line 1: I'm not sure this should be related to this PR; I'd leave this PR focues on the CDN deploy

.github/actions/github_modified_files/action.yaml
gunzip: which commits? anyway, I'd use 'git' then instead of `github`

.github/workflows/github_modified_files.yaml
gunzip, line 1: git_changed_files

gunzip (changes requested):
CDN deploy part looks good, but the documentation isn't clear

.github/actions/cdn_code_deploy/action.yaml
mamu0: You can decide to purge the entire folder or only the files that have been added and/or modified

.github/actions/cdn_code_deploy/action.yaml
mamu0: the action that checks the changes sees them between the current commit of the branch in use and the last one of the main, and if we are already in the main it checks the last one of the main and the second to last one

.github/actions/github_modified_files/action.yaml
mamu0: ditto

.github/actions/cdn_code_deploy/action.yaml
mamu0: Now you can't do it directly with that action, you have to create an artifact and then download it in a next step

PR #87, opened 2024-09-02
[CES-12] Added a Keep Alive action to keep the repository active when it contains cron actions

mamu0:
### List of changes

Added `keep-alive` action and `keep_alive_repository` workflow

### Motivation and context

If there are actions in a repository that work with a trigger `schedule`, GitHub will disable them after `60` days if there have been no commits in the repo, in this way we have a workflow that checks every day if `59` days have passed, if so, it does an empty push to revitalize the repository, otherwise it does nothing.

### Type of changes

- [x] Add new resources

- [x] No

### DIFF
--- a/.github/workflows/keep_alive_repository.yml
+++ b/.github/workflows/keep_alive_repository.yml
@@ -1,0 +1,38 @@
+name: Keep Alive
+
+on:
+  workflow_dispatch:
+  schedule:
+    - cron: '0 0 * * *' # Run at 00:00 every day
+  push:
+    branches:
+      - CES-12-keepalive
+
+jobs:
+  keep_alive:
+    runs-on: 'ubuntu-20.04'
+    permissions:
+      contents: read
+      actions: write
+
+    steps:
+      - name: Checkout
+        uses: actions/checkout@b4ffde65f46336ab88eb53be808477a3936bae11 # v4.1.7
+
+      - name: Calculate days since last commit
+        id: commit_date
+        run: |
+          LAST_COMMIT_DATE=$(git log -1 --format=%ct)
+          echo "Last commit date: $LAST_COMMIT_DATE"
+
+          CURRENT_DATE=$(date +%s)
+          echo "Current date: $CURRENT_DATE"
+
+          DIFFERENCE=$(( ($CURRENT_DATE - $LAST_COMMIT_DATE) / 86400 ))
+          echo "Days since last commit: $DIFFERENCE"
+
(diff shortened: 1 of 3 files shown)


.github/workflows/keep_alive_repository.yml
```diff
@@ -0,0 +1,35 @@
...
+          CURRENT_DATE=$(date +%s)
+          echo "Current date: $CURRENT_DATE"
+
+          DIFFERENCE=$(( ($CURRENT_DATE - $LAST_COMMIT_DATE) / 86400 ))
+          echo "Days since last commit: $DIFFERENCE"
+
+          echo "days_since_commit=$DIFFERENCE" >> $GITHUB_ENV
+
+      - name: Keep Alive
+        if: steps.commit_date.outputs.days_since_commit == '59'
```
gunzip, line 34: can you add a comment on this number?

.github/workflows/keep_alive_repository.yml
```diff
@@ -0,0 +1,35 @@
+name: Keep Alive
```
gunzip, line 1: can you add a comment that describes the purpose?

gunzip:
lgtm, let's wait for other's comments

christian-calabrese (approved):
LGTM

.github/workflows/keep_alive_repository.yml
```diff
@@ -0,0 +1,43 @@
...
+
+          # Calculate how many days have passed since the last commit
+          DIFFERENCE=$(( ($CURRENT_DATE - $LAST_COMMIT_DATE) / 86400 ))
+          echo "Days since last commit: $DIFFERENCE"
+
+          echo "days_since_commit=$DIFFERENCE" >> $GITHUB_ENV
+
+      - name: Keep Alive
+        # If 59 days have passed then execute the action which makes an empty push
+        if: steps.commit_date.outputs.days_since_commit >= '59'
```
Krusty93, line 42:
```suggestion
        if: steps.commit_date.outputs.days_since_commit >= '55'
```

I'd put some gap here, like 55, because the cron expression uses UTC timezone while our commits are on GMT+2.
So, commits at 01:00 are equivalent to 22/23:00 of the previous day (I presume `git log` returns the UTC timezone.

The risk is really minimum, but we prevent any weird condition that may happen

mamu0: Yes, no problem, a little more range can't hurt. 
The `log -1` command returns the time in `UNIX` format.

gunzip (approved):

PR #89, opened 2024-09-03
fix: fix pessimistic constraint on azure role assignment module

christian-calabrese:
### List of changes

Reduced pessimistic constraint being to strict on azure role assignment module

### Motivation and context

The constraint was forcing to use 3.106.* azure provider versions

### Type of changes

- [ ] Add new resources

- [ ] No


mamu0 (approved):
LGTM

PR #92, opened 2024-09-05
Add Doc for Azure Login mechanism

Krusty93:
### List of changes

Add documentation to explain how Azure login works and the suggested patterns to use

### Type of changes

- [ ] Add new resources

- [ ] No


website/docs/Docs/pipelines/azure-login.md
gunzip, line 1: path should be docs/pipelines/azure-login.md

Krusty93: We already have the `Docs` path:
image

gunzip: that's the first one (docs)

website/docs/pipelines/azure-login.md
```diff
@@ -0,0 +1,106 @@
...
+# Azure Login
+
+There are several ways to log into Azure by a GitHub Action. The our chosen method doesn't require maintenance or secret management as it is a passwordless approach. Once set up, it works.
+
+Only two components are needed to log into Azure:
+
+- An Azure Managed Identity resource
+- A federation between the Azure Managed Identity and GitHub
+
+The folder `infra/identity` contains the Terraform definition, often created by Cloud Engineers using a [Terraform custom module](https://github.com/pagopa/dx/tree/main/infra/modules/azure_federated_identity_with_github) to create both the Managed Identity and the federation with GitHub. Module also provides a set of default roles in the current subscription, which are likely to be enough for new repositories - more on this later.
```
gunzip, line 15: why 'often'? in which cases is not created by cloud eng?

gunzip: how are you supposed to engage cloud eng andh when?

website/docs/pipelines/azure-login.md
```diff
@@ -0,0 +1,106 @@
...
+
+Once the Managed Identity is created, only `client id` value and note it apart.
+
+Once done, you are technically ready to log into Azure from a GitHub pipeline:
+
+```yaml
+- name: Azure Login
+  uses: azure/login@v2
+  with:
+    client-id: <managed identity client id>
```
gunzip, line 25: I'd rather put the secrets placeholder

website/docs/pipelines/azure-login.md
```diff
@@ -0,0 +1,106 @@
...
+
+Then, `subscription id` and `managed identity client id` values can be stored as secrets of a specific GitHub environment. In particular:
+
+- `tenant id`: the value is always the same and could be stored as repository secret
+- `subscription id`: if the project has a single environment it could be stored as repository secret; otherwise as environment secret
+- `managed identity client id`: always as environment secret
+
+:::tip
+
+This setup is generally defined via Terraform in the `infra/repository` folder.
```
gunzip, line 57: this statement is not very clear

gunzip (commented):
the explanation looks good, but I'd suggest to add a enumerated list of things to do ie.
1) dothis...
2) then that...
even interleaved into the content

Krusty93:
> the explanation looks good, but I'd suggest to add a enumerated list of things to do ie.
> 
> 1. dothis...
> 2. then that...
>    even interleaved into the content

It wasn't my intention to write a step by step guide to achieve a goal, but instead explaining how things work (at higher level), what's the purpose of the components and write down the suggested patterns about GitHub environments.

If we want to add some practical thing to do, I would add the steps to setup the `repository` folder. However, I am afraid it could end up with something really similar to what we already have on Confluence.

Your thoughts or any suggestion?

website/docs/pipelines/azure-login.md
lucacavallaro: @gunzip implementing `infra/identity` is trivial and no support from `cloud engineers` is needed, cloud eng should maintain the linked module

gunzip: I'd remove 'often created by Cloud Engineers' in this case

Krusty93: The ideal path would include the Terraform identity definition written by developers, and the PR review and the Terraform apply done by cloud engs (devs can create identities but not giving them roles)

website/docs/pipelines/azure-login.md
Krusty93: Updated

PR #93, opened 2024-09-05
[DEVEX-137] Imported alerts in Azure Functions App module

Krusty93:
### List of changes

Add availability alert on Function App and Storage Account components in Azure Function App module.

This PR adds an optional variable `alert_config`, which brings the `action_group_id` to notify when the alert triggers. Alerts are created only in `premium` and `standard` tiers.

To avoid breaking changes, developers should manually add the `alert_config` indicating their `action_group_id`, otherwise the alert is created without actions. We can also change the approach introducing this small breaking change, but I don't think it is the right choice as informing developers about it should be enough (same code of legacy module). Instead, we can alter the default behaviour in the next weeks with a minor blast radius as many of them will have already implemented it.

### Motivation and context

Production resources should be monitored. Moreover, these alerts were already included in the legacy modules

### Type of changes

- [ ] Add new resources

- [ ] No

### DIFF
--- a/infra/modules/azure_function_app/locals.tf
+++ b/infra/modules/azure_function_app/locals.tf
@@ -1,7 +1,8 @@
 locals {
-  location_short = var.environment.location == "italynorth" ? "itn" : var.environment.location == "westeurope" ? "weu" : var.environment.location == "germanywestcentral" ? "gwc" : "neu"
-  project        = "${var.environment.prefix}-${var.environment.env_short}-${local.location_short}"
-  domain         = var.environment.domain == null ? "-" : "-${var.environment.domain}-"
+  location_short  = var.environment.location == "italynorth" ? "itn" : var.environment.location == "westeurope" ? "weu" : var.environment.location == "germanywestcentral" ? "gwc" : "neu"
+  project         = "${var.environment.prefix}-${var.environment.env_short}-${local.location_short}"
+  domain          = var.environment.domain == null ? "-" : "-${var.environment.domain}-"
+  app_name_prefix = "${local.project}${local.domain}${var.environment.app_name}"
 
   subnet = {
     enable_service_endpoints = var.subnet_service_endpoints != null ? concat(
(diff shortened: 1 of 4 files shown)


infra/modules/azure_function_app/variables.tf
```diff
@@ -155,3 +155,12 @@ variable "subnet_service_endpoints" {
   description = "(Optional) Enable service endpoints for the underlying subnet. This variable should be set only if function dependencies do not use private endpoints"
   default     = null
 }
+
+variable "alert_config" {
+  type = object({
+    action_group_id    = string
+    webhook_properties = map(string)
```
gunzip, line 162: what's the expected format for webhook_properties? can we add an example?

Krusty93: I think it can be removed, nobody has ever used it: https://github.com/search?q=org%3Apagopa+webhook_properties+language%3AHCL&type=code&l=HCL

Krusty93: removed

gunzip (approved):

mamu0 (approved):
LGTM

christian-calabrese (approved):

infra/modules/azure_function_app/locals.tf
```diff
@@ -1,29 +1,31 @@
 locals {
-  location_short = var.environment.location == "italynorth" ? "itn" : var.environment.location == "westeurope" ? "weu" : var.environment.location == "germanywestcentral" ? "gwc" : "neu"
-  project        = "${var.environment.prefix}-${var.environment.env_short}-${local.location_short}"
-  domain         = var.environment.domain == null ? "-" : "-${var.environment.domain}-"
+  location_short  = var.environment.location == "italynorth" ? "itn" : var.environment.location == "westeurope" ? "weu" : var.environment.location == "germanywestcentral" ? "gwc" : "neu"
```
gunzip, line 2: can't we use a unique map shared by all modules instead of replicating this logic in every module?

```hcl
variable "environment" {
  type = object({
    location = string
  })
}

locals {
  location_map = {
    "italynorth"        = "itn"
    "westeurope"        = "weu"
    "germanywestcentral" = "gwc"
    # Default (se la location non è presente nella mappa)
    "default"           = "neu"
  }
}

location_short = lookup(local.location_map, var.environment.location, local.location_map["default"])

```

PR #95, opened 2024-09-05
[DEVEX-165] Added GitHub Action template for CDN Deploy

mamu0:
### List of changes

Added `.github/actions/cdn_code_deploy/action.yaml` reusable action.

### Motivation and context

The Action that will deploy the **CDN** has been added.
There is the possibility of listing the paths of the files to purge if it is not necessary to do it on `*`, just adding the list into `selective_purge_paths` input.
You can call only the action for the final CDN deploy, remember that the action accept as inputs `sync_dir_name` and `working_directory`, used to point the right folder containing files to sync into the blob storage.

**NOTE**: 

- the artifact is a directory and NOT a zip file.
- you have to do `Azure Login` before run this action.

E.g. a possible job can be
```
jobs:
  deploy-cdn:
    name: CDN Deploy
    runs-on: ubuntu-22.04
    environment: dev-cd
    permissions:
        ...
    env:
        ...

    steps:
      - name: Checkout
        ...

      - name: Download Artifact
        ...

      - name: Unzip Artifact (If is not a directory)
        ...

      - name: Azure Login
        ...

      - uses: ./.github/actions/cdn_code_deploy
        name: CDN Deploy
        with:
          working_directory: '.'
          sync_folder_name: 'bundle'
          resource_group_name: 'io-test-rg-01'
          storage_account_name: 'io-test-sa-01'
          profile_name: 'io-test-dev-cdn-01'
          endpoint_name: 'io-test-dev-cdn-01'
          blob_container_name: 'io-test-dev-cdn-01'
          selective_purge_paths: "./path/file1 ./path/file2"
```

### Type of changes

- [x] Add new resources

- [x] No


.github/actions/artifact_definition/action.yaml
```diff
@@ -0,0 +1,77 @@
...
+          yarn workspaces focus --production
+
+          FORMAT=$(jq -r 'if .type == "module" then "esm" else "cjs" end' package.json)
+          SHORT_SHA=$(git rev-parse --short ${{ github.sha }})
+
+          mkdir ${{ github.run_id }}
+
+          # bundle compiled code, excluding node_modules 
+          curl -fsSL https://esbuild.github.io/dl/v0.21.5 | sh
+          ./esbuild index=$ENTRY_POINT --bundle --format=$FORMAT --platform=node --target=node20 --packages=external --outdir=${{ github.run_id }} --allow-overwrite
```
gunzip, line 50: are you sure we want to bulde code using `esbuild` before deploying this to a cdn?
usually web frameworks like NextJS use their embedded bundling method

.github/actions/artifact_definition/action.yaml
```diff
@@ -0,0 +1,77 @@
...
+    type: string
+    required: false
+
+runs:
+  using: composite
+  steps:
+    - uses: actions/checkout@692973e3d937129bcbf40652eb9f2f61becf3332 # v4.1.7
+      name: Checkout
+
+    - name: Build the artifact
```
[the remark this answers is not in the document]
mamu0: The idea behind creating multiple actions instead of a specific workflow was to build several reusable modules. However, if we notice too many specificities in the actions, I would suggest considering returning to a single workflow approach, such as [web_app_deploy](https://github.com/pagopa/dx/blob/main/.github/workflows/web_app_deploy.yaml).

.github/actions/artifact_definition/action.yaml
mamu0: ditto

.github/workflows/cdn_app_deploy.yaml
```diff
@@ -0,0 +1,139 @@
...
+      endpoint_name:
+        description: CDN endpoint name.
+        type: string
+        required: true
+      blob_container_name:
+        description: Blob container name.
+        type: string
+        required: true
+      selective_purge_paths:
+        description: |
```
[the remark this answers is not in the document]
mamu0: We could improve the speed of the `purge` step if the person running it knows that only a few files need to be updated and knows their paths

gunzip (commented):
is there

.github/actions/cdn_code_deploy/action.yaml
```diff
@@ -0,0 +1,86 @@
...
+      shell: bash
+      run: |
+        az cdn endpoint purge \
+          -g ${{ inputs.resource_group_name }} \
+          -n ${{ inputs.endpoint_name }} \
+          --profile-name ${{ inputs.profile_name }} \
+          --content-paths $(echo "${{ inputs.selective_purge_paths }}" | sed -e "s/\([^ ]*\)/'\1'/g")
+
+    # If selective_purge_paths is not defined, purge all files '/*'
+    - name: Purge CDN
```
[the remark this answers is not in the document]
mamu0: At the moment I don't think there is an automatic mechanism for this, but I see two possible ways:

- Redo the purge with a previous version of the code so as to reload the old files.
- From the Azure Portal go to restore the previous version of the files on the storage account (here you need to make sure that blob versioning is enabled)

The second option can also be done with the command `az storage blob copy start`, but it's a bit tricky.

.github/actions/cdn_code_deploy/action.yaml
```diff
@@ -0,0 +1,86 @@
...
+  profile_name:
+    description: CDN Profile name.
+    type: string
+    required: true
+  endpoint_name:
+    description: CDN endpoint name.
+    type: string
+    required: true
+  selective_purge_paths:
+    description: |
```
gunzip, line 29: does this supports glob patterns / dir or only single files?

.github/actions/cdn_code_deploy/action.yaml
```diff
@@ -0,0 +1,86 @@
...
+      Define a list of paths to include only new/modified files in the purge. Leave empty to include all files.
+      Paths must be relative to the synchronized directory (sync_dir_name).
+      E.g. if syncing ./bundle, to include ./bundle/path/file1 and ./bundle/path/file2, specify ./path/file1 ./path/file2
+      NOTE: Insert paths separated by spaces
+    type: string
+    required: false
+  sync_dir_name:
+    description: The artifact that is the name of the directory containing the files to be synchronized (Default is 'bundle')
```
gunzip, line 37: I'd default to 'dist'

gunzip (commented):
lgtm apart minor changes

.github/actions/cdn_code_deploy/action.yaml
mamu0: if you want to point to a whole directory you have to end the path with `/*` because the purge always refers to files

.github/actions/cdn_code_deploy/action.yaml
```diff
@@ -0,0 +1,86 @@
...
+  endpoint_name:
+    description: CDN endpoint name.
+    type: string
+    required: true
+  selective_purge_paths:
+    description: |
+      Define a list of paths to include only new/modified files in the purge. Leave empty to include all files.
+      Paths must be relative to the synchronized directory (sync_dir_name).
+      E.g. if syncing ./bundle, to include ./bundle/path/file1 and ./bundle/path/file2, specify ./path/file1 ./path/file2
+      NOTE: Insert paths separated by spaces
```
gunzip, line 33:
```suggestion
  List of relative paths for files to include in the purge. Leave empty to purge all files.
  Paths should be relative to the synchronized directory (sync_dir_name).
  For example, if syncing ./bundle, to include ./bundle/path/file1 and ./bundle/path/file2, specify ./path/file1 ./path/file2.
  
  To recursively purge an entire directory, use the pattern ./path/*.```

File: .github/actions/cdn_code_deploy/action.yaml [hunk 8]
```diff
@@ -0,0 +1,86 @@
+name: CDN Code Deploy
```
Krusty93, line 1: Why `Code`? Is this not generic to deploy anything?

.github/actions/cdn_code_deploy/action.yaml
```diff
@@ -0,0 +1,86 @@
+name: CDN Code Deploy 
+description: Action that Deploy an artifact to the CDN
+
+# This action allows you to deploy the passed artifact to the CDN.
```
Krusty93, line 4: Do we need a specific action for artifact deployments? Can't be the same of a generic file deployment? What code changes?

gunzip (approved):

.github/actions/cdn_code_deploy/action.yaml
```diff
@@ -0,0 +1,89 @@
...
+      Paths should be relative to the synchronized directory (sync_dir_name).
+      For example, if syncing ./dist, to include ./dist/path/file1 and ./dist/path/file2, specify "./path/file1 ./path/file2."
+      To recursively purge an entire directory, use the pattern "./path/dir/*".
+
+      The format should follow the parameters as described in the official documentation: 
+      https://learn.microsoft.com/en-us/cli/azure/cdn/endpoint?view=azure-cli-latest#az-cdn-endpoint-purge-required-parameters
+    type: string
+    required: false
+  sync_dir_name:
+    description: The asset that is the name of the directory containing the files to be synchronized (Default is 'dist')
```
Krusty93, line 40:
```suggestion
    description: Path to the files directory (Default is 'dist')
```
Moreover, I'd add the `default` field with `dist` as value

.github/actions/cdn_code_deploy/action.yaml
```diff
@@ -0,0 +1,89 @@
...
+    - name: Checkout
+      uses: actions/checkout@692973e3d937129bcbf40652eb9f2f61becf3332 # v4.1.7
+
+    ## Sync Storage
+    # will be synced the to the working_directory/sync_dir_name defined (Default will be './dist')
+    - name: Sync Storage
+      shell: bash
+      run: |
+        az storage blob sync \
+          -c ${{ inputs.blob_container_name }} \
```
Krusty93, line 60: I suggest to use the entire command name in scripts, in order to improve readability

gunzip: shouldn't inputs be escaped?

.github/actions/cdn_code_deploy/action.yaml
```diff
@@ -0,0 +1,89 @@
...
+
+      The format should follow the parameters as described in the official documentation: 
+      https://learn.microsoft.com/en-us/cli/azure/cdn/endpoint?view=azure-cli-latest#az-cdn-endpoint-purge-required-parameters
+    type: string
+    required: false
+  sync_dir_name:
+    description: The asset that is the name of the directory containing the files to be synchronized (Default is 'dist')
+    type: string
+    required: false
+  working_directory:
```
Krusty93, line 43: Is this necessary? Can't we ask for the entire folder path in the `sync_dir_name` variable?

mamu0: I think it’s simpler for a user to specify the directory name and the working directory separately, as it improves clarity and reduces confusion when setting up paths

Krusty93 (approved):

PR #96, opened 2024-09-06
[DEVEX-175] Fix infra deployment pipeline result status in case of failure

Krusty93:
### List of changes

Add `set -o pipefail` flag to the bash script that runs the `terraform apply` command

### Motivation and context

Regression, pipeline were green (passed) rather than red (failed)

### Type of changes

- [ ] Add new resources

- [ ] No


christian-calabrese (approved):
LGTM

PR #97, opened 2024-09-06
[DEVEX-158] Improve pre-commit hook

Krusty93:
### List of changes

Update pre-commit-terraform to 1.94.1
Fixed an issue where if multiple extensions were passed to `--files` argument, pre-commit skipped all steps for some reason. Now only `tf` files are passed
Added suggested step to "fix" a dependency issue of the base image container
Run pre-commit over everything if there are no `tf` files changed
Differentiate the behavior of `terraform_docs` for modules and resources, whether including or not provider info
`terraform_docs` creates `README.md` files if they don't exist
Introduced `terraform_trivy` with its `.trivyignore` file

Fixed README.md according to new rules
Fixed lock files by adding missing SHAs

### Motivation and context

Previous implementation had some issue while verifying files at PR creation and dependencies were quite old. Security checks were missing as well.

### Type of changes

- [ ] Add new resources

- [x] No

### DIFF
--- a/.github/workflows/static_analysis.yaml
+++ b/.github/workflows/static_analysis.yaml
@@ -31,7 +31,7 @@ jobs:
         uses: actions/checkout@692973e3d937129bcbf40652eb9f2f61becf3332 # v4.1.7
         with:
           fetch-depth: 0
-          ref: ${{ github.sha }}
+          # ref: ${{ github.sha }}
 
       # https://github.com/antonbabenko/pre-commit-terraform#github-actions
       - name: Git config
@@ -51,7 +51,7 @@ jobs:
           echo "Diff between ${{ github.base_ref }} and ${{ github.sha }}"
           echo "files=$(echo \"$DIFF\" | xargs echo)" >> $GITHUB_OUTPUT
 
-      - name: fix tar dependency in alpine container image
+      - name: Fix Alpine Container Image
         run: |
           apk --no-cache add tar
           # check python modules installed versions
@@ -71,7 +71,7 @@ jobs:
           pre-commit run \
             --color=always \
             --show-diff-on-failure \
-            --from-ref ${{ github.base_ref }} --to-ref ${{ github.sha }}
+            --from-ref ${{ github.base_ref }} --to-ref ${{ github.head_ref }}
             # --files "${{ steps.file_changes.outputs.files }}"
 
       - name: Run pre-commit
--- a/.github/workflows/static_analysis.yaml
+++ b/.github/workflows/static_analysis.yaml
@@ -62,11 +62,11 @@ jobs:
           # check python modules installed versions
           python -m pip freeze --local
 
-      - name: Cache pre-commit
-        uses: actions/cache@0c45773b623bea8c8e75f6c82b208c3cf94ea4f9 # v4.0.2
-        with:
-          path: ~/.cache/pre-commit
-          key: pre-commit-3|${{ hashFiles('.pre-commit-config.yaml') }}
+      # - name: Cache pre-commit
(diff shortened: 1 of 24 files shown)


.github/workflows/static_analysis.yaml
```diff
@@ -20,7 +25,7 @@ jobs:
     env:
       TERRAFORM_VERSION: "1.7.5"
     container:
-      image: ghcr.io/antonbabenko/pre-commit-terraform:v1.86.0@sha256:a1ffb8fff155934c1937580eeab8b3f68726485cd54025d2f911b0c21a9becba
+      image: ghcr.io/antonbabenko/pre-commit-terraform:v1.94.1
```
gunzip, line 28: what about version pinning here?

.github/workflows/static_analysis.yaml
```diff
@@ -47,18 +52,18 @@ jobs:
...
+          DIFF="$(git diff --name-only origin/${{ github.base_ref }} origin/${{ github.head_ref }} -- '*.tf')"
+          echo "Diff between ${{ github.base_ref }} and ${{ github.head_ref }}"
           echo "files=$(echo \"$DIFF\" | xargs echo)" >> $GITHUB_OUTPUT
 
-      - name: Cache pre-commit
-        uses: actions/cache@0c45773b623bea8c8e75f6c82b208c3cf94ea4f9 # v4.0.2
-        with:
-          path: ~/.cache/pre-commit
-          key: pre-commit-3|${{ hashFiles('.pre-commit-config.yaml') }}
+      - name: Fix Alpine Container Image
```
gunzip, line 59: why it's not cached anymore?

.pre-commit-config.yaml
```diff
@@ -1,11 +1,33 @@
...
-    rev: v1.86.0
+    rev: v1.94.1
     hooks:
       - id: terraform_fmt
       - id: terraform_docs
+        name: terraform_docs on modules
+        args:
+          - --hook-config=--create-file-if-not-exist=true
+          - --args=--hide providers
+        exclude: |
```
gunzip, line 11: can we switch to `include infra/modules` pattern instead?

.trivyignore
```diff
@@ -0,0 +1,8 @@
+# Github repository shouldn’t be public.
```
gunzip, line 1: can you add a link with the trivy related docs?

.github/workflows/static_analysis.yaml
Krusty93: I didn't notice any performance boost, it seemed quite useless

.github/workflows/static_analysis.yaml
Krusty93: Added

.trivyignore
Krusty93: Do you mean for each suppressed warning or trivy in general?

.pre-commit-config.yaml
Krusty93: probably using `[files](https://pre-commit.com/#top_level-files)`, but I have to give it a try

mamu0: Yes, if we want to specify that the only folder to check is modules, we can use `files: ^infra/modules/.*`, which will automatically exclude everything else

.trivyignore
gunzip: in general (with a list of possible configuration)

Krusty93: Updated with the main page of misconfiguration docs

gunzip (approved):

christian-calabrese (approved):

PR #100, opened 2024-09-12
Set turbo cache path

giamma1295:
Set turbo cache dir using  `TURBO_CACHE_DIR ` env variable.

https://turbo.build/repo/docs/reference/system-environment-variables#:~:text=in%20rare%20circumstances.-,TURBO_CACHE_DIR,-Sets%20the%20cache

### Type of changes

- [ ] Add new resources

- [ ] No

### DIFF
--- a/.github/workflows/docs_website_publish.yaml
+++ b/.github/workflows/docs_website_publish.yaml
@@ -27,15 +27,28 @@ jobs:
       - name: Setup yarn
         run: corepack enable
 
-      - uses: actions/setup-node@v4
+      - uses: actions/setup-node@1e60f620b9541d16bece96c5465dc8ee9832be0b #v4.0.3
         with:
           node-version-file: ".node-version"
           cache: yarn
 
+      - name: Determine turbo cache path
+        id: determine-turbo-version
+        run: |
+          TURBO_VERSION=$(yarn turbo --version)
+          echo "Turbo version: $TURBO_VERSION"
+          if [[ "$TURBO_VERSION" =~ ^1\..* ]]; then
+            TURBO_CACHE_PATH="node_modules/.cache/turbo"
+          else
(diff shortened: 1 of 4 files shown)


.github/workflows/docs_website_publish.yaml
```diff
@@ -27,15 +27,28 @@ jobs:
...
+        run: |
+          TURBO_VERSION=$(yarn turbo --version)
+          echo "Turbo version: $TURBO_VERSION"
+          if [[ "$TURBO_VERSION" =~ ^1\..* ]]; then
+            TURBO_CACHE_PATH="node_modules/.cache/turbo"
+          else
+            TURBO_CACHE_PATH=".turbo"
+          fi
+          echo "TURBO_CACHE_PATH=$TURBO_CACHE_PATH" >> $GITHUB_ENV
+          echo "Using cache path: $TURBO_CACHE_PATH"
```
lucacavallaro, line 46: Instead of determining the right turbo cache path, we can set a fixed one for all runs (using `TURBO_CACHE_DIR` environment variable)

lucacavallaro (changes requested):

gunzip:
do you think it's worth to encapsulate the node + turbo setup in its own action in order to avoid code duplication? @lucacavallaro @giamma1295

.github/workflows/docs_website_publish.yaml
giamma1295: https://github.com/pagopa/dx/pull/100/commits/2d75dfa63f857e6eb85f019b5954b06577a1333b

gunzip (approved):

lucacavallaro (approved):

gunzip:
Thank you @giamma1295 for this first external contribution to DX! 🥳

PR #102, opened 2024-09-14
[DEVEX-135] - Assets needed for terraform registry enablement

christian-calabrese:
### List of changes

- GitHub Workflow that pushes changes of single modules to the module's subrepository
- GitHub Workflow for Changeset releases
- Added package.json to all modules
- Added script to initialize a new module's folder and (optionally) github subrepository

### Motivation and context
The objective of this PR is to enable the terraform modules of this repository to be ready to be published on the terraform registry.

#### Pre-commit
To avoid the replication of the lock-modules.sh script in all the repositories, a pre-commit has been developed in this PR.
An example of a `.pre-commit-config.yaml` file that uses it:

```
repos:
  - repo: https://github.com/pagopa/dx
    rev: terraform-registry-scaffolding
    hooks:
      - id: lock_modules
        args:
          - src/common/prod
          - src/core/prod
```
More info:
- The rev field will be then populated with a tag or main
- The args field contains the list of folders where terraform configurations are present

For more information, these RFCs enter the details of the idea:

1. [Adoption of a terraform registry](https://pagopa.atlassian.net/wiki/spaces/DevEx/pages/1086554273/RFC-DX-010+Adozione+di+un+terraform+registry)
2. [Adapting DX modules to be published on the public terraform registry](https://pagopa.atlassian.net/wiki/spaces/DevEx/pages/1245937830/RFC-DX-012+Adattare+i+moduli+DX+per+pubblicazione+sul+registry+pubblico)

### Type of changes

- [ ] Add new resources

- [X] No


gunzip:
Since we are creating new repositories within the GitHub organization, we should decide whether to use a prefix (e.g., dx) to clearly distinguish DX modules from others.

christian-calabrese:
> Since we are creating new repositories within the GitHub organization, we should decide whether to use a prefix (e.g., dx) to clearly distinguish DX modules from others.

@gunzip yes, i guess we can add a simple "dx" suffix to the name of the repository in the script. I wouldn't change the name of the module's folders to avoid redundancy and breaking changes for current users

.github/workflows/relase_modules.yml
```diff
@@ -0,0 +1,46 @@
...
+            fetch-depth: 0
+
+        # Corepack is an official tool by Node.js that manages package managers versions
+        - name: Setup yarn
+          run: corepack enable
+
+        - name: Setup Node.js environment
+          uses: actions/setup-node@60edb5dd545a775178f52524783378180af0d1f8 # v4.0.2
+          with:
+            node-version: 18
```
kin0992, line 31: **question**: why not using the version of the [`.node-version` file](https://github.com/pagopa/dx/blob/main/.node-version)?

The [`setup-node` action allows you to use the `node-version-file` option](https://github.com/actions/setup-node/blob/main/docs/advanced-usage.md#node-version-file), which gets the node version from the specified file.

.github/workflows/relase_modules.yml
```diff
@@ -0,0 +1,46 @@
...
+            cache: "yarn"
+
+        - name: Install dependencies
+          run: yarn
+
+        - name: Create PR for next release
+          uses: changesets/action@f13b1baaa620fde937751f5d2c3572b9da32af23 # v1.4.5
+          with:
+              version: yarn run version
+              # publish: npm run tag
```
kin0992, line 41: **suggestion(non-blocking)**: I'd remove this
```suggestion
```

.github/actions/push-module/action.yaml
```diff
@@ -0,0 +1,83 @@
...
+        git config --global user.name "GitHub Actions"
+
+        # Configure Git for authentication
+        git config --global url."https://api:${{ inputs.github_pat }}@github.com/".insteadOf "https://github.com/"
+        git config --unset http."https://github.com/".extraheader
+    - name: Push Subtree to Module Repo
+      shell: bash
+      env:
+        MODULE_NAME: ${{ inputs.module }}
+        MODULE_VERSION: ${{ steps.get_version.outputs.module_version }}
```
gunzip, line 50: where does `outputs.module_version` come from?

.github/workflows/push_modules_to_subrepo.yml
```diff
@@ -0,0 +1,49 @@
...
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v4
+        with:
+          fetch-depth: 0
+      - name: Push module to subrepository
+        id: push-to-subrepository
+        uses: ./.github/actions/push-module
+        with:
+          repo: ${{ github.repository_owner }}/terraform-azurerm-${{ matrix.module }}
```
gunzip, line 47: don't we need to add the `dx` suffix here?

gunzip: (why not using a prefix instead?)

.github/actions/push-module/action.yaml
```diff
@@ -0,0 +1,83 @@
...
+
+        # Extract the complete history of the module subrepo
+        git subtree split --prefix=infra/modules/${MODULE_NAME} -b ${MODULE_NAME}-branch
+
+        # Check if the main branch exists on the remote
+        git fetch module-repo main --tags
+
+        # Merge the changes without overwriting the history
+        git checkout -b temp-branch module-repo/main
+        git merge --allow-unrelated-histories ${MODULE_NAME}-branch
```
gunzip, line 63: what about handling errors? ie

```suggestion
git merge --allow-unrelated-histories ${MODULE_NAME}-branch || { echo "Merge failed"; exit 1; }
```

or using set -euo pipefail

infra/scripts/add-module.sh
```diff
@@ -0,0 +1,87 @@
...
+
+  # Create the GitHub repository under the specified organization
+  echo "Creating GitHub repository '$SUBREPO_NAME' in organization '$ORG_NAME'..."
+  gh repo create "$ORG_NAME/$SUBREPO_NAME" --public --confirm
+
+  if [ $? -eq 0 ]; then
+    echo "GitHub repository created successfully: https://github.com/$ORG_NAME/$SUBREPO_NAME"    # Initialize Git in the module directory and push to the new repository
+    cd "$MODULE_DIR"
+    git init
+    git remote add origin git@github.com:$ORG_NAME/$SUBREPO_NAME.git
```
gunzip, line 75:
```suggestion
    git remote add origin git remote add origin https://github.com/$ORG_NAME/$SUBREPO_NAME.git
```

.github/workflows/push_modules_to_subrepo.yml
```diff
@@ -0,0 +1,49 @@
...
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v4
+        with:
+            fetch-depth: 2 # Ensure we have enough history for the diff
+
+      - name: Get modified modules
+        id: get-modules
+        run: |
+          folders_list=$(git diff --no-commit-id --name-only HEAD~1 infra/modules | xargs -r dirname | sed 's#^infra\/modules/\([^/]*\).*#\1#' | uniq | jq -R -c . | jq -s -c .)
```
gunzip, line 26: what about something like...

```suggestion
folders_list=$(git diff --name-only ${{ github.event.before }} ${{ github.sha }} infra/modules | ...)
```

.github/workflows/push_modules_to_subrepo.yml
```diff
@@ -0,0 +1,49 @@
...
+          echo Modules that have changed: $folders_list
+          echo "modules=$folders_list" >> $GITHUB_OUTPUT
+    outputs:
+      modules: ${{ steps.get-modules.outputs.modules }}
+
```
gunzip, line 31:
```suggestion
- name: Check if modules list is empty
  if: ${{ steps.get-modules.outputs.modules == '[]' }}
  run: echo "No modules changed. Skipping push."
```

.github/workflows/relase_modules.yml
gunzip, line 1: typo: file name should be 'release'

infra/scripts/add-module.sh
```diff
@@ -0,0 +1,87 @@
...
+# Check if the module name was provided
+if [ -z "$MODULE_NAME" ]; then
+  echo "Error: No module name provided."
+  usage
+fi
+
+PROVIDER="azurerm" # we may support different providers as aws, awscc, etc.
+SUFFIX="dx"
+SUBREPO_NAME="terraform-$PROVIDER-$SUFFIX-$MODULE_NAME"
+MODULE_DIR="modules/$MODULE_NAME"
```
gunzip, line 28: isn't

`MODULE_DIR="infra/modules/$MODULE_NAME"`

?

.github/workflows/push_modules_to_subrepo.yml
gunzip, line 47: shouldn't the provider part 'azurerm' be configurable?

infra/scripts/add-module.sh
```diff
@@ -0,0 +1,87 @@
...
+
+# Parse named arguments
+while [[ "$#" -gt 0 ]]; do
+  case $1 in
+    --name) MODULE_NAME="$2"; shift ;;
+    --gh-org) ORG_NAME="$2"; shift ;;
+    *) echo "Unknown parameter passed: $1"; usage ;;
+  esac
+  shift
+done
```
gunzip, line 17:
```suggestion
while [[ "$#" -gt 0 ]]; do
  case $1 in
    --name) MODULE_NAME="$2"; shift ;;
    --gh-org) ORG_NAME="$2"; shift ;;
    --provider) PROVIDER="$2"; shift ;;
    *) echo "Unknown parameter passed: $1"; usage ;;
  esac
  shift
done

PROVIDER="${PROVIDER:-azurerm}"

```

.github/actions/push-module/action.yaml
christian-calabrese: It's actually a typo, this step isn't used anymore. Thank you!

gunzip:
to fully enable regitry usage we need to use https://github.com/ned1313/terrahash/ around init in out pipelines

.github/workflows/push_modules_to_subrepo.yml
christian-calabrese: Actually the name has been changed from suffix to prefix because it will be put in front of the actual module name.
We cannot put the "dx" prefix before terraform or the provider name cause we would break the requirements of the terraform registry

.github/workflows/relase_modules.yml
christian-calabrese: There's already a release workflow for changeset which has a couple of differences.
@lucacavallaro do you think we can merge the two files in one without breaking what you did initially in the release.yaml workflow?

christian-calabrese: Actually, if the package.json has private: true set, the publish step shouldn't do anything. ([source](https://github.com/changesets/changesets/blob/main/docs/config-file-options.md#ignore-array-of-packages))

For this reason, the existing `release` workflow has been only slightly updated with the suggested changes above.

infra/scripts/lock-modules.sh
```diff
@@ -0,0 +1,60 @@
+#!/bin/bash
+set -euo pipefail
```
gunzip, line 2: Can we add this to the pre-commit step?

christian-calabrese: The pre-commit configuration has been just implemented and described in the description of the present PR 🎉

gunzip:
Is there any impact on existing pipelines refering this action?

christian-calabrese:
> Is there any impact on existing pipelines refering this action?

The new step only checks for modules downloaded from the terraform registry. For this reason, it shouldn't cause any problem for repositories where the registry is not used.

gunzip (approved):

mamu0 (approved):
LGTM

.github/workflows/push_modules_to_subrepo.yml
```diff
@@ -0,0 +1,51 @@
...
+      - name: Checkout repository
+        uses: actions/checkout@v4
+        with:
+          fetch-depth: 0
+      - name: Push module to subrepository
+        id: push-to-subrepository
+        uses: ./.github/actions/push-module
+        with:
+          module: ${{ matrix.module }}
+          github_pat: ${{ secrets.BOT_GH_TOKEN }}
```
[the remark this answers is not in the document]
christian-calabrese: The issue is that the action needs to push on other repositories. Do you know if the built in token can do it?

.github/actions/push-module/action.yaml
```diff
@@ -0,0 +1,81 @@
...
+        git config --global --add safe.directory /github/workspace
+
+        # Set git email and name
+        git config --global user.email "actions@github.com"
+        git config --global user.name "GitHub Actions"
+
+        # Configure Git for authentication
+        git config --global url."https://api:${{ inputs.github_pat }}@github.com/".insteadOf "https://github.com/"
+        git config --unset http."https://github.com/".extraheader
+    - name: Push Subtree to Module Repo
```
Krusty93, line 40:
```suggestion

    - name: Push Subtree to Module Repo
```

Krusty93 (approved):

PR #116, opened 2024-09-25
[CES-122] Configured tier variable for new SKU definition

mamu0:
### List of changes

This PR updates the tier definitions in the following resources:  `azure_app_service`, `azure_app_service_expose`, `azure_function_app` and `azure_function_app_exposed`.
Previously, the tiers were defined as test, standard, and premium. These have now been replaced by the new tier system: `xs`, `s`, `m`, `l`, and `xl`, providing a more granular selection for scaling resources according to workload demands.

Only for `function_app` modules we have a new environment variable: `FUNCTIONS_WORKER_PROCESS_COUNT`: that based on tier change values in this way:

- `xs/s` = 8 GB -> WPC 2
- `m` = 16GB -> WPC 4
- `l/xl` = 32GB -> WPC 8

### Motivation and context

This change is required to support a wider range of performance requirements beyond the previously available premium tier. The new tier system (`xs`, `s`, `m`, `l`, and `xl`) offers increased flexibility in managing the resource power based on the application’s workload, allowing users to scale more effectively.

### Type of changes

- [ ] Add new resources

- [x] No

### Other information

This update ensures backward compatibility with the old tier system. If a user applies test, standard, or premium, these values will be automatically mapped to the new tiers (s, m, and l respectively). This prevents disruptions for users who haven’t yet transitioned to the new tier nomenclature.


infra/modules/azure_function_app_exposed/locals.tf
```diff
@@ -4,16 +4,22 @@ locals {
   domain          = var.environment.domain == null ? "-" : "-${var.environment.domain}-"
   app_name_prefix = "${local.project}${local.domain}${var.environment.app_name}"
 
+  # For backwards compatibility
+  # If no legacy value is passed, use the original value
+  # TO DO: Remove this in the next major release and replace all local.tier with var.tier
```
gunzip, line 9: every "todo" must have a reference to some jira issue (id), otherwise they become _neverdo_

infra/modules/azure_app_service/locals.tf
```diff
@@ -15,9 +20,9 @@ locals {
   }
```
gunzip, line 21: what about something like this istead? the general concept is to use map wherever possible instead of ternaries which are harder to read and maintain:

```
# Tier mappings for better scalability and readability
 tier_mapping = {
   "test"     = "s"
   "standard" = "m"    
   "premium"  = "l"
   "xs"       = "xs"
   "s"        = "s"
   "m"        = "m"
   "l"        = "l"
   "xl"       = "xl"
 }

 # Default to the mapped tier or retain the original tier if not found
  
tier = lookup(local.tier_mapping, var.tier, var.tier)

 # Worker process count based on tier
 worker_process_count = {
   "s"  = 2
   "m"  = 4
   "l"  = 8
   "xl" = 8
   "xs" = 1
 }[local.tier]
```

infra/modules/azure_app_service/locals.tf
```diff
@@ -15,9 +20,9 @@ locals {
...
 
   app_service = {
-    sku_name               = var.tier == "test" ? "B1" : var.tier == "standard" ? "P0v3" : "P1v3"
-    zone_balancing_enabled = var.tier != "test"
-    is_slot_enabled        = var.tier == "test" ? 0 : 1
+    sku_name               = local.tier == "s" ? "B1" : local.tier == "m" ? "P0v3" : local.tier == "l" ? "P1v3" : local.tier == "xl" ? "P2v3" : "B1"
+    zone_balancing_enabled = local.tier != "s"
+    is_slot_enabled        = local.tier == "s" ? 0 : 1
```
gunzip, line 25:
```suggestion
sku_name               = lookup({
      "xs" = "B1"
       "s"  = "B1"
      "m"  = "P0v3"
       "l"  = "P1v3"
        "xl" = "P2v3"
       }, local.tier, "B1")  
    zone_balancing_enabled = local.tier != "s"
    is_slot_enabled        = local.tier == "s" ? 0 : 1
```

gunzip: even better: create a centralized `tiers.tf `file to define all tier-related variables and mappings and use them inside every locals.tf ie:

`sku_name               = local.sku_name_mapping[local.tier]`

with `tiers.tf`

```
+/**
+ * Tier Definitions
+ * Centralized mapping for legacy and new tiers
+ */
+
+locals {
+  # Tier mappings for better scalability and readability
+  tier_mapping = {
+    "test"     = "s"
+    "standard" = "m"
+    "premium"  = "l"
+    "xs"       = "xs"
+    "s"        = "s"
+    "m"        = "m"
+    "l"        = "l"
+    "xl"       = "xl"
+  }
+
+  # Default to the mapped tier or retain the original tier if not found
+  tier = lookup(local.tier_mapping, var.tier, var.tier)
+
+  # Worker process count based on tier
+  worker_process_count = {
+    "xs" = 1
+    "s"  = 2
+    "m"  = 4
+    "l"  = 8
+    "xl" = 8
+  }[local.tier]
+
+  # SKU name mapping
+  sku_name_mapping = {
+    "xs" = "B1"
+    "s"  = "B1"
+    "m"  = "P0v3"
+    "l"  = "P1v3"
+    "xl" = "P2v3"
+  }
+}

```

gunzip (changes requested):
nice job, but I'd refactor using maps instead of ternaries

gunzip (approved):

infra/modules/azure_function_app/tier.tf
```diff
@@ -0,0 +1,37 @@
...
+    "m"  = "P0v3"
+    "l"  = "P1v3"
+    "xl" = "P2mv3"
+  }
+
+  # Worker process count based on tier
+  worker_process_count_mapping = {
+    "xs" = 2
+    "s"  = 2
+    "m"  = 4
```
Krusty93, line 33: This shall be 1 (4GB -> 1 worker x 4GB)
```suggestion
    "m"  = 4
```

infra/modules/azure_app_service/tier.tf
```diff
@@ -0,0 +1,28 @@
...
+    "l"        = "l"
+    "xl"       = "xl"
+  }
+
+  # Default to the mapped tier
+  tier = local.tier_map[var.tier]
+
+  # SKU name mapping
+  sku_name_mapping = {
+    "xs" = "B1"
```
Krusty93, line 22: If everyone agrees, I'd put F1 (very rare but better than two duplicated values)
```suggestion
    "xs" = "F1"
```

infra/modules/azure_function_app/tier.tf
mamu0: ditto

infra/modules/azure_app_service/tier.tf
mamu0: it is fine for me

infra/modules/azure_function_app/tier.tf
```diff
@@ -0,0 +1,37 @@
...
+    "l"        = "l"
+    "xl"       = "xl"
+  }
+
+  # Default to the mapped tier
+  tier = local.tier_map[var.tier]
+
+  # SKU name mapping
+  sku_name_mapping = {
+    "xs" = "B1"
```
Krusty93, line 22:
```suggestion
    "xs" = "F1"
```
ditto

infra/modules/azure_function_app/tier.tf
```diff
@@ -0,0 +1,37 @@
...
+    "s"  = "B1"
+    "m"  = "P0v3"
+    "l"  = "P1v3"
+    "xl" = "P2mv3"
+  }
+
+  # Worker process count based on tier
+  worker_process_count_mapping = {
+    "xs" = 2
+    "s"  = 2
```
Krusty93, line 32:
```suggestion
    "s"  = 1
```
B1 has 1.75GB, so this shall be 1 (1 worker x 4GB=4FB)

infra/modules/azure_function_app_exposed/tier.tf
```diff
@@ -0,0 +1,37 @@
...
+  sku_name_mapping = {
+    "xs" = "B1"
+    "s"  = "B1"
+    "m"  = "P0v3"
+    "l"  = "P1v3"
+    "xl" = "P2mv3"
+  }
+
+  # Worker process count based on tier
+  worker_process_count_mapping = {
```
Krusty93, line 30: ditto

infra/modules/azure_function_app/tier.tf
```diff
@@ -0,0 +1,37 @@
...
+    "m"  = 4
+    "l"  = 8
```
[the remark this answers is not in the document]
gunzip: cc @BurnedMarshal

infra/modules/azure_app_service/tier.tf
mamu0: As I did it is already compatible with the current version, in fact it is not necessary to refactor everything that has already been created with the old form (test, standard and premium).

Regarding the proposal I think it could be a bit confusing, because the association

> s = test
> m = standard
> l = premium

and `xl` was necessary to add a more advanced `sku`, at most if there is no use case we can remove `xs` which is now "extra", but i think that could be useful in the future in my opinion.

infra/modules/azure_function_app/tier.tf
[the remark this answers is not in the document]
BurnedMarshal: `xs` and `s` are unstable becouse they have less than 4 GB of memory, set 1 for them anyway.
`m` is P0V3 with 4GB of ram, set 1.
`l` is P1V3 with 8 GB, set 2
`xl`is P2mv3 with 32 GB, set 8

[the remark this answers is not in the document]
mamu0: @Krusty93 If i've understand what you mean the right table is:

- xs = 1 (because F1 have 1 GB)
- s = 1 (because B1 have 1,75 GB)
- m = 1 (because P0v3 have 3,5 GB)
- l = 1 (because P1v3 have 3,5 GB)
- xl = 2 (because P2v3 have 7 GB)

correct?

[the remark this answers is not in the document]
Krusty93: 2 workers for P1v3, as there are 8GBs in there
4 workers for P2V3, as it has 16GB

Krusty93 (approved):

gunzip:
very good ❤️

infra/modules/azure_function_app/tier.tf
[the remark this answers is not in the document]
BurnedMarshal: We don't have in functions P2V3, but P2mv3 that has 32 GB of ram, so 8 worker are the right value :D

PR #119, opened 2024-10-01
Add test users module

mamu0:
### List of changes

Added module for test users

### Type of changes

- [x] Add new resources

- [x] No

### DIFF
--- a/infra/modules/test_users/README.md
+++ b/infra/modules/test_users/README.md
@@ -1,0 +1,44 @@
+# DX Test Users Module
+
+## Description
+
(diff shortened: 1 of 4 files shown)


gunzip (commented):
is this general purpose or target IO? My guess is that this belongs to -io-infra repo

mamu0:
> is this general purpose or target IO? My guess is that this belongs to -io-infra repo

I think @Krusty93 can answer you more precisely

gunzip (changes requested):
it looks like this does not belong here

PR #122, opened 2024-10-03
[DEVEX-190] Generalized Static Analysis action

mamu0:
### List of changes

`static_analysis.yaml` edited to be callable action
`static_analysis_dx.yaml` created specific call for this repository

### Motivation and context

The static analysis action has been modified to be callable from any other repository, and the action on the right that calls it has been created to use it as it has always been

### Type of changes

- [x] Add new resources

- [x] No

### DIFF
--- a/.github/workflows/static_analysis_dx.yaml
+++ b/.github/workflows/static_analysis_dx.yaml
@@ -1,0 +1,21 @@
+name: Terraform Validation
+
+on:
+  workflow_dispatch:
+  pull_request:
+    types: [opened, synchronize]
+    paths:
+      - infra/**
+      - .terraform-version
+      - .pre-commit-config.yaml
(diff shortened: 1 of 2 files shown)


.github/workflows/static_analysis_dx.yaml
```diff
@@ -0,0 +1,21 @@
...
+      - .trivyignore
+
+jobs:
+  tf_analysis:
+    uses: ./.github/workflows/static_analysis.yaml
+    name: Terraform Validation
+    secrets: inherit
+    with:
+      terraform_version: 1.7.5
+      pre_commit_tf_version: 1.94.1
```
Krusty93, line 20: The sha as well need to be passed other than the simple version

Krusty93 (approved):

PR #123, opened 2024-10-07
DX prefix in modules subrepos changed and package.json added to missing modules

christian-calabrese:
### Type of changes

- [ ] Add new resources

- [ ] No


mamu0 (approved):

Krusty93 (approved):

mamu0:
I see that the validate is failing, I think it's because now the terraform version is at 1.9.5 and in the various modules it is blocked because there is `~>`, maybe you can update it

PR #124, opened 2024-10-07
Added examples for modules and removed terraform version

mamu0:
### Type of changes

- [ ] Add new resources

- [ ] No


gunzip (commented):
Absolutely love this work! Quick question: is there a way to avoid repeating information in `locals.tf` files? It’s almost identical across all modules.

christian-calabrese (approved):
LGTM!

mamu0:
> Absolutely love this work! Quick question: is there a way to avoid repeating information in `locals.tf` files? It’s almost identical across all modules.

Thank you! Yes changing locals is a main point of another task

PR #129, opened 2024-10-07
Added pre-commit hook for terraform provider execution

mamu0:
### List of changes

Added a new script and pre commit hook that can be called.

### Type of changes

- [ ] Add new resources

- [ ] No


christian-calabrese (approved):
LGTM

PR #133, opened 2024-10-09
[CES-169] Add whitelist configuration for Event Hub module

mamu0:
### List of changes

Updated eventhub module with allowed sources configuration.

### Motivation and context

Because is necessary to access the eventhub from other sources we need to whitelist them.

### Type of changes

- [ ] Add new resources

- [x] No


infra/modules/azure_event_hub/variables.tf
```diff
@@ -63,6 +63,15 @@ variable "subnet_pep_id" {
...
 
+variable "allowed_sources" {
+  type = list(object({
+    subnet_ids = list(string)
+    ips        = list(string)
+  }))
+  default     = []
+  description = "List of allowed sources (subnets and ip masks)"
+}
+
```
christian-calabrese, line 74: Do we actually need a list of objects with lists inside?
Maybe the object is sufficient
```suggestion
variable "allowed_sources" {
  type = object({
    subnet_ids = optional(list(string), [])
    ips        = optional(list(string), [])
  })
  default     = {}
  description = "List of allowed sources (subnets and ip masks)"
}

```

gunzip: it depdends. if subnet and ips are related is better to use a list of objects, otherwise an object with two lists suffices

mamu0: I think @christian-calabrese is right, making a list of objects can create confusion for those who configure the variable, better an object with just the two lists inside for subnets and ips

Krusty93 (approved):

PR #135, opened 2024-10-11
[FIX] Eventhub whitelist variable now have optional values

mamu0:
### List of changes

Variable `allowed_sources`
### Motivation and context

Added optional values to avoid mandatory empty string

### Type of changes

- [ ] Add new resources

- [x] No


christian-calabrese (approved):
LGTM

PR #136, opened 2024-10-11
[CES-193] Add XXL tier to Function App module

Krusty93:
### List of changes

Add XXL tier to Function App modules, to exploit the `P3mv3` SKU

### Motivation and context

Wallet's Function App needs more vCPU to perform some critical cpu-intensive operations. We noticed performance increases using this SKU, which is reserved for use cases like this.

The number of worker process reached its limit - 10. Then, no more SKU will be added. In case of performance thirsty, the current ones will be replaced by the bigger ones

### Type of changes

- [ ] Add new resources

- [ ] No

### DIFF
--- a/infra/modules/azure_function_app/examples/complete/main.tf
+++ b/infra/modules/azure_function_app/examples/complete/main.tf
@@ -25,7 +25,7 @@ module "azure_function_app" {
   source = "../../"
 
   environment         = local.environment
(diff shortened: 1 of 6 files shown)


christian-calabrese (approved):
LGTM

mamu0 (approved):

PR #138, opened 2024-10-14
fix: get naming convention module from git

christian-calabrese:
### List of changes

Get the naming convention module from git instead of local path

### Motivation and context

When we'll download the DevEx terraform modules from the terraform registry, it will break due to the missing "dependency" model azure_naming_convention.
For this reason, it's important to get this dependent module from a remote source (in our case, github or the registry when set up)

image

### Type of changes

- [ ] Add new resources

- [X] No


gunzip:
can't we publish the `naming_convention` module first and get it from terraform registry?

christian-calabrese:
> can't we publish the `naming_convention` module first and get it from terraform registry?

I think the alternatives to download the naming convention module from other modules are:
- from the terraform registry without specifying a tag => I think the script that checks hashes would break at any release and it's probably not a good practice
- from the main of the dx repo => Not very elegant
- from the terraform registry pinning always the last version => I'm not really a fan of what we used to do in the past with the storage account module

gunzip:
why do you need to pin the latest version? from the terraform registry you can point to a major version and get free updates

gunzip:
> I think the script that checks hashes would break at any release

that's the expected behaviour if you don't update the lock file

christian-calabrese:
This work has already been executed in another PR

PR #141, opened 2024-10-18
[DEVEX-192] Add support to use a custom location for GitHub Federated Identities

Krusty93:
### List of changes

Add location variable

### Motivation and context

Moving away from WestEurope, previously hardcoded in the module dependencies (will be broken soon)

### Type of changes

- [ ] Add new resources

- [ ] No


christian-calabrese (approved):
LGTM

PR #144, opened 2024-10-22
[CES-199] - Avoid modules dependencies from existing resources

christian-calabrese:
### Type of changes

- [ ] Add new resources

- [ ] No

### DIFF
--- a/infra/modules/azure_app_service_plan_autoscaler/variables.tf
+++ b/infra/modules/azure_app_service_plan_autoscaler/variables.tf
@@ -16,13 +16,54 @@ variable "autoscale_name" {
 
 variable "target_service" {
   type = object({
-    app_service_name  = optional(string)
-    function_app_name = optional(string)
+    app_service_id       = optional(string)
+    app_service_name     = optional(string)
+    function_app_id      = optional(string)
+    function_app_name    = optional(string)
+    app_service_plan_id  = optional(string)
+    location = optional(string)
   })
+
+  description = <<EOT
+  Configuration for the target service (App Service or Function App) to which the autoscaler will be applied.
+  
+  For existing services:
+  - Provide either 'app_service_name' or 'function_app_name'.
+  
+  For new services being created concurrently:
+  - Provide 'app_service_id' or 'function_app_id' (depending on the service type).
+  - 'app_service_plan_id' and 'location' must also be provided.
+
+  Note: Only one service type (App Service or Function App) should be specified.
+  EOT
 
   validation {
     condition     = (var.target_service.app_service_name != null) != (var.target_service.function_app_name != null)
-    error_message = "Only one between \"app_service_name\" and \"function_app_name\" can have a value. It is not possible to set both of them \"null\"."
+    error_message = "Exactly one of 'app_service_name' or 'function_app_name' must be provided for existing services."
+  }
+
+  validation {
+    condition     = (var.target_service.app_service_id != null) != (var.target_service.function_app_id != null)
+    error_message = "Exactly one of 'app_service_id' or 'function_app_id' must be provided for new services."
+  }
+
+  validation {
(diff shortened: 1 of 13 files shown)


infra/modules/azure_app_service_plan_autoscaler/variables.tf
```diff
@@ -16,13 +16,54 @@ variable "autoscale_name" {
...
+    function_app_name    = optional(string)
+    app_service_plan_id  = optional(string)
+    location = optional(string)
   })
 
+  description = <<EOT
+  Configuration for the target service (App Service or Function App) to which the autoscaler will be applied.
+  
+  For existing services:
+  - Provide either 'app_service_name' or 'function_app_name'.
```
gunzip, line 31: What about having an app service autoscaler module AND a function autoscaler module instead? you may consider to share most of the code but this configuration.

gunzip: another option is to something like this instead:

```hcl
variable "target_service" {
  type = object({
    app_service = optional(object({
      app_service_id      = string
      app_service_name    = string
      app_service_plan_id = string
      location            = string
    }))
    function_app = optional(object({
      function_app_id   = string
      function_app_name = string
      location          = string
    }))
  })

  validation {
    condition = length([for v in [var.target_service.app_service, var.target_service.function_app] : v if v != null]) == 1
    error_message = "You must define either app_service or function_app, but not both."
  }
}
```

infra/modules/azure_app_service_plan_autoscaler/variables.tf
```diff
@@ -16,13 +16,54 @@ variable "autoscale_name" {
...
+  
+  For new services being created concurrently:
+  - Provide 'app_service_id' or 'function_app_id' (depending on the service type).
+  - 'app_service_plan_id' and 'location' must also be provided.
+
+  Note: Only one service type (App Service or Function App) should be specified.
+  EOT
+
   validation {
```
gunzip, line 40: see comment above to simplify this logic. I think this would be a breaking change anyway.

infra/modules/azure_role_assignments/modules/cosmos/variables.tf
```diff
@@ -6,7 +6,8 @@ variable "principal_id" {
 variable "cosmos" {
   description = "A list of CosmosDB role assignments"
   type = list(object({
-    account_name        = string
+    account_name        = optional(string)
+    account_id          = optional(string)
```
gunzip, line 10: should be documented into var description

infra/modules/azure_role_assignments/modules/event_hub/variables.tf
```diff
@@ -4,47 +4,48 @@ variable "principal_id" {
...
+    event_hub_names     = optional(list(string))
+    event_hub_ids       = optional(list(string))
     role                = string
   }))
 
+  description = <<EOT
+  List of Event Hub configurations for role assignments.
+  For existing Event Hubs:
+    - Provide namespace_name and event_hub_names
+  For new Event Hubs being created concurrently:
```
gunzip, line 20: what do you mean here with "new Event Hubs being created concurrently"?

infra/modules/azure_role_assignments/variables.tf
```diff
@@ -6,7 +6,8 @@ variable "principal_id" {
 variable "cosmos" {
   description = "A list of CosmosDB role assignments"
   type = list(object({
-    account_name        = string
+    account_name        = optional(string)
+    account_id          = optional(string)
```
gunzip, line 10: should be documented

gunzip (changes requested):

christian-calabrese:
Closed cause the issue has been tackled with single PRs

PR #145, opened 2024-10-24
[DEVEX-180] Added test configuration for each modules

mamu0:
### List of changes

Created a `tests` folder for each module in the `dx` repository, containing specific test configurations for each module. The structure for all of them is as follows: 
within each module, there is a tests folder containing a subfolder for each specific configuration (usually named setup, with the resources to be created), and a file with the `.tftest.hcl` extension that executes the various verification steps.
```
modules
└── example_module
    └── tests
          ├── setup
          └── main.tf
                └── example_module.tftest.hcl
```

### Motivation and context

This PR was necessary to streamline the testing phase for resources when creating or modifying a Terraform module, making it easier for automation, which will be introduced in a follow-up PR. 
For the correct test execution you need to use the commands

- `terraform init`
- `terraform test`

If you want to see more detail use `-verbose`

### Type of changes

- [ ] Add new resources

- [x] No


christian-calabrese (approved):

infra/modules/azure_cosmos_account/examples/complete/main.tf
```diff
@@ -25,7 +25,7 @@ module "cosmos_db" {
 
   secondary_geo_locations = [
     {
-      location          = "italynorth"
+      location          = "westeurope"
```
gunzip, line 28: why this change?

gunzip (commented):
this is huge! 🚀

infra/modules/azure_cosmos_account/examples/complete/main.tf
mamu0: because there was an error, the primary and secondary region cannot match

PR #155, opened 2024-11-11
Allow storage account's system assigned identity use CMK

christian-calabrese:
### List of changes
Allow the storage account's system assigned identity to use the CMK without a user assigned one
Creating the customer managed key kv key automatically
Adding relevant permissions to the right managed identity to use the customer managed key
Allow the usage of pre-existing key vault keys as customer managed keys

### Motivation and context
Some storage accounts that are being migrated from WEU to ITN make use of the system assigned managed identity for CMK setup

### Type of changes

- [ ] Add new resources

- [X] No


infra/modules/azure_storage_account/variables.tf
```diff
@@ -42,12 +42,12 @@ variable "customer_managed_key" {
...
     type                      = optional(string, null)
-    key_name                  = optional(string)
+    key_name                  = optional(string, null)
     user_assigned_identity_id = optional(string, null)
-    key_vault_key_id          = optional(string, null)
+    key_vault_id              = optional(string, null)
   })
-  description = "(Optional) Customer managed key to use for encryption. Currently type can only be set to 'kv'."
-  default     = { enabled = false, key_name = null }
+  description = "(Optional) Customer managed key to use for encryption. Currently type can only be set to 'kv'. If the key vault is in the same account, and key_name is not set, the key and relevant permissions will be automatically created."
```
Krusty93, line 49: `tenant` rather than `account`?

infra/modules/azure_storage_account/cmk.tf
```diff
@@ -0,0 +1,38 @@
+# tfsec:ignore:azure-keyvault-ensure-key-expiry
+resource "azurerm_key_vault_key" "key" {
+  for_each     = (local.cmk_flags.kv && var.customer_managed_key.key_name == null ? toset(["kv"]) : toset([]))
+  name         = "${replace("${module.naming_convention.prefix}-st-${module.naming_convention.suffix}", "-", "")}-cmk-kv"
```
Krusty93, line 4: What do you think of a format like `io-p-itn-<domain>-[<app_name>]-st-cmk-01`? I took it from the current PEP format as `io-p-itn-wallet-sql-pep-01`

mamu0 (approved):
LGTM

Krusty93:
I might have already asked - what are these storage accounts? why do they use managed identities? what key do they use?

infra/modules/azure_storage_account/locals.tf
```diff
@@ -38,4 +38,18 @@ locals {
       dns_zone = "privatelink.table.core.windows.net"
     }
   }
+
+  cmk_flags = {
+    kv = (var.customer_managed_key.enabled && var.customer_managed_key.type == "kv")
+  }
+
+  cmk_info = {
+    kv = local.cmk_flags.kv ? {
```
Krusty93, line 47: here you can use new azurerm provider functions! 
Quite easy: https://registry.terraform.io/providers/hashicorp/azurerm/latest/docs/guides/4.0-upgrade-guide#example-usage

Krusty93 (approved):

infra/modules/azure_storage_account/locals.tf
christian-calabrese: That would be perfect! My only thought is for who is using dx modules with the azurerm v3 provider
If we introduce a major it would mean that we should force every user to the v4 to make use of the feature and any new coming fix

gunzip: what's the proposed solution to the breaking change?

PR #174, opened 2024-12-04
[CES-441] Added Storage Account Queue resource definition

mamu0:
### List of changes

Added `azurerm_storage_queue` resource and `queue_enabled` variable

### Motivation and context

to allow the creation of a queue in a storage account, the appropriate terraform resource has been added, and a variable to enable/disable the creation, default to `false`

### Type of changes

- [ ] Add new resources

- [x] No


Krusty93 (commented):
What's the purpose to have a Storage Queue in this module?

gunzip:
what if we want more than one queue?

gunzip:
If we can implement a queue for each file, it would likely integrate smoothly with code generation.

gunzip:
If the module would merely result in a wrapper we don't need it and the added flag on the storage account itself suffices.

mamu0:
> If the module would merely result in a wrapper we don't need it and the added flag on the storage account itself suffices.

With the latest commit, I’ve attempted to provide more context and purpose to this resource. Let me know if it would be useful, otherwise, we can stick with creating resources only when needed. @pagopa/engineering-team-cloud-eng @gunzip

lucacavallaro:
@mamu0 I don't find this change useful, as `azurerm_storage_queue` resource is already easy to implement

PR #175, opened 2024-12-05
Added vars to postgresql module

michaeldisaro:
### List of changes
Added vars:
- replica_enabled
- geo_redundant_backup_enabled
- high_availability_enabled
- auto_grow_enabled

Used these vars to enables replica and backup features instead of using tier to calculate them.

### Motivation and context
We need to explicitly enabled backup and replica features instead of having them calculated from the tier.

### Type of changes

- [ ] Add new resources

- [ ] No


Krusty93 (commented):
What is blocking/concerning you to set the tier with required features?

michaeldisaro:
> What is blocking/concerning you to set the tier with required features?

Something is simply not working in ITN, for example backup geo-redundancy, and something is not wanted, for example the replica. 
These vars allows for more granularity and choice.

gunzip:
We must constraint granularity and choices to predefined tiers anyway.

gunzip (changes requested):
We must constraint granularity and choices to predefined tiers anyway.

michaeldisaro:
> We must constraint granularity and choices to predefined tiers anyway.

Ok, so I'll migrate to ITN in QX 2025+ to wait for geo-redundant backups in ITN. That's fine.

michaeldisaro:
> We must constraint granularity and choices to predefined tiers anyway.

https://github.com/pagopa/dx/pull/175/commits/f5aca60106ab48b3923570f2060353930b7eef3b

infra/modules/azure_postgres_server/variables.tf
```diff
@@ -101,6 +101,30 @@ variable "replica_zone" {
...
+variable "replica_enabled" {
+  type        = bool
+  description = "(Optional) Enables replica for PostgreSQL Flexible Server."
+  default     = false
+}
+
+variable "geo_redundant_backup_enabled" {
+  type        = bool
+  description = "(Optional) Enables geo redundancy for PostgreSQL Flexible Server backups."
+  default     = false
```
gunzip, line 113: this is a breaking change if you put 'false' here. should be opt-out to keep consistent behavior

gunzip (changes requested):

infra/modules/azure_postgres_server/variables.tf
michaeldisaro: suggest the change

PR #176, opened 2024-12-05
[CES-440] Added new name creation method for all resources

mamu0:
### List of changes

Edited Naming convention module with a new name generation method proposal

### Motivation and context

To avoid possible human changes during naming creation for all resources inside naming convention module now exist a name output that is composed of a usual prefix and suffix, but with resource type abbreviation in the middle
E.g. If resource type is `"Storage Account"` in italynorth, subscription `PROD-IO` and domain `modules-test` we will have `io-p-itn-modules-test-st-01`

### Type of changes

- [ ] Add new resources

- [x] No


infra/modules/azure_naming_convention/variables.tf
```diff
@@ -40,3 +40,78 @@ variable "environment" {
 
   description = "Values which are used to generate resource names and location short names. They are all mandatory except for domain, which should not be used only in the case of a resource used by multiple domains."
 }
+
+variable "resource_type" {
```
gunzip, line 44: why do we need this? can't we set this value inside tf modules?

mamu0: the module needs to know somehow what kind of resource it is, what do you mean by tf modules?

gunzip: Probably I'm missing something. Ehen you define some resource you either

a) use some terraform provider resource ie. `azurerm_cosmosdb_account`. if you have to also specify `resource_type = "Cosmos DB"` doesn't this results in a duplication ?
b) use some dx module which can specify `resource_type` itself

infra/modules/azure_naming_convention/locals.tf
```diff
@@ -13,4 +13,62 @@ locals {
...
+    "Virtual Network"        = "vnet"
+    "Network Security Group" = "nsg"
+    "Application Gateway"    = "agw"
+    "Load Balancer"          = "lb"
+    "Public IP Address"      = "pip"
+    "DNS Zone"               = "dns"
+
+    # Databases
+    "SQL Database"         = "sql"
+    "Cosmos DB"            = "cosmo"
```
gunzip, line 44: Shouldn't it be `cosmos` or `cosno`?

infra/modules/azure_naming_convention/variables.tf
mamu0: Ah, okay, I think I understand what you mean. Yes, that makes sense, but a module doesn’t always create just the specific resource it is intended for. 
For example, some modules also create private endpoints, subnets, and storage accounts, so, if we used only the module as a reference, we wouldn’t be able to assign names to all the other resources

gunzip: I intended to handle the two cases differently (if possible):

a) Attempt to derive the resource type directly from the Terraform resource, avoiding the need for the user to specify the resource type manually.

b) Apply (a) for DX modules, or alternatively, include the resource type within the DX module resources ourselves.

mamu0: I have an idea on how we could improve this functionality. We could pass the module name as an input and generate all the required values as an output. 
For example, if we pass the module name `azure_cosmos_account`, the output could have a structure like this:
```
{
  "cosmos": "...",
  "private_endpoint": "..."
}
```
This way, each module would define its own set of names to be used

Alternatively, we could simplify it further by always generating the output with all possible abbreviations, and then the relevant one can be referenced. 
For example: `module.naming_convention.name.cosmos`

mamu0: Yes, that’s correct. However, the issue I see is that each module creates multiple resources with different names. For example, if the module context suggests it is for an `app_service`, it would need to generate names for the `app_service`, `private_endpoints`, `service_plan`, and `subnet`. This introduces additional complexity. By using a map with all the values as I’ve done now, the module can simply reference the appropriate names where needed (e.g., for `azurerm_app_service`, the `name` will be `module.naming_convention.names.app_service`).

.changeset/eighty-cows-wave.md
```diff
@@ -0,0 +1,5 @@
+---
+"azure_naming_convention": patch
+---
+
+Added new name creation method with resource type input and name output
```
Krusty93, line 5:
```suggestion
Updated name generation which includes resource type input and add those names in output
```

Krusty93 (commented):
Very efficient to simplify naming convention adoption 👏🏻

infra/modules/azure_naming_convention/README.md
```diff
@@ -1,5 +1,36 @@
 # DX - Common Variables Validation
 
+This module generates output values that follow the naming conventions used in DX. It takes the following inputs:
```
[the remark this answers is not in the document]
mamu0: This will be resolved with multiple environments, look at this [PR](https://github.com/pagopa/dx/pull/179), in this way you can configure different environments, so with other `app_names` or without `domains`

Krusty93:
Duplicated #179

PR #179, opened 2024-12-09
[CES-574] Update naming convention module with support for multiple environments

mamu0:
### List of changes

The naming_convention module has been updated with a new input that now accepts a list of environments. It can return various outputs to construct names or the names output, which will contain all possible names for each Azure resource.

### Motivation and context

This change was made to allow developers to generate names more conveniently, without needing to instantiate the module multiple times for multiple environments or instances.

### Type of changes

- [ ] Add new resources

- [ ] No


lucacavallaro (commented):
The API is ok but I would maintain also the single "environment" input variabile to avoid breaking compatibility from previous versions.

mamu0:
> The API is ok but I would maintain also the single "environment" input variabile to avoid breaking compatibility from previous versions.

As I mentioned earlier, since the change is quite significant, altering both the input and output types, maintaining the current structure would increase the module’s complexity.
Considering it hasn’t been widely used and we are moving forward with the registry, I suggest not maintaining backward compatibility. Instead, we could either refer to the old version where needed or update it to support the new one. 
What do you think, @pagopa/engineering-team-cloud-eng?

gunzip:
I agree with @mamu0

infra/modules/azure_naming_convention/variables.tf
```diff
@@ -1,42 +1,38 @@
-variable "environment" {
-  type = object({
+variable "environments" {
```
Krusty93, line 1: Does this change break all the other modules?

infra/modules/azure_naming_convention/variables.tf
```diff
@@ -1,42 +1,38 @@
...
 
   validation {
-    condition     = contains(["italynorth", "westeurope", "germanywestcentral", "spaincentral", "northeurope"], var.environment.location)
-    error_message = "Allowed values for \"location\" are \"italynorth\" \"westeurope\", \"germanywestcentral\", \"spaincentral\", \"northeurope\"."
+    condition     = alltrue([for env in var.environments : contains(["d", "u", "p"], env.env_short)])
+    error_message = "Allowed values for env_short are 'd', 'u', 'p'."
   }
 
   validation {
-    condition     = var.environment.domain == null ? true : length(replace(var.environment.domain, "-", "")) >= 2
```
Krusty93, line 27: Why validations on `domain` and `app_name` have been removed?

mamu0: To be honest, it seemed unnecessary to me, as the domain is optional and the app_name is not. 
However, I am not entirely sure why there was a two-character limit in place

Krusty93: Just to be sure a team choose a meaningful domain name (at least 2 chars) and a non-empty app name

infra/modules/azure_naming_convention/variables.tf
Krusty93: To me it's ok

infra/modules/azure_naming_convention/locals.tf
```diff
@@ -7,10 +7,81 @@ locals {
...
+      env.app_name
+      ]) => {
+      prefix          = env.prefix
+      env_short       = env.env_short
+      location        = env.location
+      domain          = env.domain
+      app_name        = env.app_name
+      instance_number = env.instance_number
+      project         = "${env.prefix}-${env.env_short}-${lookup(local.location_map, env.location, "neu")}"
+      app_suffix      = [for i in range(env.instance_number) : format("%02d", i + 1)]
```
Krusty93, line 23: What does this line?

infra/modules/azure_naming_convention/locals.tf
```diff
@@ -7,10 +7,81 @@ locals {
...
+    "blob_storage"           = "blob"
+    "queue_storage"          = "queue"
+    "table_storage"          = "table"
+    "file_storage"           = "file"
+    function_storage_account = "stfn"
+
+    # Networking
+    "dns_zone"                    = "dns"
+    "api_management"              = "apim"
+    "api_management_autoscale"    = "apim"
```
Krusty93, line 44:
```suggestion
    "api_management_autoscale"    = "as"
```

infra/modules/azure_naming_convention/locals.tf
```diff
@@ -7,10 +7,81 @@ locals {
...
+  resource_abbreviations = {
+    # Compute
+    "virtual_machine" = "vm"
+
+    # Storage
+    "storage_account"        = "st"
+    "blob_storage"           = "blob"
+    "queue_storage"          = "queue"
+    "table_storage"          = "table"
+    "file_storage"           = "file"
```
Krusty93, line 38: Are these used somewhere?

infra/modules/azure_naming_convention/locals.tf
```diff
@@ -7,10 +7,81 @@ locals {
...
+    function_storage_account = "stfn"
+
+    # Networking
+    "dns_zone"                    = "dns"
```
Krusty93, line 42: Don't think this is used as the DNS zone has only the domain in the name?

infra/modules/azure_naming_convention/locals.tf
mamu0: app_suffix? Now `instance_number` is a number between 1 and 99, with this `for` it generates a list with all suffixes in the classical form 01, 02, ...
Example, if instance number is 3, app_suffix is a list ["01", "02", "03"]
In this way we don't have the only 01 case as before

infra/modules/azure_naming_convention/locals.tf
mamu0: in `dx` no, in the entire org are there cases where it can be useful? Otherwise we can remove them

infra/modules/azure_naming_convention/locals.tf
mamu0: yes the dns name are different, I don't think it's necessary here, I'll remove it 👍

infra/modules/azure_api_management/locals.tf
```diff
@@ -1,7 +1,7 @@
 locals {
   apim = {
-    name           = "${module.naming_convention.prefix}-apim-${module.naming_convention.suffix}"
-    autoscale_name = var.tier == "l" ? "${module.naming_convention.prefix}-apim-as-${module.naming_convention.suffix}" : null
+    name           = module.naming_convention.name.api_management["1"]
```
gunzip, line 3: what does the '1' index means in this context?

infra/modules/azure_api_management/variables.tf
```diff
@@ -14,12 +14,12 @@ variable "environment" {
...
     domain          = optional(string)
     app_name        = string
-    instance_number = string
+    instance_number = number
   })
 
   validation {
-    condition     = length("${var.environment.prefix}${var.environment.env_short}reg${var.environment.domain == null ? "" : replace(var.environment.domain, "-", "")}${var.environment.app_name}-apim-${var.environment.instance_number}") <= 50
-    error_message = "Azure API Management name must contain between 1 and 50 characters. Current value is \"${var.environment.prefix}${var.environment.env_short}reg${var.environment.domain == null ? "" : var.environment.domain}${var.environment.app_name}-apim-${var.environment.instance_number}\""
+    condition     = length("${var.environment.prefix}${var.environment.env_short}reg${var.environment.domain == null ? "" : replace(var.environment.domain, "-", "")}${var.environment.app_name}-apim-00") <= 50
```
gunzip, line 21: where's instance number gone?

infra/modules/azure_naming_convention/README.md
```diff
@@ -1,5 +1,199 @@
...
+
+For configurations with multiple environments, the following outputs are generated:
+
+- `prefixes`:  A list of all prefixes, where each combines prefix, env_short, location_short, optionally including domain and app_name. The format is:
+`[prefix-env_short-location_short-domain-app_name,..]` (e.g. `dx-d-itn-test-eg` if domain is test and app_name is eg).
+- `suffixes`: A map where each key is a prefix, and the value is a map of `instance_number` values as two-digit strings. Example:
+
+  ```hcl
+  {
+    "dx-d-itn-test-eg" = {
```
gunzip, line 74: why not just use the env_short name? I'll try to use one structure for both multi env and mono env. ie

```
{

  "d" = {
      "app_service" = {
        "1" = "dx-d-itn-test-eg-app-01",
        "2" = "dx-d-itn-test-eg-app-02",
        ...
      },
      "blob_storage" = {
        "1" = "dx-d-itn-test-eg-blob-01",
        "2" = "dx-d-itn-test-eg-blob-02",
        ...
      },
      "cosmos_db_nosql" = {
        "1" = "dx-d-itn-test-eg-conmo-01",
        "2" = "dx-d-itn-test-eg-conmo-02",
        ...
      },
      "virtual_network" = {
        "1" = "dx-d-itn-test-eg-vnet-01"
        "2" = "dx-d-itn-test-eg-vnet-02",
        ...
      },
      ...
    },

   "p" = { ... }

]


```

infra/modules/azure_naming_convention/README.md
```diff
@@ -1,5 +1,199 @@
 # DX - Common Variables Validation
 
+This module generates output values that follow the naming conventions used in DX. It takes the following inputs:
+
+- `prefix`: The prefix for the naming convention (e.g., io, dx, ...).
+- `env_short`: A short identifier for the environment (e.g., “d”, "p", "u").
+- `location`: The Azure region or location (e.g. italynorth, westeurope, ...).
+- `domain`: (Optional) The domain name to include in the naming convention.
+- `app_name`: The application name or identifier.
+- `instance_number`: The instance number that can be between 1 and 99.
```
gunzip, line 10: it's not clear from this description what 'instance number' means

infra/modules/azure_naming_convention/README.md
```diff
@@ -1,5 +1,199 @@
 # DX - Common Variables Validation
 
+This module generates output values that follow the naming conventions used in DX. It takes the following inputs:
+
+- `prefix`: The prefix for the naming convention (e.g., io, dx, ...).
+- `env_short`: A short identifier for the environment (e.g., “d”, "p", "u").
+- `location`: The Azure region or location (e.g. italynorth, westeurope, ...).
+- `domain`: (Optional) The domain name to include in the naming convention.
```
gunzip, line 8: I'll try to clarify what 'domain' means, same for 'app_name'

infra/modules/azure_naming_convention/README.md
```diff
@@ -1,5 +1,199 @@
...
+      app_name        = "app"
+      instance_number = 2
+    }
+  ]
+}
+```
+
+To generate the names, you need to use the `output` values generated by this module.
+The output is composed of the following values:
+> module.naming_convention.names.APP_PREFIX.RESOURCE_TYPE.INSTANCE_NUMBER
```
gunzip, line 155: what about

`module.naming_convention.ENV_SHORT.RESOURCE_TYPE.INSTANCE_NUMBER` ?

infra/modules/azure_naming_convention/locals.tf
gunzip: what about adding validation for instance number > 1 <= 99?

gunzip (changes requested):

infra/modules/azure_naming_convention/locals.tf
mamu0: it's already present in `variables.tf` file

```
  validation {
    condition     = alltrue([for env in var.environments : env.instance_number > 0 && env.instance_number < 100])
    error_message = "The variable \"instance_number\" only accepts values in the range [1-99]."
  }
```

infra/modules/azure_naming_convention/README.md
mamu0: _ENV_SHORT_ is not unique, if there is multiple environments with same `env_short` will not work correctly

infra/modules/azure_naming_convention/README.md
mamu0: Here, I’m not entirely sure what the domain actually represents; it’s been used in the organization for a long time, but its meaning has never been very clear.
I think the “app name” is self-explanatory, it’s simply the resource’s identifier. However, if you have suggestions for improving it, I’m open to ideas.

In any case, these are values already present across all repositories and modules, so changing their meaning could result in a significant update, which isn’t the main focus of this PR

infra/modules/azure_naming_convention/README.md
mamu0: ENV_SHORT is not unique, if there is multiple environments with same env_short will not work correctly

infra/modules/azure_api_management/variables.tf
mamu0: it's just the validate for the number of characters, since now instance number is a `number` it may not be 2 digits, but the string value that the suffix assumes is always 2 digits, so to have the same result just put 2 digits, I used `00` symbolically, the possible values are all those from 01 to 99

infra/modules/azure_api_management/locals.tf
mamu0: it is the first instance, so the one that will have `01` as a suffix, in the ReadMe I have inserted the explanation of what form the various values ​​will have, in short we have a name for each instance

.changeset/neat-tigers-fix.md
```diff
@@ -0,0 +1,5 @@
+---
```
gunzip, line 1: very good!!

infra/modules/azure_api_management/locals.tf
gunzip: why do we hardcode `1` for this resource and dont' use `instance_number` ? do we expect to have just one instance for environment?

infra/modules/azure_api_management/network.tf
```diff
@@ -32,7 +32,7 @@ resource "azurerm_private_dns_a_record" "apim_scm_azure_api_net" {
 
 # Define security group
 resource "azurerm_network_security_group" "nsg_apim" {
-  name                = "${module.naming_convention.prefix}-apim-nsg-${module.naming_convention.suffix}"
+  name                = module.naming_convention.name.apim_network_security_group["1"]
```
gunzip, line 35: ditto

infra/modules/azure_app_service/app_service_plan.tf
```diff
@@ -1,7 +1,7 @@
 resource "azurerm_service_plan" "this" {
   count = local.app_service_plan.enable ? 1 : 0
 
-  name                   = "${module.naming_convention.prefix}-asp-${module.naming_convention.suffix}"
+  name                   = module.naming_convention.name.app_service_plan["1"]
```
gunzip, line 4: ditto (why hardcode `1` ?)

infra/modules/azure_app_service/locals.tf
```diff
@@ -12,7 +12,7 @@ locals {
   }
 
   app_service = {
-    name                   = "${module.naming_convention.prefix}-app-${module.naming_convention.suffix}"
+    name                   = module.naming_convention.name.app_service["1"]
```
gunzip, line 15: ditto

infra/modules/azure_app_service_exposed/app_service_plan.tf
```diff
@@ -1,7 +1,7 @@
 resource "azurerm_service_plan" "this" {
   count = local.app_service_plan.enable ? 1 : 0
 
-  name                   = "${module.naming_convention.prefix}-asp-${module.naming_convention.suffix}"
+  name                   = module.naming_convention.name.app_service_plan["1"]
```
gunzip, line 4: ditto

infra/modules/azure_naming_convention/README.md
gunzip: I'd add a description for app_name that make constraints explicit, ie in case there's a limit on the length. Moreover I'd add in the description that is the "resource identifier".

infra/modules/azure_naming_convention/README.md
gunzip: In which case there are multiple environments with the same `env_short` names?

gunzip: Anyway, we assume that APP_PREFIX `dx-d-itn-test-app` is unique, but it can be computed by the naming module itself

infra/modules/azure_api_management/locals.tf
mamu0: yes the modules always create only one instance, but you are right about using `istance_number`, it is not necessarily always 1, I will fix it thanks

infra/modules/azure_naming_convention/README.md
mamu0: Never inside the modules, but as @lucacavallaro  and @BurnedMarshal had pointed out to us, it happened that they used different environments and therefore were forced to instantiate the module multiple times. This is why we opted for this multi-environment solution in the end.

The problem here would then be that in a single instance you can call all the environments you want, and since, for example, you can make multiple environments with only `app_name` that varies, or only the `env`, the key had to be made as a composite value.

If we find a way to simplify the key obviously it would be better, for now I have inserted the most effective form.

gunzip: do these "different environments" differentiate by `env_short`? or what else? in the first case `env_short` would be a unique key.

PR #183, opened 2024-12-11
[CES-599] Remove dependency from azurerm-terraform-v3 repository for GitHub runner module

Krusty93:
### List of changes

Remove dependency from azurerm-terraform-v3 repository.
Added possibility to set a resource group and a region
Added support for azurerm v4
Switched from `az_api` to `azurerm_container_app_job`

### Motivation and context

Due to module dependency, we were unable to set neither custom resource groups or region. Moreover, we were limited to use azurerm v3.

This PR contains breaking changes but recreating the runner is not a problem

### Type of changes

- [ ] Add new resources

- [X] No


infra/modules/github_selfhosted_runner_on_container_app_jobs/main.tf
```diff
@@ -1,35 +1,22 @@
 terraform {
+
   required_providers {
     azurerm = {
       source  = "hashicorp/azurerm"
-      version = "~>3.86"
+      version = ">= 3.110, < 5.0"
```
gunzip, line 6: does this works even for v3 codebases or needs a provider upgrade? in other workds: is this a breaking change?

infra/modules/github_selfhosted_runner_on_container_app_jobs/container_app_job.tf
```diff
@@ -0,0 +1,82 @@
...
+
+    scale {
+      max_executions              = var.container_app_environment.max_instances
+      min_executions              = var.container_app_environment.min_instances
+      polling_interval_in_seconds = var.container_app_environment.polling_interval_in_seconds
+
+      rules {
+        name             = "github-runner-rule"
+        custom_rule_type = "github-runner"
+        metadata = merge({
```
gunzip, line 26: can you add a link to this metadata documentation?

infra/modules/github_selfhosted_runner_on_container_app_jobs/main.tf
Krusty93: No, as clients can (and do) manage version on their side setting the preferred one

gunzip (approved):

PR #187, opened 2024-12-11
[CES-605] Upgrade Ubuntu version used by managed GitHub runner

Krusty93:
### List of changes

Replace old versions of Ubuntu with latest

### Type of changes

- [ ] Add new resources

- [X] No

### DIFF
--- a/.github/workflows/docker_image_build.yaml
+++ b/.github/workflows/docker_image_build.yaml
@@ -39,7 +39,7 @@ jobs:
 
   release:
     name: Build and Push to Registry
-    runs-on: ubuntu-20.04
+    runs-on: ubuntu-latest
     permissions:
       packages: write
 
--- a/.github/workflows/function_app_deploy.yaml
+++ b/.github/workflows/function_app_deploy.yaml
@@ -43,7 +43,7 @@ jobs:
 
   build:
     name: Build Artifact
-    runs-on: ubuntu-20.04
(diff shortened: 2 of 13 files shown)


.github/workflows/docker_image_build.yaml
```diff
@@ -39,7 +39,7 @@ jobs:
 
   release:
     name: Build and Push to Registry
-    runs-on: ubuntu-20.04
+    runs-on: ubuntu-latest
```
gunzip, line 42: this change has the potential to break the workflow in future since it's a dangling label. what about fixing it to 24.x instead?

Krusty93: IMHO:

In any case we should watch out for changes in GitHub runners, but:
- pinning the version we have to do a PR in any case
- not pinning the version we shall do a PR only if something is breaking (which I see ok), and restoring the x-1 version

In short, I suggest to spend time on these details only if we need to. On the other hand, we should keep an eye on this coming changes from providers

lucacavallaro: I like `ubuntu-latest` because it forces us to fix/upgrade our workflows as soon as possible!

gunzip (approved):

gunzip (approved):

mamu0 (approved):

PR #188, opened 2024-12-12
[CES-607] Add support to custom startup command in AppService modules

Krusty93:
### List of changes

Override app startup command if PM2 entry point file is provided in AppService modules

### Motivation and context

More teams are using PM2 to exploit full hardware capacity

### Type of changes

- [ ] Add new resources

- [X] No

### Other information

depends on #204

### DIFF
--- a/infra/modules/azure_app_service_exposed/locals.tf
+++ b/infra/modules/azure_app_service_exposed/locals.tf
@@ -9,6 +9,7 @@ locals {
     zone_balancing_enabled = local.tier != "s" && local.tier != "xs"
     is_slot_enabled        = local.tier == "s" || local.tier == "xs" ? 0 : 1
     always_on              = local.tier == "xs" ? false : true
+    command_line           = var.pm2_startup_file_name == null ? null : "pm2 start ${var.pm2_startup_file_name} -i max --no-daemon"
   }
 
   application_insights = {
--- a/infra/modules/azure_app_service_exposed/locals.tf
+++ b/infra/modules/azure_app_service_exposed/locals.tf
@@ -9,7 +9,10 @@ locals {
     zone_balancing_enabled = local.tier != "s" && local.tier != "xs"
     is_slot_enabled        = local.tier == "s" || local.tier == "xs" ? 0 : 1
     always_on              = local.tier == "xs" ? false : true
-    command_line           = var.pm2_startup_file_name == null ? null : "pm2 start ${var.pm2_startup_file_name} -i max --no-daemon"
+    startup_command = (var.stack == "node"
+      ? (var.startup_command == "" ? "pm2 start index.js -i max --no-daemon" : var.startup_command)
+      : (var.stack == "java" && var.startup_command == "" ? null : var.startup_command)
+    )
   }
 
   application_insights = {
--- a/.changeset/popular-bears-draw.md
+++ b/.changeset/popular-bears-draw.md
@@ -1,0 +1,9 @@
+---
+"azure_app_service_exposed": major
+"azure_app_service": major
+---
+
+Two breaking changes introduced in this version:
+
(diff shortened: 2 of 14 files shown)


BurnedMarshal (approved):
LGTM, maybe we can add a validation limiting the property `pm2_startup_file_name` with `node` stack

infra/modules/azure_app_service_exposed/locals.tf
```diff
@@ -9,6 +9,7 @@ locals {
     zone_balancing_enabled = local.tier != "s" && local.tier != "xs"
     is_slot_enabled        = local.tier == "s" || local.tier == "xs" ? 0 : 1
     always_on              = local.tier == "xs" ? false : true
+    command_line           = var.pm2_startup_file_name == null ? null : "pm2 start ${var.pm2_startup_file_name} -i max --no-daemon"
```
rodoviti, line 12: **question**: do you think this way will also work with NextJS apps?

We have some NextJS apps running on App Service that are currently started with `yarn start`, which is defined as follow: `"start": "./node_modules/next/dist/bin/next start"`

PS: I found [this discussion](https://github.com/vercel/next.js/discussions/31461) that may be useful

gunzip:
Why not allow users to define *any* startup command instead of binding it specifically to PM2?

Krusty93:
> Why not allow users to define _any_ startup command instead of binding it specifically to PM2?

To keep things single. I don't know if we ever need to add custom commands. What are your thoughts?

gunzip:
Why do you want to constraint users to a peculiar usage of pm2 ? One user may have a pm2.json configuration in its own codebase for example

Krusty93:
> Why do you want to constraint users to a peculiar usage of pm2 ? One user may have a pm2.json configuration in its own codebase for example

What's your proposal then? Wrapping the TF property? Because I liked the idea where devs don't need to know anything about it, but just saying "here's the file". However, I could be wrong and going against devs preferences

gunzip:
I vot to just forward the TF `command_line` property without constraint it to `pm2`. Consider that this module works for java as well.

BurnedMarshal:
> I vote to just forward the TF `command_line` property without constraint it to `pm2`. Consider that this module works for java as well.

I understand your point, but the use of PM2 is not a suggestion it's recommended by Microsoft. You what to shift the responsibility to the team to use it or the capability to configure it, but they don't use it. A good module handle the resources on the good way giving awareness to the team.

gunzip:
@BurnedMarshal that's right, but in this case, if we want to make `pm2` mandatory on every `NodeJS` App Service, then we should make `pm2_startup_file_name` mandatory, with a default value, when NodeJS is specified as runtime.

BurnedMarshal:
> > @BurnedMarshal that's right, but in this case, if we want to make `pm2` mandatory on every `NodeJS` App Service, then we should make `pm2_startup_file_name` mandatory, with a default value, when NodeJS is specified as runtime.
> 
> @BurnedMarshal Are we confident to enable PM2 everywhere? Consider that developers will find a drift in the terraform configuration and they will be more interested in fixing the breaking change rather than doing tests and fine tuning on the PM2 configuration

Every team should keep ownership on this, evaluating changing accordingly with their roadmap an requisites. But DevEX guidelines should be aligned with the CP documentation and recommendation.

christian-calabrese (commented):
Great work with the tests! Could you please add a changeset? I think this may concur for a major

christian-calabrese:
Due to the possibility of breaking changes, it will be merged after the adoption of the registry

gunzip:
depends on #204

gunzip:
Blocked by CES-965

Krusty93:
> @Krusty93 is it still relevant?

Yes, the plan is to release the breaking change in the next quarter within the others

Krusty93:
Obsolete

PR #189, opened 2024-12-13
[CES-598] Added GitHub Runner in azure_core_infra module

mamu0:
### List of changes

- Added `github_runner` module into `azure_core_infra`
- Added `log_analytics` module into `azure_core_infra`
- Added Key Vault Access roles to CI/CD identities
- Updated DEV resources with new configuration

### Motivation and context

To ensure the proper functionality of the Self-Hosted Runner, we have integrated the GitHub Runner configuration into the Azure Core Infrastructure module. This setup ensures that the runner and its associated job are available for the main repository responsible for creating the core infrastructure for the subscription.

### Type of changes

- [ ] Add new resources

- [x] No


infra/modules/azure_core_infra/_modules/github_runner/container_app_environment.tf
```diff
@@ -0,0 +1,20 @@
+resource "azurerm_container_app_environment" "cae" {
+  name                = "${var.project}-github-runner-cae"
```
Krusty93, line 2: Can we update the name?

infra/modules/azure_core_infra/_modules/github_runner/container_app_jobs.tf
```diff
@@ -0,0 +1,96 @@
...
+
+  # Prevent false plan changes on secret part
+  lifecycle {
+    ignore_changes = [secret]
+  }
+
+  tags = var.tags
+}
+
+resource "azurerm_key_vault_access_policy" "keyvault_containerapp" {
```
Krusty93, line 88: I don't think it's right as the kv uses RBAC

Krusty93:
I think you can get inspiration from #183

infra/modules/azure_core_infra/_modules/github_runner/container_app_jobs.tf
mamu0: you're right! I fix it

infra/modules/azure_core_infra/_modules/github_runner/container_app_jobs.tf
```diff
@@ -0,0 +1,98 @@
+resource "azurerm_container_app_job" "container_app_job" {
```
gunzip, line 1: why don't we use https://github.com/pagopa/dx/tree/main/infra/modules/github_selfhosted_runner_on_container_app_jobs ?

mamu0: On second thought, the job shouldn’t be configured here at all. Since this is a module in the DX repository that can be invoked across different repositories, it is not a core element to be created initially, it should be created under resources instead!

gunzip: right!

infra/modules/azure_core_infra/main.tf
```diff
@@ -136,5 +155,45 @@ module "dns" {
     name = module.network.vnet.name
   }
 
+  tags = var.tags
+}
+
+#------#
+# LOGS #
```
gunzip, line 162: why do we need this resource? (to log wehat?) I'd add a comment.

mamu0: is mandatory for github runner environment resource

infra/modules/azure_core_infra/variables.tf
```diff
@@ -59,3 +59,10 @@ variable "nat_enabled" {
   description = "Flag to enable nat gateway creation"
   default     = false
 }
+
+
+variable "gh_runner_snet" {
+  type        = string
+  default     = "10.0.242.0/23"
```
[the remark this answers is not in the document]
mamu0: it is the same value that we have used in other contexts, and in any case if the module is instantiated without configuring any subnet it is a value that does not affect any other IP range and leaves all the initial ones free

infra/identity/dev/main.tf
```diff
@@ -58,3 +58,33 @@ module "federated_identities" {
...
+
+module "roles_ci" {
+  source       = "../../modules/azure_role_assignments"
+  principal_id = module.federated_identities.federated_ci_identity.id
+
+  key_vault = [
+    {
+      name                = "dx-d-itn-common-kv-01"
+      resource_group_name = "dx-d-itn-common-rg-01"
+      roles = {
```
christian-calabrese, line 70:
```suggestion
      name                = "${local.project}-itn-common-kv-01"
      resource_group_name = "${local.project}-itn-common-rg-01"
      roles = {
```

infra/identity/dev/main.tf
```diff
@@ -58,3 +58,33 @@ module "federated_identities" {
...
+}
+
+module "roles_cd" {
+  source       = "../../modules/azure_role_assignments"
+  principal_id = module.federated_identities.federated_cd_identity.id
+
+  key_vault = [
+    {
+      name                = "dx-d-itn-common-kv-01"
+      resource_group_name = "dx-d-itn-common-rg-01"
```
christian-calabrese, line 84:
```suggestion
      name                = "${local.project}-itn-common-kv-01"
      resource_group_name = "${local.project}-itn-common-rg-01"
```

infra/modules/azure_core_infra/main.tf
gunzip: I'd add this as a comment

infra/modules/azure_core_infra/variables.tf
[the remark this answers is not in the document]
gunzip: I'd clarify with a comment the origin of this value. If it's a random value we should say it is random. If we have constraints we should clarify which ones.

infra/modules/azure_core_infra/main.tf
mamu0: I don’t think a comment is necessary. Yes, I added this resource because it’s essential for the runner, but it wasn’t added specifically for it, it’s a common resource. Adding a comment could create confusion, as it might be interpreted as “this is only for the GitHub runner,” which isn’t the case

infra/modules/azure_core_infra/variables.tf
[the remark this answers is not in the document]
Krusty93: It was the first IP range available in IO WEU VNet at the time.

As we are providing this module for isolated and ephemeral test environments, it might be good to have default values for IP blocks, to accelerate things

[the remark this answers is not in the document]
gunzip: What do you propose as default?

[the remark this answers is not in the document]
Krusty93: The ones added in this PR are ok to me

infra/modules/azure_core_infra/_modules/log_analytics/variables.tf
Krusty93, line 1: Could you please add descriptions here?

Krusty93 (approved):

PR #190, opened 2024-12-13
[#CES-613] Autoscaler support shared App Service  Plan

BurnedMarshal:
### List of changes
Can be provided a list of services to an autoscaler definition.
Requests scale settings are generated for each resource using the same App Service Plan

### Motivation and context
Team require to use a single App Service Plan to more services. The autoscale have metrics linked to App Service Plan like CPU, but other metrics are available on single services attached to it, like requests.
We need to creare request scale rules based on each service.

### Type of changes

- [ ] Add new resources

- [ ] No


infra/modules/azure_app_service_plan_autoscaler/variables.tf
```diff
@@ -14,16 +14,21 @@ variable "autoscale_name" {
   default     = null
 }
 
-variable "target_service" {
-  type = object({
+variable "target_services" {
```
Krusty93, line 17: I suggest to change the variable to this structure, even though a breaking change is introduced:

```hcl
type = object({
    app_service_name  = optional(list(string))
    function_app_name = optional(list(string))
})
```

Looks smoother to me :) 

cc @gunzip

gunzip: Agree

BurnedMarshal: Fixed in [f8de935](https://github.com/pagopa/dx/pull/190/commits/f8de9355a42d7688b83992c3f5de0ea22ea3bc6b)

gunzip:
is this still relevant @BurnedMarshal  ?

BurnedMarshal:
> is this still relevant @BurnedMarshal ?

Yes, this PR complete the support for Autoscaler Settings plan on shared App Service Plan. Without it teams need to redefine the autoscaler outside the shared plan like [here](https://github.com/pagopa/io-auth-n-identity-domain/blob/main/infra/resources/prod/shared_plan.tf#L45)
The PR require a fix on static analysis. cc. @Krusty93

mamu0:
Closing in favor of #543 which replaces this PR and includes the latest changes. Thanks for your work here!

PR #195, opened 2024-12-18
[CES-624] Add support for Durable Function in Function App TF module

Krusty93:
### List of changes

Add private endpoint to table storage
Add RBAC to manage table storage

### Motivation and context

Add support to run Durable Functions in Function Apps

### Type of changes

- [ ] Add new resources

- [X] No


infra/modules/azure_function_app/variables.tf
```diff
@@ -168,4 +168,10 @@ variable "application_insights_key" {
   description = "(Optional) Application Insights key"
   sensitive   = true
   default     = null
-}
\ No newline at end of file
+}
+
+variable "contains_durable_functions" {
```
gunzip, line 173:
```suggestion
variable "has_durable_functions" {
```

Krusty93: Done

gunzip:
Quoting the [MS docs](https://learn.microsoft.com/en-us/azure/azure-functions/durable/durable-functions-azure-storage-provider#durable-functions-1x): "For performance-sensitive workloads, however, configuring a non-default storage account is recommended. Durable Functions uses Azure Storage heavily, and using a dedicated storage account isolates Durable Functions storage usage from the internal usage by the Azure Functions host."

For tires > development we shoud use another storage account.

infra/modules/azure_function_app/storage_account.tf
```diff
@@ -14,6 +14,24 @@ resource "azurerm_storage_account" "this" {
   tags = var.tags
 }
 
+resource "azurerm_storage_account" "durable_function" {
```
gunzip, line 17: **question**: what's the reason behind the choice to not use the dx storage account module?

gunzip (commented):
**thought**: 

1. can we isolate the storage account part into another module?
1. I don't know if the function default storage account settings differ from the durable storage account one; if they don't, we may reuse the shared module at [1] instead of configuring them twice
1. We should backport this into `azure_function_app_exposed` as well; having a reusable module ie. `azure_function_app_storage` may probably help to not duplicate code
1. we can do this with another PR, but small tier should reuse the default storage account (maybe [1] helps even for this scenario)

what about these?

infra/modules/azure_function_app/variables.tf
```diff
@@ -15,7 +15,12 @@ variable "environment" {
 
   validation {
     condition     = length("${var.environment.prefix}${var.environment.env_short}reg${var.environment.domain == null ? "" : replace(var.environment.domain, "-", "")}${var.environment.app_name}stfn${var.environment.instance_number}") <= 24
-    error_message = "Storage Account name must have less than 25 characters. Current value is \"${var.environment.prefix}${var.environment.env_short}reg${var.environment.domain == null ? "" : var.environment.domain}${var.environment.app_name}st${var.environment.instance_number}\""
+    error_message = "Storage Account name must have less than 25 characters. Current value is \"${var.environment.prefix}${var.environment.env_short}reg${var.environment.domain == null ? "" : var.environment.domain}${var.environment.app_name}stfn${var.environment.instance_number}\""
```
gunzip, line 18: why the 'reg' infix ? it's difficult to obtain less than 25 chars with this naming ie.

https://github.com/pagopa/dx-playground/actions/runs/12938884631/job/36090011729#step:9:341

"dxdregplaygrounddfstfnd01"

Krusty93: ´reg´stands for region and is a placeholder (I can't get the real value here).

I propose to use `stfd` instead of `stfnd`as resource suffix, to match the length capacity of the host's storage account (which is `stfn`)

gunzip: this would be a breaking change for existing storages, isn't it?

infra/modules/azure_function_app/storage_account.tf
Krusty93: Agreed to refactor in another PR using a "local" storage account module, not exposed to customers

infra/modules/azure_function_app/variables.tf
Krusty93: It is not as we're introducing the durable storage account in this pr

gunzip (approved):
We should add in the module docs an example host.json

```
"extensions": {
    "durableTask": {
      "storageProvider": {
        "connectionStringName": "DfStorageConnectionName"
      },
      "hubName": "%SLOT_TASKHUB_NAME%",
      "useGracefulShutdown": true
    }
  }
```

Krusty93:
> We should add in the module docs an example host.json
> 
> ```
> "extensions": {
>     "durableTask": {
>       "storageProvider": {
>         "connectionStringName": "DfStorageConnectionName"
>       },
>       "hubName": "%SLOT_TASKHUB_NAME%",
>       "useGracefulShutdown": true
>     }
>   }
> ```

I'm going to make another PR with some docs for all modules (maybe AI generated)

mamu0 (approved):

PR #196, opened 2024-12-19
Make registry modules lock scripts resilient

christian-calabrese:
### List of changes

- Registry modules locking script made more resilient
- Allow the script to be correctly invoked via pre-commits configs
- Made infra/scripts versioned via changeset
- Switch all modules invocations from git to registry source

### Motivation and context

The lock modules script will be executed at every push and every plan or apply executed via github actions.
For this reason, it must be as resilient as possible to not cause issues in the ci/cd process.

To configure this pre-commit in your repo, it's possible to follow the following extract of the `.pre-commit-config.yaml` file:

```
repos:
  - repo: https://github.com/pagopa/dx
    rev: 0.0.1
    hooks:
      - id: lock_modules
        exclude: ^.*/(_modules|modules|\.terraform)(/.*)?$
        files: src/(common|core|github_runner|identity|migration|repository)
```

### Type of changes

- [ ] Add new resources

- [ ] No


infra/github-runner/dev/tfmodules.lock.json
```diff
@@ -0,0 +1,3 @@
+{
+  "container_app_job_selfhosted_runner.naming_convention": "807e8fafaf3cda8d1df7cc5c624715555ff150e87a8df0becc7e5cab3e54f855"
```
Krusty93, line 2: What is this?

christian-calabrese: This is an actual example of the files being generated by the lock script. You can look at it as the .terraform.lock.hcl, but for modules.
Something similar will be created in every terraform configuration folder where at least a registry module is found

christian-calabrese: This is how it is actually computed:
https://github.com/pagopa/dx/pull/196/files#diff-fd9d06553ede4d2f63d44ae6a3b64f2946925b649419339513f4d6f09a9dcb91R80

mamu0 (approved):

Krusty93 (approved):

PR #197, opened 2024-12-19
Fix java_server_version syntax in azure_app_service modules

christian-calabrese:
### List of changes

Add app_command_line to java based app services

### Motivation and context

Java based app services needs the application startup command to be set by the user to launch the jar file.
### Type of changes

- [ ] Add new resources

- [X] No

### DIFF
--- a/infra/modules/azure_app_service_exposed/locals.tf
+++ b/infra/modules/azure_app_service_exposed/locals.tf
@@ -9,10 +9,6 @@ locals {
     zone_balancing_enabled = local.tier != "s" && local.tier != "xs"
     is_slot_enabled        = local.tier == "s" || local.tier == "xs" ? 0 : 1
     always_on              = local.tier == "xs" ? false : true
-    command_line = {
-      node = null
-      java = "java -jar /home/site/wwwroot/app.jar --server.port=80"
-    }
   }
 
   application_insights = {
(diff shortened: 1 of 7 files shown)


mamu0 (approved):

infra/modules/azure_app_service_exposed/locals.tf
```diff
@@ -9,6 +9,10 @@ locals {
     zone_balancing_enabled = local.tier != "s" && local.tier != "xs"
     is_slot_enabled        = local.tier == "s" || local.tier == "xs" ? 0 : 1
     always_on              = local.tier == "xs" ? false : true
+    command_line = {
+      node = null
```
gunzip, line 13: question: is there a way to owerwrite this for the module user?

Krusty93: Not in this PR, but I think @christian-calabrese could bring the changes of the #188 (still to resolve)

christian-calabrese: We can surely do it, I'll integrate it in this PR

Krusty93 (approved):

infra/modules/azure_app_service_exposed/locals.tf
Krusty93: Obsolete after latest changes. I solve this conversation

PR #199, opened 2024-12-23
[CES-614] Added custom roles into Core Infra module

mamu0:
### List of changes

Added custom roles definition into module for core infrastructure.

### Motivation and context

To permit least privilege for RBAC we can define custom roles in core infrastructure, the added role is to list secrets, useful to avoid Role `API Management Service Contributor` for a CI identity, with this we can use `API Management Service Reader` in combo with new one `PagoPA API Management Service List Secrets`

### Type of changes

- [ ] Add new resources

- [x] No

### DIFF
--- a/infra/modules/azure_core_infra/_modules/custom_roles/roles.tf
+++ b/infra/modules/azure_core_infra/_modules/custom_roles/roles.tf
@@ -1,0 +1,16 @@
+resource "azurerm_role_definition" "pagopa_apim_list_secrets" {
+  name        = "PagoPA API Management Service List Secrets"
+  scope       = local.subscription_id
+  description = "List only access to API Management secrets"
+
+  permissions {
+    actions = [
+      "Microsoft.ApiManagement/service/*/listSecrets/action",
+    ]
+    not_actions = []
+  }
+
+  assignable_scopes = [
+    local.subscription_id
+  ]
+}
--- a/.changeset/poor-wasps-wink.md
+++ b/.changeset/poor-wasps-wink.md
@@ -1,0 +1,5 @@
+---
+"azure_core_infra": patch
+---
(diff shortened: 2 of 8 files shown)


infra/modules/azure_core_infra/_modules/custom_roles/roles.tf
```diff
@@ -0,0 +1,16 @@
+resource "azurerm_role_definition" "pagopa_apim_list_secrets" {
```
gunzip, line 1: I'd remove every reference to 'pagopa' as these are general purpose modules

infra/modules/azure_core_infra/_modules/custom_roles/roles.tf
```diff
@@ -0,0 +1,16 @@
...
+  name        = "PagoPA API Management Service List Secrets"
```
gunzip, line 2:
```suggestion
resource "azurerm_role_definition" "dx_apim_list_secrets" {
  name        = "API Management Service List Secrets"
```

infra/modules/azure_core_infra/_modules/custom_roles/roles.tf
mamu0: It should be evaluated better, because the name chosen must be the same for all subscriptions

infra/modules/azure_core_infra/_modules/custom_roles/roles.tf
mamu0: ditto

Krusty93:
Is this ready for review?

mamu0:
This PR is now moved into this https://github.com/pagopa/eng-azure-governance/pull/159

PR #200, opened 2025-01-07
Github runner labels can be overridden

christian-calabrese:
### List of changes

Github runner labels can be overridden with a variable

### Motivation and context

In scenarios where a repository’s code needs to be deployed to multiple subscriptions of the same environment (prod-a and prod-b), GitHub Actions must identify the appropriate runner to invoke. 
To achieve this, it is essential to allow specifying the label that identifies the runner `runs-on: prod-a` or `runs-on: prod-b`. 

Additionally, the GitHub runner must be configured to scale out dynamically when triggered with the corresponding label (prod-a or prod-b).

### Type of changes

- [ ] Add new resources

- [ ] No


.github/workflows/function_app_deploy.yaml
```diff
@@ -101,7 +111,10 @@ jobs:
     name: Deploy
     if: ${{ !github.event.act }}
     needs: [build]
-    runs-on: ${{ inputs.use_private_agent == true && 'self-hosted' || 'ubuntu-latest' }}
+    # Use inputs.override_labels if set; otherwise, fall back to inputs.environment. 
+    # When inputs.use_labels and inputs.use_private_agent are true, apply the selected labels. 
+    # Default to 'self-hosted' if inputs.use_private_agent is true, or 'ubuntu-latest' otherwise.
+    runs-on: ${{ inputs.use_labels && inputs.use_private_agent && (inputs.override_labels != '' && inputs.override_labels || inputs.environment) || inputs.use_private_agent && 'self-hosted' || 'ubuntu-latest' }}
```
mamu0, line 117: I’m not entirely sure about this, but if you want override_labels to not be empty, you can simply use:
```suggestion
    runs-on: ${{ inputs.use_labels && inputs.use_private_agent && inputs.override_labels != '' || inputs.use_private_agent && 'self-hosted' || 'ubuntu-latest' }}
```
override_labels already has a default value, so the condition with just inputs.override_labels is redundant. Additionally, inputs.environment is required, so its value will exist anyway

mamu0 (approved):

.github/workflows/function_app_deploy.yaml
```diff
@@ -31,6 +31,16 @@ on:
         type: boolean
         required: false
         default: true
+      use_labels:
+        description: Use labels to start the right environment's GitHub runner. If use_labels is true, also use_private_agent must be set to true
+        type: boolean
+        required: false
+        default: false
+      override_labels:
```
mamu0, line 39: where is this input used?

christian-calabrese: In the runs-on condition at line 117

PR #201, opened 2025-01-09
[PE-881] Output PEP's recordsets for App Services

christian-calabrese:
### List of changes

Output PEP's recordsets for App Services

### Motivation and context

In order to enable private reachability of an app service located in a peered vnet, this information is needed. In fact, the dns records pointing to the pep must also be created in the peered vnet's private dns zone.
### Type of changes

- [ ] Add new resources

- [X] No

### DIFF
--- a/infra/modules/azure_app_service/outputs.tf
+++ b/infra/modules/azure_app_service/outputs.tf
@@ -17,14 +17,14 @@ output "app_service" {
       name         = azurerm_linux_web_app.this.name
       principal_id = azurerm_linux_web_app.this.identity[0].principal_id
       pep = {
-        record_sets = azurerm_private_endpoint.app_service_sites.private_dns_zone_configs.record_sets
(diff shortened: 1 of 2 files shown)


infra/modules/azure_app_service/outputs.tf
```diff
@@ -16,10 +16,16 @@ output "app_service" {
       id           = azurerm_linux_web_app.this.id
       name         = azurerm_linux_web_app.this.name
       principal_id = azurerm_linux_web_app.this.identity[0].principal_id
+      pep = {
+        record_sets = azurerm_private_endpoint.app_service_sites.private_dns_zone_configs.record_sets
+      }
```
mamu0, line 21: if there are nothing else from _pep_ that we need, we can do this
```suggestion
      pep_record_sets = azurerm_private_endpoint.app_service_sites.private_dns_zone_configs.record_sets
```

mamu0 (approved):

PR #224, opened 2025-01-21
[CES-669] Fix an issue where the AI sampling percentage was not set when using Instrumentation Key in Function Apps

Krusty93:
### List of changes

Fix variable validation and add unit tests

### Motivation and context

The AI sampling percentage was not set when the AI Instrumentation Key were used instead of connection string in Function Apps

### Type of changes

- [ ] Add new resources

- [X] No


gunzip:
not needed for this PR, but consider that `application_insights_key` will be deprecated soon: 

"On March 31, 2025, support for instrumentation key ingestion will end. Instrumentation key ingestion will continue to work, but we'll no longer provide updates or support for the feature."

https://learn.microsoft.com/en-us/azure/azure-monitor/app/connection-strings

infra/modules/azure_function_app/variables.tf
```diff
@@ -177,7 +182,7 @@ variable "action_group_id" {
 
 variable "application_insights_key" {
   type        = string
-  description = "(Optional) Application Insights key"
+  description = "(Optional) Application Insights instrumentation key. Do not use if you are using a connection string, which is instead preferrable"
```
gunzip, line 185: I'd write that is deprecated from march 2025

gunzip (approved):

gunzip:
this should be backported in azure_function_app_exposed as well. 

See also https://github.com/pagopa/dx/blob/main/infra/modules/azure_function_app/examples/complete/main.tf#L24

gunzip:
since 

1. the AI key is gonna be deprecated soon
1. AI integration should be optional

I'm gonna revert this PR.

PR #226, opened 2025-01-21
[CES-670] Create workflow to run Terraform tests in CI

Krusty93:
### List of changes

Add workflow to run Terraform tests when a change is applied

### Motivation and context

Do not break things

### Type of changes

- [X] Add new resources

- [X] No


christian-calabrese (approved):

mamu0 (approved):

.github/workflows/infra_test.yaml
```diff
@@ -0,0 +1,107 @@
...
+    needs: [get-modified-modules, get-all-modules]
+    if: always()
+    environment: dev-ci
+    permissions:
+      id-token: write
+      contents: read
+    strategy:
+      fail-fast: false
+      matrix:
+        module: ${{ fromJSON(needs.get-all-modules.outputs.modules || needs.get-modified-modules.outputs.modules) }}
```
gunzip, line 70: **question**: why do we try with all modeules then only modified ones?

Krusty93: We are checking which one has the value, and the order doesn't matter as either one of the two jobs will run

gunzip: **question**: in which case `all-modules` doesn't have a value but modified modules has?

gunzip: **question**: since only one will run, doesn't this block when checking 
`needs: [get-modified-modules, get-all-modules]` which needs both?

Krusty93: No, because of the `if: always()`

infra/modules/azure_function_app/locals.tf
```diff
@@ -45,4 +45,6 @@ locals {
   private_dns_zone = {
     resource_group_name = var.private_dns_zone_resource_group_name == null ? var.virtual_network.resource_group_name : var.private_dns_zone_resource_group_name
   }
+
+  fake = "fake value"
```
gunzip, line 49: what's this?

Krusty93: I trick to trigger the pipeline while it is not into master yet

PR #230, opened 2025-01-22
[CES-679] Make AI integration optional

gunzip:
### List of changes

Change the validation on AI variables to allow module users to create functions without integrating with Application Insights.

### Motivation and context

There was a breaking change that made default AI 
integration mandatory:

https://github.com/pagopa/dx/pull/224

which breaks existing function apps and even the example in this repo:

https://github.com/pagopa/dx/blob/main/infra/modules/azure_function_app/examples/complete/main.tf#L24

### Type of changes

- [ ] Add new resources

- [X] No

### DIFF
--- a/.changeset/curly-rockets-give.md
+++ b/.changeset/curly-rockets-give.md
@@ -1,0 +1,5 @@
+---
+"azure_function_app": patch
+---
+
+Make AI integration optional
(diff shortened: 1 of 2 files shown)


gunzip:
we will revert https://github.com/pagopa/dx/pull/224 instead

PR #232, opened 2025-01-23
[CES-679] Revert FIX for sampling bug when using AI key instead of Connection String

gunzip:
### List of changes

Reverts #224

### Motivation and context

We won't support a deprecated feature, teams must
use connection strings in their AI setup.

### Type of changes

- [ ] Add new resources

- [X] No


gunzip:
@Krusty93 if you agree I'm gonna merge this one to avoid a breaking change. For example, on dx-playground I cannot create functions enaymore since they don't use AI (we're assessing Dynatrace)

Krusty93 (approved):

PR #236, opened 2025-01-24
[CES-690] Dynamically get terraform version from file in static analysis workflow

christian-calabrese:
### List of changes
- Added reusable workflow to retrieve the terraform version from the .terraform-version file
- Include the new workflow in static analysis to optionally retrieve the terraform version
- Added possibility to print verbose logs of pre-commit runs

### Motivation and context
All repositories include the .terraform-version file. This PR aims to avoid the need of updating the terraform version in cicd pipelines by automatically retrieving the version from the file.

### Type of changes

- [ ] Add new resources

- [X] No

### DIFF
--- a/.github/workflows/get_terraform_version.yaml
+++ b/.github/workflows/get_terraform_version.yaml
@@ -3,7 +3,7 @@ on:
     outputs:
       terraform_version:
         value: ${{ jobs.get-terraform-version.outputs.terraform_version }}
-        description: "List of all modules in the format [\"name1\", \"name2\"]"
+        description: "The terraform version retrieved from the .terraform-version file. If missing, it returns a default recommended version."
 jobs:
   get-terraform-version:
     runs-on: ubuntu-latest
--- a/.github/workflows/get_terraform_version.yaml
+++ b/.github/workflows/get_terraform_version.yaml
@@ -1,0 +1,27 @@
+on:
+  workflow_call:
+    outputs:
+      terraform_version:
+        value: ${{ jobs.get-terraform-version.outputs.terraform_version }}
+        description: "List of all modules in the format [\"name1\", \"name2\"]"
+jobs:
+  get-terraform-version:
+    runs-on: ubuntu-latest
+    name: Get terraform version
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v3
+
+      - name: Determine Terraform version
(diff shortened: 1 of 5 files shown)


.github/workflows/get_terraform_version.yaml
```diff
@@ -0,0 +1,27 @@
...
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2
+
+      - name: Determine Terraform version
+        id: get-terraform-version
+        run: |
+          if [ -f ".terraform-version" ]; then
+              TERRAFORM_VERSION=$(cat .terraform-version)
+              echo "Terraform version found in .terraform-version file: $TERRAFORM_VERSION"
```
gunzip, line 20: if ! [[ "$TERRAFORM_VERSION" =~ ^[0-9]+\.[0-9]+\.[0-9]+$ ]]; then
            echo "Error: Invalid Terraform version format: $TERRAFORM_VERSION"
            exit 1
          fi

.github/workflows/get_terraform_version.yaml
```diff
@@ -0,0 +1,31 @@
...
+        id: get-terraform-version
+        run: |
+          if [ -f ".terraform-version" ]; then
+              TERRAFORM_VERSION=$(cat .terraform-version)
+              if ! [[ "$TERRAFORM_VERSION" =~ ^[0-9]+\.[0-9]+\.[0-9]+$ ]]; then
+                echo "Error: Invalid Terraform version format: $TERRAFORM_VERSION"
+                exit 1
+              fi
+              echo "Terraform version found in .terraform-version file: $TERRAFORM_VERSION"
+          else
```
christian-calabrese, line 25: @pagopa/engineering-team-devex I'm not sure whether to return an error or return a default version in case the user has not provided a terraform version and the .terraform-version file is not present.

I'm probably more polarized to raise an error

gunzip (approved):

.github/workflows/get_terraform_version.yaml
gunzip: a default version would be useful in order to prevent users forced upgrades imho (we will upgrade the default version for them)

Krusty93 (approved):

PR #238, opened 2025-01-27
Add lock to PostgreSQL flexible server module

christian-calabrese:
### List of changes

Add lock to PostgreSQL flexible server module

### Motivation and context

Currently, the module that creates postgresql flexible servers don't create locks. For production workloads is highly suggested to create one.

### Type of changes

- [ ] Add new resources

- [X] No


infra/modules/azure_postgres_server/variables.tf
```diff
@@ -261,4 +261,10 @@ variable "diagnostic_settings" {
     )
     error_message = "log_analytics_workspace_id and diagnostic_setting_destination_storage_id are mandatory if diagnostic is enabled."
   }
+}
+
+variable "needs_lock" {
```
Krusty93, line 266: What do you think?
```suggestion
variable "enable_lock" {
```

christian-calabrese: I like it, fixed!

Krusty93 (approved):

PR #244, opened 2025-01-27
[CES-676] Avoid buffering Terraform output

gunzip:
### List of changes

During Terraform plan and apply avoid buffering
in order to print output in realtime and not only
when the Terraform command exit.

### Motivation and context

This change let's users watch Terraform plan/apply
output without waiting for the end of the command.

### Type of changes

- [ ] Add new resources

- [X] Yes, users may be impacted applying this change
- [ ] No


gunzip:
impossible to obtain since we cannot disable buffering on self host github runners https://github.com/orgs/community/discussions/52284

PR #248, opened 2025-01-28
[CES-702] Don't run unnecessary actions in changeset PRs

christian-calabrese:
### List of changes
Added a condition to avoid the run in changeset PRs

### Motivation and context
Some github workflows are improperly run in changeset PRs (Code Review, Validate GitHub Pages Website, Terraform Validation).
To avoid this, a condition has been added to these workflows in order to discard the run when the source branch of the PR is `changeset-release/main`

### Type of changes

- [ ] Add new resources

- [X] No

### DIFF
--- a/.github/workflows/code-review.yaml
+++ b/.github/workflows/code-review.yaml
@@ -11,7 +11,7 @@ on:
 jobs:
   js_code_review:
     # Don't run in Changeset PRs
-    if: ${{ github.event_name != 'pull_request' || github.event.pull_request.head.ref != 'changeset-release/main' }}
+    if: ${{ github.actor != 'dx-pagopa-bot' }}
     uses: ./.github/workflows/js_code_review.yaml
     name: Code Review
     secrets: inherit
--- a/.github/workflows/code-review.yaml
+++ b/.github/workflows/code-review.yaml
@@ -10,6 +10,8 @@ on:
 
 jobs:
   js_code_review:
+    # Don't run in Changeset PRs
+    if: ${{ github.event_name != 'pull_request' || github.event.pull_request.head.ref != 'changeset-release/main' }}
     uses: ./.github/workflows/js_code_review.yaml
     name: Code Review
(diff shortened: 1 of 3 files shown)


gunzip (commented):
what about checking the actor instead?

something like: `if: ${{ github.actor != 'github-actions[bot]' }}` (I dunno what's the right name)

lucacavallaro (changes requested):
@christian-calabrese I was expecting to see changes only on `.github/workflows` 👀

gunzip (approved):

.github/workflows/code-review.yaml
```diff
@@ -10,6 +10,8 @@ on:
 
 jobs:
   js_code_review:
+    # Don't run in Changeset PRs
+    if: ${{ github.actor != 'dx-pagopa-bot' }}
```
kin0992, line 14: **question**: doing this we are assuming that every team is going to use the devex GitHub Bot to create a release

is this our expected behavior?

kin0992: let's take a look at the developer portal, for example: https://github.com/pagopa/developer-portal/blob/main/.github/workflows/changelog.yaml

if they are going to migrate to this workflow, their workflow is going to be simpler, but the PR is going to be opened by the `dx-pagopa-bot`

christian-calabrese: These three workflows are specific to this repository
The workflow that is actually imported by other projects is `js_code_review.yaml` (e.g.[ io-wallet](https://github.com/pagopa/io-wallet/blob/master/.github/workflows/code-review.yaml), [io-fims](https://github.com/pagopa/io-fims/blob/main/.github/workflows/code-review.yaml))

kin0992 (approved):

gunzip (approved):

lucacavallaro (approved):

Krusty93 (approved):

PR #251, opened 2025-01-29
[CES-324] Updated Core Infra module with new resources (APIM, Cosmos and Storage Account)

mamu0:
### List of changes

In the `azure_core_infra` module the services

- APIM
- Cosmos DB
- Storage Account

have been added, which can now be created during the infrastructure deployment phase

### Motivation and context

The change was necessary to allow all developers to have these resources from the beginning

### Type of changes

- [x] Add new resources

- [x] No

### DIFF
--- a/infra/modules/azure_core_infra/_modules/apim/locals.tf
+++ b/infra/modules/azure_core_infra/_modules/apim/locals.tf
@@ -1,0 +1,15 @@
+locals {
+  apim = {
+    name           = "${var.prefix}-apim-${var.suffix}"
+    autoscale_name = var.tier == "l" ? "${var.prefix}-apim-as-${var.suffix}" : null
+    sku_name = lookup(
+      {
+        "s" = "Developer_1",
+        "m" = "Standard_1",
+        "l" = "Premium_1"
+      },
+      var.tier,
+      "Premium_1" # Default or consider throwing an error
+    )
+  }
+}
--- a/infra/modules/azure_core_infra/_modules/apim/main.tf
+++ b/infra/modules/azure_core_infra/_modules/apim/main.tf
@@ -1,0 +1,103 @@
+resource "azurerm_api_management" "this" {
+  name                          = local.apim.name
+  resource_group_name           = var.resource_group_name
+  location                      = var.location
+  publisher_name                = var.publisher_name
+  publisher_email               = var.publisher_email
+  notification_sender_email     = null
+  sku_name                      = local.apim.sku_name
+  zones                         = var.tier == "l" ? ["1", "2", "3"] : null
+  public_network_access_enabled = var.enable_public_network_access
+
+  # Managed identity type: System
+  identity {
+    type = "SystemAssigned"
+  }
+
+  virtual_network_type = var.virtual_network_type_internal ? "Internal" : "None"
+
+  # subnet
+  dynamic "virtual_network_configuration" {
+    for_each = var.virtual_network_type_internal ? ["dummy"] : []
+    content {
+      subnet_id = azurerm_subnet.snet.id
+    }
+  }
+
+  sign_up {
+    enabled = false
+    terms_of_service {
+      enabled          = false
+      consent_required = false
+    }
+  }
+
+  tags = var.tags
+}
+
+#-------------#
+# Autoscaling #
+#-------------#
+
+# NOTE: only Premium sku support autoscaling
+resource "azurerm_monitor_autoscale_setting" "this" {
+  count               = var.tier == "l" && var.autoscale != null && var.autoscale.enabled ? 1 : 0
+  name                = local.apim.autoscale_name
+  resource_group_name = var.resource_group_name
+  location            = var.location
+  target_resource_id  = azurerm_api_management.this.id
+
+  profile {
+    name = "default"
+
+    capacity {
+      default = var.autoscale.default_instances
+      minimum = var.autoscale.minimum_instances
+      maximum = var.autoscale.maximum_instances
+    }
+
+    rule {
+      metric_trigger {
+        metric_name              = "Capacity"
+        metric_resource_id       = azurerm_api_management.this.id
+        metric_namespace         = "microsoft.apimanagement/service"
+        time_grain               = "PT1M"
+        statistic                = "Average"
+        time_window              = var.autoscale.scale_out_time_window
+        time_aggregation         = "Average"
+        operator                 = "GreaterThan"
+        threshold                = var.autoscale.scale_out_capacity_percentage
+        divide_by_instance_count = false
+      }
+
+      scale_action {
+        direction = "Increase"
+        type      = "ChangeCount"
+        value     = var.autoscale.scale_out_value
+        cooldown  = var.autoscale.scale_out_cooldown
+      }
+    }
+
+    rule {
+      metric_trigger {
+        metric_name              = "Capacity"
+        metric_resource_id       = azurerm_api_management.this.id
+        metric_namespace         = "microsoft.apimanagement/service"
+        time_grain               = "PT1M"
+        statistic                = "Average"
+        time_window              = var.autoscale.scale_in_time_window
+        time_aggregation         = "Average"
+        operator                 = "LessThan"
+        threshold                = var.autoscale.scale_in_capacity_percentage
+        divide_by_instance_count = false
+      }
+
+      scale_action {
+        direction = "Decrease"
+        type      = "ChangeCount"
+        value     = var.autoscale.scale_in_value
+        cooldown  = var.autoscale.scale_in_cooldown
+      }
+    }
+  }
+}
--- a/infra/modules/azure_core_infra/_modules/apim/network.tf
+++ b/infra/modules/azure_core_infra/_modules/apim/network.tf
@@ -1,0 +1,64 @@
+resource "azurerm_subnet" "snet" {
+  name                 = "${var.project}-apim-snet-${var.suffix}"
+  virtual_network_name = var.virtual_network.name
+  resource_group_name  = var.virtual_network.resource_group_name
+  address_prefixes     = [var.subnet_cidr]
+}
+
+# Define the A Records for APIM
+
+resource "azurerm_private_dns_a_record" "apim_azure_api_net" {
+  name                = azurerm_api_management.this.name
+  zone_name           = data.azurerm_private_dns_zone.azure_api_net.name
+  resource_group_name = var.virtual_network.resource_group_name
+  ttl                 = 3600
+  records             = azurerm_api_management.this.private_ip_addresses
+
+  tags = var.tags
+}
+
+resource "azurerm_private_dns_a_record" "apim_management_azure_api_net" {
+  name                = azurerm_api_management.this.name
+  zone_name           = data.azurerm_private_dns_zone.management_azure_api_net.name
+  resource_group_name = var.virtual_network.resource_group_name
+  ttl                 = 3600
+  records             = azurerm_api_management.this.private_ip_addresses
+
+  tags = var.tags
+}
+
+resource "azurerm_private_dns_a_record" "apim_scm_azure_api_net" {
+  name                = azurerm_api_management.this.name
+  zone_name           = data.azurerm_private_dns_zone.scm_azure_api_net.name
+  resource_group_name = var.virtual_network.resource_group_name
+  ttl                 = 3600
+  records             = azurerm_api_management.this.private_ip_addresses
+
+  tags = var.tags
+}
+
+# Define security group
+resource "azurerm_network_security_group" "nsg_apim" {
+  name                = "${var.prefix}-apim-nsg-${var.suffix}"
+  resource_group_name = var.virtual_network.resource_group_name
+  location            = var.location
+
+  security_rule {
+    name                       = "managementapim"
+    priority                   = 100
+    direction                  = "Inbound"
+    access                     = "Allow"
+    protocol                   = "Tcp"
+    source_port_range          = "*"
+    destination_port_range     = "3443"
+    source_address_prefix      = "ApiManagement"
+    destination_address_prefix = "VirtualNetwork"
+  }
+
+  tags = var.tags
+}
+
+resource "azurerm_subnet_network_security_group_association" "snet_nsg" {
+  subnet_id                 = azurerm_subnet.snet.id
+  network_security_group_id = azurerm_network_security_group.nsg_apim.id
+}
--- a/infra/modules/azure_core_infra/_modules/apim/outputs.tf
+++ b/infra/modules/azure_core_infra/_modules/apim/outputs.tf
@@ -1,0 +1,23 @@
+output "id" {
+  value = azurerm_api_management.this.id
+}
+
+output "name" {
+  value = azurerm_api_management.this.name
+}
+
+output "private_ip_addresses" {
+  value = azurerm_api_management.this.private_ip_addresses
+}
+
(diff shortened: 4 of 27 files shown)


infra/modules/azure_core_infra/_modules/apim/locals.tf
```diff
@@ -0,0 +1,15 @@
...
+    name           = "${var.prefix}-apim-${var.suffix}"
+    autoscale_name = var.tier == "l" ? "${var.prefix}-apim-as-${var.suffix}" : null
+    sku_name = lookup(
+      {
+        "s" = "Developer_1",
+        "m" = "Standard_1",
+        "l" = "Premium_1"
+      },
+      var.tier,
+      "Premium_1" # Default or consider throwing an error
```
gunzip, line 12: better to throw an error imho

infra/modules/azure_core_infra/_modules/apim/main.tf
```diff
@@ -0,0 +1,103 @@
+resource "azurerm_api_management" "this" {
+  name                          = local.apim.name
+  resource_group_name           = var.resource_group_name
+  location                      = var.location
+  publisher_name                = var.publisher_name
+  publisher_email               = var.publisher_email
+  notification_sender_email     = null
```
gunzip, line 7: why null?

infra/modules/azure_core_infra/_modules/apim/main.tf
```diff
@@ -0,0 +1,103 @@
+resource "azurerm_api_management" "this" {
```
gunzip, line 1: can we use `azure_api_management` module instead ?

infra/modules/azure_core_infra/_modules/apim/network.tf
```diff
@@ -0,0 +1,64 @@
...
+  tags = var.tags
+}
+
+# Define security group
+resource "azurerm_network_security_group" "nsg_apim" {
+  name                = "${var.prefix}-apim-nsg-${var.suffix}"
+  resource_group_name = var.virtual_network.resource_group_name
+  location            = var.location
+
+  security_rule {
```
gunzip, line 46: what about adding a comment about these settings?

infra/modules/azure_core_infra/_modules/apim/outputs.tf
```diff
@@ -0,0 +1,23 @@
+output "id" {
```
gunzip, line 1: question: why we don't just

```
output "apim" {
  value = azurerm_api_management
}

```

infra/modules/azure_core_infra/_modules/apim/variables.tf
```diff
@@ -0,0 +1,109 @@
...
+
+variable "enable_public_network_access" {
+  type        = bool
+  description = "Enable public network access"
+  default     = false
+}
+
+variable "virtual_network_type_internal" {
+  type        = bool
+  description = "The type of virtual network you want to use, if true it will be Internal and you need to specify a subnet_id, otherwise it will be None"
```
gunzip, line 60:
```suggestion
  description = "If set to true, the virtual network will be internal, requiring you to specify a subnet_id. Otherwise, it will be None"
```

infra/modules/azure_core_infra/_modules/cosmos/main.tf
```diff
@@ -0,0 +1,40 @@
+resource "azurerm_cosmosdb_account" "this" {
```
gunzip, line 1: question: why don't we use [azure_cosmos_account](https://github.com/pagopa/dx/tree/main/infra/modules/azure_cosmos_account) ?

infra/modules/azure_core_infra/_modules/cosmos/outputs.tf
```diff
@@ -0,0 +1,19 @@
+output "name" {
```
gunzip, line 1: same question about apim output

infra/modules/azure_core_infra/_modules/cosmos/variables.tf
```diff
@@ -0,0 +1,79 @@
+variable "prefix" {
+  type        = string
+  description = "env prefix, short environment and short location amd domain"
```
gunzip, line 3: this description is not very clear

infra/modules/azure_core_infra/_modules/cosmos/variables.tf
```diff
@@ -0,0 +1,79 @@
...
+}
+
+variable "suffix" {
+  type        = string
+  description = "the instance number"
```
gunzip, line 8:
```suggestion
  description = "Instance number"
```

infra/modules/azure_core_infra/_modules/cosmos/variables.tf
```diff
@@ -0,0 +1,79 @@
...
+  type        = map(any)
+  description = "Resources tags"
+}
+
+variable "virtual_network" {
+  type = object({
+    name                = string
+    resource_group_name = string
+  })
+  description = "Virtual network in which to create the subnet"
```
gunzip, line 31:
```suggestion
  description = "The virtual network where the subnet will be created"
```

infra/modules/azure_core_infra/_modules/storage_account/main.tf
```diff
@@ -0,0 +1,54 @@
+#tfsec:ignore:azure-storage-queue-services-logging-enabled
+resource "azurerm_storage_account" "this" {
```
gunzip, line 2: can we use https://github.com/pagopa/dx/tree/main/infra/modules/azure_storage_account ?

infra/modules/azure_core_infra/_modules/storage_account/outputs.tf
```diff
@@ -0,0 +1,20 @@
+output "name" {
```
gunzip, line 1: same comment about apim output

infra/modules/azure_core_infra/develop_resources.tf
```diff
@@ -0,0 +1,92 @@
+#----------------#
```
gunzip, line 1: if `azure_core_infra` is supposed to be used both in production and development BUT these new resources are only for development, what about moving them inside a dedicated module? ie. `azure_core_infra_extra` ?

gunzip: moreover: if these resources are only for development, does it make sense to have tiers > 's' ?

infra/modules/azure_core_infra/_modules/apim/main.tf
mamu0: I remember we discussed this in the past. It’s to avoid creating internal dependencies between modules; in this way, the `azure_core_infra` module doesn’t strictly depend on other `dx` modules, making updates or changes less problematic.

Does this make sense to you as well, @pagopa/engineering-team-cloud-eng ?

gunzip: what's the advantage of non creating internal dependency? referencing the module in the registry lets us use semver now.

infra/modules/azure_core_infra/develop_resources.tf
mamu0: Creating a `dx` module that groups other dx modules doesn’t seem like a good idea. If someone needs them, they can simply call the respective modules and create the resources individually.

mamu0: The resources here are not restricted to the `s` tier; other options can be selected. The restriction only applies when the testing flag is set to `true`, but this limitation wouldn’t exist in production.

In any case, the configuration is naturally more limited because it’s a common environment created once at startup, a too complex configuration could become a problem. That’s why I believe using separate modules is the better approach if we have a more specific configuration. For this reason, I made it optional and primarily intended for development.

This last point is just my opinion, but we can discuss it further @pagopa/engineering-team-cloud-eng

gunzip: that's the concept of reusable stacks, higher level abstractions. if we expect that all developement environment follow this settings, it's useful to have a representation of this repeteable environment. if we don't have this expectation I agree it's not useful. do we?

infra/modules/azure_core_infra/_modules/cosmos/main.tf
mamu0: ditto

infra/modules/azure_core_infra/_modules/storage_account/main.tf
mamu0: ditto

infra/modules/azure_core_infra/_modules/apim/main.tf
Krusty93: I think we should agree to a common pattern, as the question come out very often on the latest PRs.

We shall choose either between:
- create a folder like `internal_modules` and create there a bunch of modules reused by the ones actually consumed by the customers. For example, a local Storage Account module could be re-used by several others like Function App, Function App Exposed, Azure Core Infra, and finally the Storage Account. These local modules ARE not meant to be used by customers and should not be pushed to the registry. They simply help us to reuse code, make fixes once and be quicker. This is the option I personally prefer.
- each module could reference other modules: locally, pinning the commit's sha, via registry, exploiting semver. I'm afraid to create circular dependencies and promoting dependency hell conditions
- do not put reference between modules and use resources only (current approach)

gunzip: why it's an issue to reuse public modules internally? I'm assuming we use semver.

mamu0: between the 3 options I would discard the second. I think the first option is the best, even if we would have modules very similar to those exposed, the same goes for the third, but with an even greater risk of creating different copies of the same module.
I would say that the choice of creating a folder of internally recallable modules is the one that creates the least problems, but it must be updated in parallel with the exposed modules.

gunzip: I guess they must be published anyway to be found when the use imports 'public' modules from the registry, isn't it?

mamu0: sorry, for the 2 I had misread, yes by using a commit you avoid the problem of being able to take incompatible versions. But it is not clear to me why we cannot use the registry instead of the commit

gunzip: "via registry" is a viable option

Krusty93: For the second, what if the module design is different than what another module needs? Should we create another variable that could be used externally as well?
Sticking to the storage account example, the function app creates using private networking connectivity but in the meanwhile the function app exposed does not set any network restriction rule. Should we add a variable to control che connectivity to the storage account module then? Are not we supposed to make choices and limit the possibilities of the customers?

gunzip: > For the second, what if the module design is different than what another module needs? Should we create another variable that could be used externally as well?

It depends. If we're confident that the change will have only internal impacts, creating another module could be an option. But... are we dealing with a real case, or is this hypothetical? Ultimately, it's still part of a public module.

The way we organize the code should align with our **maintainability** strategy. Right now, it's difficult to maintain due to _significant_ duplication, which is only expected to increase.

gunzip: Consider also that you can override all settings in derived modules.

ie.

```hcl
resource "azurerm_storage_account" "common" {
  name                     = var.account_name
  resource_group_name      = var.resource_group_name
  location                 = var.location
  account_tier             = var.account_tier
  account_replication_type = "XXX"
}

module "common_storage" {
  source              = "../azure_storage_common"
  account_replication_type = "GRS"  # Override
}

```

Krusty93: In your idea, is `azure_storage_common` supposed to be used by customers? Because overriding setting in derived modules means creating an input variable for each property

gunzip: it probably won't be used by customers as it's a base module. we can adopt some standard naming (better) or folder structure for these modules, but I doubt they can be hidden in the registry since are used by other modules (sources must be available to the end users as well). 

there are some strategies we can adopt if you fear that end users could use the base modules. for example we can use some var that errors in case it's not overwritten in the base module, ie:

```
variable "__DANGEROUSLY_USE_INTERNAL_MODULE" {
  type        = boolean
  description = "This variable must be overridden by derived modules to ensure proper function."
  default     = "ERROR: This module is not meant to be used directly!"
}
```

this does not prevent the base module usage, but at least customers would be aware :-) 

What do you mean with "overriding setting in derived modules means creating an input variable _for each property_". I guess we can just override what differs from base module even without using variables (like the `account_replication_type` property above)

gunzip: An alternative would be to keep reference to "internal modules" though GitHub URLs while public stay on the registry, but you'll lose semantic versioning for internal ones which is undesiderable imho.

mamu0: I think the best approach is to publish all modules in the registry and reference them explicitly by version. This ensures consistency, avoids duplication, and makes maintenance easier with SemVer.

Publishing **internal** modules but discouraging their use seems unnecessary... if they are public, they should be usable. Instead of avoiding dependencies, maybe we should focus on clear documentation.

So a better approach is to create reusable modules where needed and reference them directly, rather than duplicating logic across multiple places.

gunzip: agree 👍

infra/modules/azure_core_infra/develop_resources.tf
mamu0: In this case, the module is meant to create default resources for a subscription, so things that ideally shouldn’t need modifications once deployed. Adding **APIM** makes sense since it’s a _infra_ component, and everything else is built around it. However, for **Cosmos DB** and _Storage Accounts_, I’m not entirely sure, as some instances might be created later for different needs.

It’s important to define which resources are truly fixed and should be created here. Also, making them fully configurable like the original modules adds unnecessary complexity, it would be easier for consumers to call the modules directly if full flexibility is needed. Instead, I’d expect some level of preconfiguration, especially since these are unique resources.

What do you think?

gunzip: My vision is that the `azure_core_infra` module should be usable in _both_ development and production environments. However, the changes in this PR are relevant only for setting up a _development_ environment.  

Once this has been determined, it...
 
1. Allows us to make certain resources fixed (tier: dev) rather than configurable.
1. Encourages the use of a separate module (e.g., `azure_core_infra_dev`) instead of modifying the existing one.  

To achieve this, I’d prefer encapsulating the existing modules and configuring them with development-specific settings (see discussion above).  

What about this approach?

mamu0: The resources are not locked to the development environment, we can rename `develop_resources.tf`. From a configuration perspective, there are no limitations, the entire module is designed to be used in any environment. I don’t see the need to create a separate, restricted version for development.

What I’d like to clarify is which resources we are adding. The inclusion of APIM makes sense, as having one per subscription by default is reasonable. However, for Cosmos DB and Storage Account, what purpose do they serve as default resources? What scope should they have?

At the moment, this module would provision two generic resources without a defined use case. If they are part of the core infrastructure, it would be helpful to understand their intended purpose.

PR #252, opened 2025-01-30
Set fixed-rate logging strategy to AI function runtime, reduce log of type traces

gquadrati:
### Type of changes

- [ ] Add new resources

- [ ] No

### DIFF
--- a/infra/modules/azure_function_app/function_app.tf
+++ b/infra/modules/azure_function_app/function_app.tf
@@ -50,8 +50,16 @@ resource "azurerm_linux_function_app" "this" {
     length(local.function_app.name) > 32 && !(contains(keys(var.app_settings), "AzureFunctionsWebHost__hostid")) ? { AzureFunctionsWebHost__hostid = "production" } : {},
     var.app_settings,
     local.application_insights.enable ? {
+      # SDK AI Sampling (for dependencies)
       # https://docs.microsoft.com/en-us/azure/azure-monitor/app/sampling
-      APPINSIGHTS_SAMPLING_PERCENTAGE = var.application_insights_sampling_percentage
+      APPINSIGHTS_SAMPLING_PERCENTAGE = var.application_insights_sampling_percentage,
+
+      # Runtime AI Sampling (for requests, traces, metrics, etc..)
+      # https://learn.microsoft.com/en-us/azure/azure-functions/configure-monitoring?tabs=v2#overriding-monitoring-configuration-at-runtime
+      AzureFunctionsJobHost__logging__applicationInsights__samplingSettings__maxSamplingPercentage     = var.application_insights_sampling_percentage,
+      AzureFunctionsJobHost__logging__applicationInsights__samplingSettings__initialSamplingPercentage = var.application_insights_sampling_percentage,
+      AzureFunctionsJobHost__logging__logLevel__default                                                = "Information"
+      AzureFunctionsJobHost__logging__logLevel__Function                                               = "Warning"
     } : {},
     local.function_app.has_durable == 1 ? {
       #https://learn.microsoft.com/en-us/azure/azure-functions/durable/durable-functions-storage-providers#hostjson-configuration
@@ -64,7 +72,11 @@ resource "azurerm_linux_function_app" "this" {
       [
         "SLOT_TASK_HUBNAME",
         "APPINSIGHTS_SAMPLING_PERCENTAGE",
-        "AzureFunctionsWebHost__hostid"
+        "AzureFunctionsWebHost__hostid",
+        "AzureFunctionsJobHost__logging__applicationInsights__samplingSettings__maxSamplingPercentage",
+        "AzureFunctionsJobHost__logging__applicationInsights__samplingSettings__initialSamplingPercentage",
+        "AzureFunctionsJobHost__logging__logLevel__default",
+        "AzureFunctionsJobHost__logging__logLevel__Function"
       ],
       var.sticky_app_setting_names,
     )
--- a/infra/modules/azure_function_app/function_app_slot.tf
+++ b/infra/modules/azure_function_app/function_app_slot.tf
@@ -46,8 +46,14 @@ resource "azurerm_linux_function_app_slot" "this" {
       FUNCTIONS_WORKER_PROCESS_COUNT = local.function_app.worker_process_count,
     },
     local.application_insights.enable ? {
+      # SDK AI Sampling (for dependencies)
       # https://docs.microsoft.com/en-us/azure/azure-monitor/app/sampling
-      APPINSIGHTS_SAMPLING_PERCENTAGE = "100"
+      APPINSIGHTS_SAMPLING_PERCENTAGE = "100",
+
+      # Runtime AI Sampling (for requests, traces, metrics, etc..)
+      AzureFunctionsJobHost__logging__applicationInsights__samplingSettings__maxSamplingPercentage     = "100",
+      AzureFunctionsJobHost__logging__applicationInsights__samplingSettings__initialSamplingPercentage = "100",
+      AzureFunctionsJobHost__logging__logLevel__default                                                = "Debug"
     } : {},
     # https://learn.microsoft.com/en-us/azure/azure-functions/errors-diagnostics/diagnostic-events/azfd0004#options-for-addressing-collisions
     length("${azurerm_linux_function_app.this.name}-${local.function_app_slot.name}") > 32 && !(contains(keys(var.slot_app_settings), "AzureFunctionsWebHost__hostid")) ? { AzureFunctionsWebHost__hostid = local.function_app_slot.name } : {},


infra/modules/azure_function_app/function_app.tf
```diff
@@ -50,8 +50,16 @@ resource "azurerm_linux_function_app" "this" {
     length(local.function_app.name) > 32 && !(contains(keys(var.app_settings), "AzureFunctionsWebHost__hostid")) ? { AzureFunctionsWebHost__hostid = "production" } : {},
     var.app_settings,
     local.application_insights.enable ? {
+      # SDK AI Sampling (for dependencies)
       # https://docs.microsoft.com/en-us/azure/azure-monitor/app/sampling
-      APPINSIGHTS_SAMPLING_PERCENTAGE = var.application_insights_sampling_percentage
+      APPINSIGHTS_SAMPLING_PERCENTAGE = var.application_insights_sampling_percentage,
+
+      # Runtime AI Sampling (for requests, traces, metrics, etc..)
```
gunzip, line 57:
```suggestion
      # Azure Function Host (runtime) AI Sampling (for requests, traces, metrics, etc..)
```

For the casual reader 'Runtime' may refer to some 'runtime' behavior.

infra/modules/azure_function_app/function_app.tf
```diff
@@ -50,8 +50,16 @@ resource "azurerm_linux_function_app" "this" {
...
+      # https://learn.microsoft.com/en-us/azure/azure-functions/configure-monitoring?tabs=v2#overriding-monitoring-configuration-at-runtime
+      AzureFunctionsJobHost__logging__applicationInsights__samplingSettings__maxSamplingPercentage     = var.application_insights_sampling_percentage,
+      AzureFunctionsJobHost__logging__applicationInsights__samplingSettings__initialSamplingPercentage = var.application_insights_sampling_percentage,
+      AzureFunctionsJobHost__logging__logLevel__default                                                = "Information"
+      AzureFunctionsJobHost__logging__logLevel__Function                                               = "Warning"
```
gunzip, line 62: this will set at 'Information' (which is somewhat verbose) level even categories like 'Microsoft'. is this expected?

infra/modules/azure_function_app/function_app_slot.tf
```diff
@@ -46,8 +46,14 @@ resource "azurerm_linux_function_app_slot" "this" {
       FUNCTIONS_WORKER_PROCESS_COUNT = local.function_app.worker_process_count,
     },
     local.application_insights.enable ? {
+      # SDK AI Sampling (for dependencies)
       # https://docs.microsoft.com/en-us/azure/azure-monitor/app/sampling
-      APPINSIGHTS_SAMPLING_PERCENTAGE = "100"
+      APPINSIGHTS_SAMPLING_PERCENTAGE = "100",
```
gunzip, line 51: why 100?

gunzip (commented):
Must be backpointet into `azure_function_app_exposed` module as well

infra/modules/azure_function_app/function_app_slot.tf
gunzip: ref. https://pagopa.atlassian.net/browse/CES-704

infra/modules/azure_function_app/function_app_slot.tf
```diff
@@ -46,8 +46,14 @@ resource "azurerm_linux_function_app_slot" "this" {
...
+
+      # Runtime AI Sampling (for requests, traces, metrics, etc..)
+      AzureFunctionsJobHost__logging__applicationInsights__samplingSettings__maxSamplingPercentage     = "100",
+      AzureFunctionsJobHost__logging__applicationInsights__samplingSettings__initialSamplingPercentage = "100",
+      AzureFunctionsJobHost__logging__logLevel__default                                                = "Debug"
```
gunzip, line 56:
```suggestion
      AzureFunctionsJobHost__logging__logLevel__default                                                = "Information"
```

infra/modules/azure_function_app/function_app.tf
gquadrati: "Information" level is the default one, and it's needed for logging requests, metrics and other useful info. See documentation [here](https://learn.microsoft.com/en-us/azure/azure-functions/configure-monitoring?tabs=v2#configure-categories)
That's why we left it as default and we override the Function logLevel.

gunzip: Where in the docs did you find that 'Information' is the default level? I cannot find anything related to default values in the page you've shared. 

What about set "Microsft" to 'Warning' anyway?

gunzip: Where in the docs did you find that 'Information' is the default level? I cannot find anything related to default values in the page you've shared. 

What about set "Microsft" to 'Warning' anyway?

Krusty93:
N.B. If the changes will be approved, they should be backported in these other modules:
- function app exposed
- app service (?)
- app service exposed (?)

infra/modules/azure_function_app/function_app_slot.tf
```diff
@@ -46,8 +46,15 @@ resource "azurerm_linux_function_app_slot" "this" {
       FUNCTIONS_WORKER_PROCESS_COUNT = local.function_app.worker_process_count,
     },
     local.application_insights.enable ? {
+      # SDK AI Sampling (for dependencies)
       # https://docs.microsoft.com/en-us/azure/azure-monitor/app/sampling
-      APPINSIGHTS_SAMPLING_PERCENTAGE = "100"
+      APPINSIGHTS_SAMPLING_PERCENTAGE = "100",
```
gunzip, line 51: @Krusty93 we should let this be configurable as well (same defaults as production?)

infra/modules/azure_function_app/function_app.tf
gquadrati: > Where in the docs did you find that 'Information' is the default level? I cannot find anything related to default values in the page you've shared.

Actually it's not specified what the default is, but the documentation says that setting the default to `Warning` will not log the requests (Host.Results) neither the metrics (Host.Aggregator).

Anyway I agree that we can set `Microsoft` to `Warning`.

infra/modules/azure_function_app/function_app_slot.tf
gquadrati: Is this because of the canary deployment?

infra/modules/azure_function_app/function_app.tf
gquadrati: @gunzip I'd set `Worker` to `Warning` as well, what do you think?

gunzip: I don't know how this effects traces sent from AI SDK, we should take a look.

infra/modules/azure_function_app/function_app.tf
```diff
@@ -50,8 +50,17 @@ resource "azurerm_linux_function_app" "this" {
...
     var.app_settings,
     local.application_insights.enable ? {
+      # SDK AI Sampling (for dependencies)
       # https://docs.microsoft.com/en-us/azure/azure-monitor/app/sampling
-      APPINSIGHTS_SAMPLING_PERCENTAGE = var.application_insights_sampling_percentage
+      APPINSIGHTS_SAMPLING_PERCENTAGE = var.application_insights_sampling_percentage,
+
+      # Azure Function Host (runtime) AI Sampling (for requests, traces, metrics, etc..)
+      # https://learn.microsoft.com/en-us/azure/azure-functions/configure-monitoring?tabs=v2#overriding-monitoring-configuration-at-runtime
+      AzureFunctionsJobHost__logging__applicationInsights__samplingSettings__minSamplingPercentage     = var.application_insights_sampling_percentage,
```
gunzip, line 59: do we need `isEnabled=true` as well?

gquadrati: If we want to let appinsights be enabled by default, sure

gunzip:
`draft` since the PR requires tests using 2.x AI SDK to check if this is a breaking change or not.

infra/modules/azure_function_app/function_app.tf
gunzip: it's enabled by default so we don't need to

infra/modules/azure_function_app/function_app.tf
gunzip: I thought about this a little bit more. What about stick to defaults for `logLevel`? We may just set sampling percentage with this PR.

infra/modules/azure_function_app/function_app_slot.tf
gunzip: Dunno. I guess it won't be affected anyway.

BurnedMarshal: For the canary deployment we what that all the API call must be logged, a single failure in a single endpoint with a very low request rate is crucial.
Es. the service has 2 API, the first with a call rate of 1k req/min the second with 1 req/min. The sampling will make the canary monitoring on the second inefficient.

gunzip:
I'm unsure about the 'patch' changeset since this wil require a restart once applied. Maybe it would be better to use a major release?

kin0992:
> I'm unsure about the 'patch' changeset since this wil require a restart once applied. Maybe it would be better to use a major release?

@gunzip In my opinion, this qualifies as a `patch` since the scope is limited to the module, and the change does not impact any running resources.  

Are you assuming that the reference to this Terraform module follows the pattern `~> x`? If so, it's the responsibility of the team using this module to ensure that the feature is properly propagated to the production environment.

From this module's perspective, it’s simply a `patch` to me.

kin0992 (approved):
Well done!

PR #254, opened 2025-01-31
[CES-706] In Azure Github Env Bootstrap module Infra CD identity can manage IAM roles

Krusty93:
### List of changes

Add `Role Based Access Control Administrator` role at subscription scope to Infra CD identity in `azure_github_environment_bootstrap`

### Motivation and context

This change is due when a team has more than one resource group. In this scenario, in fact, it is necessary that the GitHub pipeline can assign permissions to the new resource

### Type of changes

- [ ] Add new resources

- [X] No

### DIFF
--- a/infra/modules/azure_github_environment_bootstrap/id_infra_cd_iam.tf
+++ b/infra/modules/azure_github_environment_bootstrap/id_infra_cd_iam.tf
@@ -6,6 +6,13 @@ resource "azurerm_role_assignment" "infra_cd_subscription_reader" {
   description          = "Allow ${var.repository.name} Infra CD identity to read resources at subscription scope"
 }
 
+resource "azurerm_role_assignment" "infra_cd_subscription_rbac_admin" {
+  scope                = var.subscription_id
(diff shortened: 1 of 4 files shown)


infra/modules/azure_github_environment_bootstrap/id_infra_cd_iam.tf
```diff
@@ -6,6 +6,13 @@ resource "azurerm_role_assignment" "infra_cd_subscription_reader" {
   description          = "Allow ${var.repository.name} Infra CD identity to read resources at subscription scope"
 }
 
+resource "azurerm_role_assignment" "infra_cd_subscription_rbac_admin" {
```
gunzip, line 9: this widens the actual rights to everyone who's able to run CD, isn't it?

Krusty93: Yes, but at least a PR is needed. I don't like it, but having a common subscription between all domain teams I haven't found a better approach. If you have some idea, they're welcome.

gunzip (approved):

PR #258, opened 2025-01-31
[CES-636] Added API Management list secrets Custom role to CI

mamu0:
### List of changes

Added [this](https://github.com/pagopa/eng-azure-governance/blob/main/src/01_custom_roles/01_apim_list_secrets.tf) role into CI default configuration.

### Motivation and context

To avoid Contributor built in role for CI.

### Type of changes

- [ ] Add new resources

- [ ] No


gunzip (approved):

Krusty93 (changes requested):
I believe these changes must be backported into [azure_github_environment_bootstrap](https://github.com/pagopa/dx/tree/main/infra/modules/azure_github_environment_bootstrap) as well?

mamu0:
> I believe these changes must be backported into [azure_github_environment_bootstrap](https://github.com/pagopa/dx/tree/main/infra/modules/azure_github_environment_bootstrap) as well?

You're right! I've added it in latest commit

infra/modules/azure_github_environment_bootstrap/id_infra_ci_iam.tf
```diff
@@ -89,3 +89,11 @@ resource "azurerm_role_assignment" "infra_ci_rg_ext_pagopa_dns_reader" {
   principal_id         = azurerm_user_assigned_identity.infra_ci.principal_id
   description          = "Allow ${var.repository.name} Infra CI identity to read DNS Zone records at resource group level"
 }
+
+# API Management
+resource "azurerm_role_assignment" "infra_ci_subscription_apim_secrets" {
```
Krusty93, line 94: Thanks for adding! However, we already have an optional reference to the subscription APIM as variable, you can use the calculated count property like it has been done here:
https://github.com/pagopa/dx/blob/ac3c90b61c0870a62cb4f5c04e366716d719f56e/infra/modules/azure_github_environment_bootstrap/id_infra_cd_iam.tf#L48

Krusty93 (approved):

PR #260, opened 2025-02-03
[CES-712] Get pre-commit terraform version from file in static analysis workflow

### LINKED ISSUE #256: [Static Analysis worksflow] Retrieve pre-commit version from .pre-commit-config.yaml
Can you make [pre_commit_tf_tag](https://github.com/pagopa/dx/blob/main/.github/workflows/static_analysis.yaml#L10) input optional and then add the "same" logic as you did with [terraform_version](https://github.com/pagopa/dx/blob/main/.github/workflows/static_analysis.yaml#L6C7-L6C24) input?

What i mean is to add a new action, such as [get-terraform-version](https://github.com/pagopa/dx/blob/main/.github/actions/get-terraform-version/action.yaml), to retrieve the pre-commit version from `.pre-commit-config.yaml` file and use it inside [static_analysis.yaml](https://github.com/pagopa/dx/blob/main/.github/workflows/static_analysis.yaml) workflow if `pre_commit_tf_tag` input is missing.

What do you think about that?

christian-calabrese:
### List of changes

Retrieving pre-commit docker image version from the .pre-commit-config.yaml file
### Motivation and context

resolves #256 
### Type of changes

- [ ] Add new resources

- [X] No

### DIFF
--- a/.github/actions/get-pre-commit-version/action.yaml
+++ b/.github/actions/get-pre-commit-version/action.yaml
@@ -8,12 +8,10 @@ inputs:
 outputs:
   pre_commit_terraform_version:
     description: "The pre-commit-terraform version retrieved or the default."
-    value: ${{ steps.get-version.outputs.pre_commit_terraform_version }}
 runs:
   using: "composite"
   steps:
     - name: Get Pre-Commit Terraform Version
-      id: get-version
       shell: bash
       run: |
         # Default version from input
@@ -41,10 +39,5 @@ runs:
           fi
         fi
 
-        # Save to GITHUB_ENV first (correct for composite actions)
-        echo "pre_commit_terraform_version=$PRE_COMMIT_VERSION" >> "$GITHUB_ENV"
-
-    - name: Set Output
-      id: set-output
-      shell: bash
-      run: echo "pre_commit_terraform_version=${{ env.pre_commit_terraform_version }}" >> "$GITHUB_OUTPUT"
+        # Output the pre-commit Terraform version
+        echo "pre_commit_terraform_version=$PRE_COMMIT_VERSION" >> $GITHUB_OUTPUT
--- a/.github/actions/get-pre-commit-version/action.yaml
+++ b/.github/actions/get-pre-commit-version/action.yaml
@@ -1,0 +1,38 @@
+name: Get Pre-Commit Terraform Version
+description: Retrieve the `pre-commit-terraform` version from the `.pre-commit-config.yaml` file or return a default version.
+inputs:
+  default_version:
+    description: "Default pre-commit-terraform version to use if the `.pre-commit-config.yaml` file is not found."
+    required: false
+    default: "v1.96.2"
+outputs:
+  pre_commit_terraform_version:
+    description: "The pre-commit-terraform version retrieved or the default."
+runs:
+  using: "composite"
+  steps:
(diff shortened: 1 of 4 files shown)


.github/actions/get-pre-commit-version/action.yaml
```diff
@@ -0,0 +1,38 @@
...
+outputs:
+  pre_commit_terraform_version:
+    description: "The pre-commit-terraform version retrieved or the default."
+runs:
+  using: "composite"
+  steps:
+    - name: Get Pre-Commit Terraform Version
+      shell: bash
+      run: |
+        # Default version from input
```
gunzip, line 17:
```suggestion
if ! command -v yq &>/dev/null; then
  echo "Error: yq command not found. Please ensure yq is installed."
  PRE_COMMIT_VERSION="$DEFAULT_VERSION"
else
  # existing extraction logic…
fi

        # Default version from input
```

.github/actions/get-pre-commit-version/action.yaml
```diff
@@ -0,0 +1,38 @@
...
+        # Check if the .pre-commit-config.yaml file exists
+        if [ -f ".pre-commit-config.yaml" ]; then
+            # Extract the version using yq with partial match
+            PRE_COMMIT_VERSION=$(yq '.repos[] | select(.repo | contains("antonbabenko/pre-commit-terraform")) | .rev' .pre-commit-config.yaml)
+
+            # Validate the version format (must start with "v" followed by digits)
+            if [[ "$PRE_COMMIT_VERSION" =~ ^v[0-9]+\.[0-9]+\.[0-9]+$ ]]; then
+                echo "Pre-commit Terraform version found: $PRE_COMMIT_VERSION"
+            else
+                echo "Error: Invalid or missing Pre-commit Terraform version. Using default: $DEFAULT_VERSION"
```
gunzip, line 29:
```suggestion
                echo "Error: Invalid or missing Pre-commit Terraform version. Using default: ${DEFAULT_VERSION}"
```

gunzip (approved):

.github/actions/get-pre-commit-version/action.yaml
```diff
@@ -0,0 +1,43 @@
+name: Get Pre-Commit Terraform Version
+description: Retrieve the `pre-commit-terraform` version from the `.pre-commit-config.yaml` file or return a default version.
+inputs:
+  default_version:
+    description: "Default pre-commit-terraform version to use if the `.pre-commit-config.yaml` file is not found."
+    required: false
+    default: "v1.96.2"
```
mamu0, line 7: In the default is not needed digest?
`v1.96.2@sha256:01f870b7689b5a09c1a370914fcddcac42c4b6478c9d369e1d2590dd0a66ffd0`

christian-calabrese: To make it work it's not needed indeed
If we'd like to add it in order to pin the docker image version, we should find a way to extract it from the ghcr repository

Krusty93: @mamu0 did you do a similar job to get the sha of the github docker image [here](https://github.com/pagopa/github-self-hosted-runner-azure)?

mamu0 (approved):

PR #262, opened 2025-02-03
[CES-718] Removed provider definition inside Storage Account module

mamu0:
### List of changes

Removed provider definition into storage account module, because it is inherited from calling project
Updated test inside module

### Motivation and context

in a module it is not necessary to define the provider as this is defined in the project that will instantiate the module

### Type of changes

- [ ] Add new resources

- [x] No

### DIFF
--- a/.changeset/short-masks-dress.md
+++ b/.changeset/short-masks-dress.md
@@ -1,0 +1,5 @@
+---
+"azure_storage_account": patch
+---
+
+Removed provider definition and test updated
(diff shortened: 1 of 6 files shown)


gunzip (approved):

gunzip:
it looks like trivy validation doesn't pass https://github.com/pagopa/dx/actions/runs/13117606907/job/36595646294?pr=262#step:7:29

christian-calabrese (approved):

PR #278, opened 2025-02-06
[CES-726] Fix terraform tests that fail due to incorrect tags and resource groups usage

christian-calabrese:
### List of changes

- Fixed tags to comply with the new tag policy
- All tests now use a central resource group instead of creating new ones

### Motivation and context

Terraform tests fail due to incorrect tags and resource groups usage

### Type of changes

- [ ] Add new resources

- [ ] No

### DIFF
--- a/infra/modules/azure_app_service/tests/setup/main.tf
+++ b/infra/modules/azure_app_service/tests/setup/main.tf
@@ -32,7 +32,7 @@ data "azurerm_subnet" "pep" {
 }
 
 data "azurerm_resource_group" "rg" {
-  name = "${var.environment.prefix}-${var.environment.env_short}-itn-${var.environment.app_name}-rg-${module.naming_convention.suffix}"
+  name = "${var.environment.prefix}-${var.environment.env_short}-itn-test-rg-${module.naming_convention.suffix}"
 
 }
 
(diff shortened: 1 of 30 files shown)


infra/modules/azure_app_service/tests/setup/main.tf
```diff
@@ -31,17 +31,17 @@ data "azurerm_subnet" "pep" {
   resource_group_name  = "${module.naming_convention.project}-network-rg-01"
 }
 
-resource "azurerm_resource_group" "rg" {
-  name     = "${module.naming_convention.prefix}-rg-app-${module.naming_convention.suffix}"
-  location = var.environment.location
+data "azurerm_resource_group" "rg" {
+  name = "${var.environment.prefix}-${var.environment.env_short}-itn-${var.environment.app_name}-rg-${module.naming_convention.suffix}"
```
mamu0, line 35: It is probably better to avoid the dependency on the **app_name** `${var.environment.app_name}`, because the rg name is defined separately
```suggestion
  name = "${var.environment.prefix}-${var.environment.env_short}-itn-test-rg-${module.naming_convention.suffix}"

```

christian-calabrese: Done!

mamu0 (approved):

PR #287, opened 2025-02-17
Add new DX modules for Container App and Container App Environment

mamu0:
### List of changes

Added new modules for container app and container app environment.
Minor change in core module with new Private DNS Zone for container app.

### Type of changes

- [x] Add new resources

- [x] No

### Other information

Resolves: #CES-739


infra/modules/azure_container_app/container_app.tf
```diff
@@ -0,0 +1,171 @@
...
+    name                  = "Consumption"
+    workload_profile_type = "Consumption"
+    minimum_count         = 1
+    maximum_count         = 1
+  }
+
+  zone_redundancy_enabled = local.sku.zone_redundancy
+
+  lifecycle {
+    ignore_changes = [
```
Krusty93, line 22: Why this block?

infra/modules/azure_container_app/container_app.tf
```diff
@@ -0,0 +1,171 @@
...
+  }
+
+  tags = var.tags
+}
+
+resource "azurerm_container_app" "this" {
+  name                         = "${module.naming_convention.prefix}-ca-${module.naming_convention.suffix}"
+  container_app_environment_id = !var.create_container_app_environment ? var.container_app_environment_id : azurerm_container_app_environment.this[0].id
+  resource_group_name          = var.resource_group_name
+  revision_mode                = "Single"
```
Krusty93, line 39: (Gonna be a long discussion)
If the revision mode is set on Single, devs can't use A/B testing or multiple "deployment slots" (to use AppService jargon).

Probably, we need to support Multiple by default but first we need a GitHub workflow able to manage multiple slots. I know there are different strategies which move the responsibility more or less on the TF code than the workflow. I think we need to take the decision right now at the same design phase

infra/modules/azure_container_app/container_app.tf
```diff
@@ -0,0 +1,171 @@
...
+    allow_insecure_connections = false
+    external_enabled           = true
+    target_port                = 8080
+    traffic_weight {
+      percentage      = 100
+      latest_revision = true
+    }
+  }
+
+  dynamic "registry" {
```
Krusty93, line 57: We never use private registry, everything is hosted on public GitHub registry

infra/modules/azure_container_app/locals.tf
```diff
@@ -0,0 +1,44 @@
...
+        max = 1
+      }
+      zone_redundancy = true
+    }
+    "xl" = {
+      cpu : 1.25
+      memory : "2.5Gi"
+      replicas = {
+        min = 1
+        max = 2
```
Krusty93, line 35: I'm not sure about these min/max values

Shall we give more flexibility?

infra/modules/azure_container_app/main.tf
```diff
@@ -0,0 +1,22 @@
+terraform {
+  required_providers {
+    azurerm = {
+      source  = "hashicorp/azurerm"
+      version = ">= 3.110, < 5.0"
```
Krusty93, line 5: As this is a new module, would it make sense to use `~> 4` constraint?

infra/modules/azure_container_app/outputs.tf
```diff
@@ -0,0 +1,15 @@
+output "name" {
```
Krusty93, line 1: I suggest to group everything under the same object and add another object for cae values

infra/modules/azure_container_app/variables.tf
```diff
@@ -0,0 +1,207 @@
...
+  description = "Values which are used to generate resource names and location short names. They are all mandatory except for domain, which should not be used only in the case of a resource used by multiple domains."
+}
+
+variable "resource_group_name" {
+  type        = string
+  description = "Resource group to deploy resources to"
+}
+# ------------ CONTAINER ENVIRONMENT ------------ #
+
+variable "create_container_app_environment" {
```
Krusty93, line 26: Elsewhere within this module and in other modules as well, we didn't use these kind of flags to enable/disable components creation, but instead they are based on variable providing the optional value.
In other words, to keep the same behaviour I suggest to delete this variable and calculate conditions on the `container_app_environment_id` actual value

infra/modules/azure_container_app/variables.tf
```diff
@@ -0,0 +1,207 @@
...
+  default     = false
+}
+
+variable "container_app_environment_id" {
+  type        = string
+  description = "The ID of the container app environment to deploy the container app to. If not provided, a new container app environment will be created."
+  default     = null
+}
+
+variable "log_analytics_workspace_id" {
```
Krusty93, line 38: This one should be mandatory

infra/modules/azure_container_app/variables.tf
```diff
@@ -0,0 +1,207 @@
...
+  description = "The offer type for the Container. Valid values are 's', 'm', 'l' and 'xl'."
+  default     = "s"
+
+  validation {
+    condition     = contains(["s", "m", "l", "xl"], var.tier)
+    error_message = "Valid values for tier are 's', 'm', 'l' and 'xl'."
+  }
+}
+
+variable "container_app_template" {
```
Krusty93, line 57: We could also add support to have multiple containers on the same app. What do you guys think?

infra/modules/azure_container_app/variables.tf
```diff
@@ -0,0 +1,207 @@
...
+  type = object({
+    image = string
+    name  = optional(string, "")
+    envs  = optional(map(string), {})
```
Krusty93, line 61: Nothing wrong here, I simply want to ask a question.
Does it worth to call this variable `app_settings` in order to keep the same nomenclature devs already know?

infra/modules/azure_container_app/locals.tf
gunzip: what about limit the module to 3 tiers leaving space in future for `xs` and `xl`?

infra/modules/azure_container_app/variables.tf
gunzip: it depends if we expect that `app_settings` contains sensitive value or not. in the second case I'd add a 'public' prefix to make it clear

infra/modules/azure_container_app/examples/simple/main.tf
```diff
@@ -0,0 +1,34 @@
...
+  resource_group_name  = "${local.project}-network-rg-01"
+}
+
+module "container_app" {
+  source = "../../"
+
+  environment         = local.environment
+  resource_group_name = azurerm_resource_group.example.name
+
+  create_container_app_environment = true
```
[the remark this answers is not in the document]
mamu0: Initially, I had implemented it that way, but during testing, I encountered a limitation: when running _terraform apply_ without explicitly setting `container_app_environment_id`, Terraform fails because it cannot determine whether the ID will be `null` or not, that's because the id is `(known after apply)`.

infra/modules/azure_container_app/variables.tf
gunzip: it depends. if we don't expect that these settings contain sensitive values I'd add a `public` prefix to make it clear

infra/modules/azure_container_app/examples/complete/main.tf
```diff
@@ -0,0 +1,45 @@
...
+    image = "nginx:latest"
+    name  = "nginx"
+
+    envs = {
+      key1 = "value1"
+      key2 = "value2"
+    }
+  }
+
+  key_vault = {
```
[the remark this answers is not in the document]
mamu0: oh this is a mistake, I had removed all the management of the key vault because we can access a keyvault as in other cases with `role_assignment`, and here it should not be necessary to configure a secret

infra/modules/azure_container_app/examples/complete/main.tf
```diff
@@ -0,0 +1,45 @@
...
+
+  container_app_environment_id = azurerm_container_app_environment.example.id
+
+  tier = "m"
+
+  container_app_template = {
+    image = "nginx:latest"
+    name  = "nginx"
+
+    envs = {
```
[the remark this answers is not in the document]
mamu0: why do you say to specify public?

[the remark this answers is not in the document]
mamu0: what do you mean? there are no secrets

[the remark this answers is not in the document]
gunzip: i wonder if we should consider to have different variables for secrets vs public settings since sensitive values make the whole object sensitive and values hidden

infra/modules/azure_container_app/container_app.tf
mamu0: perfect, then I remove it 👍

infra/modules/azure_container_app/variables.tf
mamu0: Initially, I followed that approach for consistency, but during testing, I encountered a limitation: when `container_app_environment_id` is not explicitly set, Terraform fails during apply because it cannot determine in advance whether the value will be null or not.

infra/modules/azure_container_app/variables.tf
mamu0: the variable can be transformed into a `list` and then used `dynamic` in the resource, but at the configuration level by the user it would become more complex, because the `probes` must be placed in this same variable making it quite large

infra/modules/azure_container_app/examples/complete/main.tf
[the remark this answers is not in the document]
mamu0: ah ok now I understand, we can split the variable in two, but then it would still have to be merged at resource level, so if there was even just one secret it would still result in the whole block being sensitive

infra/modules/azure_container_app/locals.tf
mamu0: yes we can change the values, what do you suggest for the different tiers?

infra/modules/azure_container_app/examples/complete/main.tf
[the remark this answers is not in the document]
gunzip: then it's ok to keep it as is but I agree with @Krusty93 in renaming this to `app_settings`

infra/modules/azure_container_app/container_app.tf
gunzip: what about defaulting to single only for tiers <= `s` ?

infra/modules/azure_container_app/locals.tf
gunzip: we should assess the ccurrently used settings, for example for self-care https://github.com/search?q=repo%3Apagopa%2Fselfcare-external-api-backend+cpu&type=code

gunzip: small looks thin, what about:

```
"s" = {
      cpu    = 0.5
      memory = "1Gi"
      replicas = {
        min = 0
        max = 1
      }
    }
    "m" = {
      cpu    = 1
      memory = "2Gi"
      replicas = {
        min = 1
        max = 1
      }
    }
    "l" = {
      cpu : 1.25
      memory : "2.5Gi"
      replicas = {
        min = 1
        max = 2
      }
    }
```

infra/modules/azure_container_app/examples/simple/main.tf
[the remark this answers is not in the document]
Krusty93: It's a weird behaviour. What's the specific error?

infra/modules/azure_container_app/locals.tf
mamu0: i suggest to have a configuration with
```
      cpu    = 0.25
      memory = "0.5Gi"
```
maybe an `xs` only for tests, what do you think?

infra/modules/azure_container_app/container_app.tf
mamu0: from what I see now all the container apps created have `Single` as the review mode, maybe we can only add a `revision_mode` variable with `Single` as default

infra/modules/azure_container_app/examples/simple/main.tf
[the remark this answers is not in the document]
mamu0: I'm closing this conversation because there's another with more details one on the same topic

infra/modules/azure_container_app/locals.tf
mamu0: I modified the tiers as @gunzip suggested, but added `xs` for testing

infra/modules/azure_container_app/container_app.tf
```diff
@@ -0,0 +1,149 @@
+resource "azurerm_container_app_environment" "this" {
+  count = var.create_container_app_environment ? 1 : 0
+
+  name                       = "${module.naming_convention.prefix}-cae-${module.naming_convention.suffix}"
```
Krusty93, line 4: Does the `prefix` contain the `app_name`? I'm worried that if a team domain wants to build multiple environments they'll end up in naming collision

infra/modules/azure_container_app/container_app.tf
lucacavallaro: What are the drawbacks of choosing `Multiple` as default?

mamu0: If I understand correctly, even if `Multiple` should become the standard today we still do not have the incremental rollout workflow. With a variable set by default to `Single` it is true that it will then have to be modified, but I believe it is only an initial effort, then once the strategy is consolidated it can also be removed and set to `Multiple`.
If we want, we can also consider making the variable and setting it to `Multiple` by default, but I'm not sure how much it could impact the costs

infra/modules/azure_container_app/container_app.tf
mamu0: yes `prefix` contain `app_name`, we want to use `project` and add a `name` variable?

infra/modules/azure_container_app/container_app.tf
Krusty93: > What are the drawbacks of choosing `Multiple` as default?
You can have multiple versions running at the same time serving production traffic, both automations and humans should be aware of this and act properly to avoid surprises

> If I understand correctly, even if `Multiple` should become the standard today we still do not have the incremental rollout workflow.
Exactly, for this reason I suggest to add the workflow into this PR...

> With a variable set by default to Single it is true that it will then have to be modified, but I believe it is only an initial effort, then once the strategy is consolidated it can also be removed and set to Multiple.
... especially because the next step after this PR is making the workflow

infra/modules/azure_container_app/variables.tf
Krusty93: I had a try and I had the same error with the conditional access to `data` blocks. Any idea @christian-calabrese ?

infra/modules/azure_container_app/container_app.tf
Krusty93: This made me think to a bigger issue.

If a team uses CAE, it is likely to put multiple apps on the same CAE. I can't see any real use case where CAE hosts a single CA, so the current module design faces these issues:
- it can only create a new CAE with a single CA
- CA and CAE names are the same (`app_name`)
- in order to create more CAs on the same CAE, the exported CAE Id must be used as input in other instances of this module to create more CA on the same CAE

Like:
```hcl
module "cae_ca" {
  # creates CAE and CA with the same name 
}

module "ca2" {
  cae_id = module.cae_ca.cae_id
  # creates CA2
}

module "ca3" {
  cae_id = module.cae_ca.cae_id
  # creates CA3
}
```

It does not make sense to me: the CA1 is tied to CAE and cannot be deleted (ever).

A more realistic design would be:
- a module to create a CAE
- a module to create one or multiple CAs on a given CAE (probably the latter is better)

With this approach, we also solve the problem with data conditional access

Krusty93 (changes requested):

gunzip:
> a module to create a CAE
> a module to create one or multiple CAs on a given CAE (probably the latter is better)

As a general adivice we must implement the single responsibility principle for each module. Only then, we can have modules that are a composition of other modules.

Krusty93:
> > a module to create a CAE
> > a module to create one or multiple CAs on a given CAE (probably the latter is better)
> 
> As a general adivice we must implement the single responsibility principle for each module. Only then, we can have modules that are a composition of other modules.

It's correct but it depends also on the real use cases.
Theoretically, the difference between AppService/AppService Plan and CA/CAE is nothing: both needs the other part as underlying infrastructure; both ASP and CAE can run multiple apps.

However, our usage pattern is radically different: ASP should run one critical workload at time, while CAE is designed to work with multiple CAs at time. For these reasons I see the need of different module design

infra/modules/azure_container_app/container_app.tf
mamu0: Great point! I agree with this solution. By splitting the modules, we would have more configuration flexibility and could add more CAs without complicating the logic.
If everyone else agrees @pagopa/engineering-team-devex @pagopa/engineering-team-cloud-eng, I can start making the changes. Would it be okay to do everything in this PR, or would it be better to create a new one?

infra/modules/azure_container_app/networking.tf
```diff
@@ -0,0 +1,39 @@
...
+  delegation {
+    name = "Microsoft.App/environments"
+    service_delegation {
+      name    = "Microsoft.App/environments"
+      actions = ["Microsoft.Network/virtualNetworks/subnets/join/action"]
+    }
+  }
+}
+
+resource "azurerm_private_endpoint" "this" {
```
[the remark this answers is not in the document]
mamu0: Something might be missing. As it is now, it creates the PEP and links it to the environment, but the recordset should be generated by the Container App, and nothing is being done at the networking level there

mamu0:
> > As a general adivice we must implement the single responsibility principle for each module. Only then, we can have modules that are a composition of other modules.

Yes, managing separate modules would improve maintainability and reduce complexity. Since we don’t have a tool for stack creation yet, we can add an example covering the full use case (creating the CAE and multiple CAs within it), like we did for core infra.

infra/modules/azure_container_app/container_app.tf
gunzip: It's ok to do that in this PR.

infra/modules/azure_container_app/networking.tf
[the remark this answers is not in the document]
Krusty93: I agree with you. I suggest to make some tests and understand better what causes the issue (maybe a TF provider bug, or the fact PEPs are still in preview, who knows) and mitigate it somehow

infra/modules/azure_container_app/container_app.tf
mamu0: @Krusty93 Now we have two modules, one for container app and one for container app environment!

.changeset/kind-steaks-applaud.md
```diff
@@ -0,0 +1,5 @@
+---
+"azure_container_app": patch
```
christian-calabrese, line 2: **todo**: a new changeset file for azure_container_app_environment module is needed for the first release as well

infra/modules/azure_container_app/container_app.tf
```diff
@@ -0,0 +1,119 @@
...
+  workload_profile_name        = "Consumption"
+
+  identity {
+    type = "SystemAssigned"
+  }
+
+  ingress {
+    allow_insecure_connections = false
+    external_enabled           = true
+    target_port                = 8080
```
christian-calabrese, line 15: **issue**: I think that the target_port cannot be hardcoded to 8080. It should reflect the port exposed by the containerized application. For example, if I dockerize an api application that listens on port 3000, target_port should be set to 3000, hence decided by the developer.
I may be wrong, so I invoke also @Krusty93 who's more experienced on container apps

**suggestion**: Use a variable to ask for the port exposed by the application

infra/modules/azure_container_app/main.tf
```diff
@@ -0,0 +1,22 @@
...
+    azurerm = {
+      source  = "hashicorp/azurerm"
+      version = "~> 4"
+    }
+  }
+}
+
+module "naming_convention" {
+  source  = "pagopa/dx-azure-naming-convention/azurerm"
+  version = "~> 0"
```
christian-calabrese, line 12:
```suggestion
  version = "~> 0.0"
```

infra/modules/azure_container_app_environment/main.tf
```diff
@@ -0,0 +1,58 @@
...
+    azurerm = {
+      source  = "hashicorp/azurerm"
+      version = "~> 4"
+    }
+  }
+}
+
+module "naming_convention" {
+  source  = "pagopa/dx-azure-naming-convention/azurerm"
+  version = "~> 0"
```
christian-calabrese, line 12:
```suggestion
  version = "~> 0.0"
```

infra/modules/azure_container_app/variables.tf
```diff
@@ -0,0 +1,129 @@
...
+  type        = string
+  description = "Resource group to deploy resources to"
+}
+# ------------ CONTAINER ENVIRONMENT ------------ #
+
+variable "container_app_environment" {
+  type = object({
+    id = string
+    private_dns_zone = object({
+      name                = string
```
Krusty93, line 30: **suggestion**: The zone should be always the same, in the format `privatelink.<region>.azurecontainerapps.io`.
To keep the approach consistent with [other modules](https://github.com/pagopa/dx/blob/33cea146fc1849a7c8d61d29805967adf7b19281/infra/modules/azure_app_service/data.tf#L6), I suggest to set a `data` to get the DNS zone id instead of asking the same value to the user. I'd stick to only ask for the [resource group name](https://github.com/pagopa/dx/blob/33cea146fc1849a7c8d61d29805967adf7b19281/infra/modules/azure_app_service/locals.tf#L31)

infra/modules/azure_container_app/networking.tf
```diff
@@ -0,0 +1,9 @@
+resource "azurerm_private_dns_a_record" "this" {
+  name                = replace(azurerm_container_app.this.ingress[0].fqdn, ".azurecontainerapps.io", "")
```
Krusty93, line 2: **issue**: I ran a test with the Azure Portal, and the created A record  has the random value part of the FQDN only. I thought about it and this is a clever approach, as it works for all the apps inside the same environment. In fact, the FQDN format is: `<ca-name>--<revision-id>.<random-cae-name>.<region>.azurecontainerapps.io`. The DNS zone name is `privatelink.italynorth.azurecontainerapps.io` and the A record is `<random-cae-name>`, while the application URL is `https://<ca-name>.<random-cae-name>.<region>.azurecontainerapps.io`

This code should be moved into the cae module, but I suspect it is unnecessary as the record is automatically added by the `azurerm_private_endpoint` resource as for any other kind of service (it's asking for the private dns zone id after all).

Also, the variable `private_endpoint_ip` is no longer necessary

Krusty93 (changes requested):
More considerations over networking

gunzip:
what about providing the ability to creare shared volumes? https://learn.microsoft.com/en-us/azure/container-apps/storage-mounts?tabs=smb&pivots=azure-cli#azure-files

this will be useful for example for every scenario where you would have used `VOLUME` directive in Docker files.

infra/modules/azure_container_app/container_app.tf
mamu0: i think you are right, it's better with a configurable variable

mamu0:
> what about providing the ability to creare shared volumes? https://learn.microsoft.com/en-us/azure/container-apps/storage-mounts?tabs=smb&pivots=azure-cli#azure-files
> 
> this will be useful for example for every scenario where you would have used `VOLUME` directive in Docker files.

I looked into the possibility of using shared volumes in **Azure Container Apps**, and it turns out that it’s not directly supported within the `azurerm_container_app` resource. Instead, if shared storage is needed, it must be configured at the environment level using the `azurerm_container_app_environment_storage` resource, which requires an **Azure Storage Account** and a **File Share**.

Before proceeding, we should evaluate whether this setup is actually necessary, as it introduces additional dependencies. I’m not entirely sure what the specific use case for this feature would be. @Krusty93

infra/modules/azure_container_app/networking.tf
Krusty93: By creating the correct private dns zone `privatelink.italynorth.azurecontainerapps.io`, it creates the record
image

mamu0: Ah, okay, great! So the issue was just with the name of the private DNS zone. I’ll make the corrections and try again. Thanks

Krusty93 (approved):

PR #295, opened 2025-02-19
Force `node_modules` in `function_app_deploy.yaml`

lucacavallaro:
Lagacy projects that uses this workflow needs `node-modules` as package resolution algorithm. This change will force this mechanism even if the root project adopts Plug'n'Play.

### DIFF
--- a/.github/workflows/function_app_deploy.yaml
+++ b/.github/workflows/function_app_deploy.yaml
@@ -48,6 +48,9 @@ concurrency:
 
 env:
   BUNDLE_NAME: bundle
+  # we fall back to node-moules, even in case pnp is configured, in order to avoid bundling dependendencies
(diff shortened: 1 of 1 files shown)


gunzip (approved):
is there a chance that this would break some pipeline? in other words: might be a breaking change?

lucacavallaro:
> is there a chance that this would break some pipeline? in other words: might be a breaking change?

Nope, this workflow works only on `node_modules` projects (due the `"bundleDependencies": true` trick used to make the ZIP artifact) 

Note that this is a deprecated workflow, that should be used only on legacy project with the old programming model.

This change was needed because we have mixed monorepos with legacy and modern functions project. The modern ones use the `web_app_deplpy` workflow.

PR #298, opened 2025-02-20
[CES-767] Upgrade Postgres SKUs to v5

Krusty93:
### List of changes

Tier L now uses `GP_Standard_D4ds_v5` and tier M moved to `GP_Standard_D2ds_v5`

### Motivation and context

PagoPA owns reservations only for v5

### Type of changes

- [ ] Add new resources

- [ ] No

### DIFF
--- a/.changeset/khaki-waves-prove.md
+++ b/.changeset/khaki-waves-prove.md
@@ -1,5 +1,5 @@
 ---
-"azure_postgres_server": minor
+"azure_postgres_server": major
 ---
 
 L tier now uses GP_Standard_D4ds_v5 as SKU, and M tier uses GP_Standard_D2ds_v5
--- a/.changeset/khaki-waves-prove.md
+++ b/.changeset/khaki-waves-prove.md
@@ -2,4 +2,4 @@
 "azure_postgres_server": minor
 ---
 
-L tier now uses GP_Standard_D4ds_v5 as SKU
+L tier now uses GP_Standard_D4ds_v5 as SKU, and M tier uses GP_Standard_D2ds_v5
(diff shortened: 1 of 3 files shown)


gunzip (commented):
does this cause a resource delete ?

Krusty93:
> does this cause a resource delete ?

It should trigger an update in place, buy nobody's using the module yet: we have only one reference from CGN repo, but this change have been asked by them

gunzip:
As a general priciple, it's ok to maintain `minor` for update in place, but 'nobody's using the module' still triggers a major for breaking changes :-)

It looks like this is worth an Azure Policy that limits available SKU, isn't it? Maybe we can contribute to the global repo?

gunzip (approved):

pasqualedevita (approved):

pasqualedevita:
> > It looks like this is worth an Azure Policy that limits available SKU, isn't it? Maybe we can contribute to the global repo?
> 
> Probably @pasqualedevita any though?

We can add only a policy in audit mode because sku version changes but old resources can have active reservation.

So the generic policy should be:
1. for new resources must use the latest sku (v5)
2. existing resources should migrate to v5 if not already reserved.

GP_Standard_D2s_v3 is very old (maybe deprecated in next 1 or 2 year) so probably we can deny this sku

pasqualedevita:
changing sku cause a downtime because db will be restarted, maybe we can insert this information into changelog

pasqualedevita:
Postgresql SKU policy is already active for all PROD subscriptions

https://github.com/pagopa/eng-azure-governance/blob/main/src/03_policy_set/01_postgresql_prod.tf

christian-calabrese (approved):

PR #311, opened 2025-02-21
[CES-774] Update bootstrap module to accept a list of resource group ids

Krusty93:
### List of changes

Add an optional variable to accept a list of resource group to apply IAM roles to

### Motivation and context

Teams may have multiple resource groups under their ownership, and applying IAM roles over and over again on each of them is a slow task

### Type of changes

- [ ] Add new resources

- [X] No

### DIFF
--- a/.changeset/smart-chicken-turn.md
+++ b/.changeset/smart-chicken-turn.md
@@ -1,0 +1,5 @@
+---
+"azure_github_environment_bootstrap": minor
+---
+
+Add an optional variable to accept a list of resource group to apply IAM roles to
--- a/infra/modules/azure_github_environment_bootstrap/README.md
+++ b/infra/modules/azure_github_environment_bootstrap/README.md
@@ -6,8 +6,8 @@
 
 | Name | Version |
 |------|---------|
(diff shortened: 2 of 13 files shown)


gunzip (commented):
**question**: what's the difference between primary and secondary resource groups?

gunzip:
Sorry, let me clarify further. The difference in usage is clear; what I meant to ask is whether the module handles them differently. If it doesn’t, we should use a single variable for all resource groups.

gunzip:
> This way, it becomes clearer that the main resource group is always used (meaning for_each initially iterates over a single element), and any additional resource groups are simply appended to the same list

Yep! Good idea to distinguish between one required rg and other optional rgs.

Krusty93:
> 
> Alternatively, you could use a single configuration for the different resources by leveraging `for_each` and creating a local variable like this:
> 
> ```
> locals {
>   resource_group_ids = concat([azurerm_resource_group.main.id], var.additional_resource_group_ids)
> }
> ```
> 
> This way, it becomes clearer that the main resource group is always used (meaning `for_each` initially iterates over a single element), and any additional resource groups are simply appended to the same list

Code is cleaner, but got this error
image


I'm not sure there's a way to workaround this

mamu0 (approved):

PR #315, opened 2025-02-25
Allow contextual creation of  role assignments and related resources

christian-calabrese:
Due to the presence of data sources, it is not possible to contextually create the target resource (storage account, redis cache, apim, ...) and the related role assignments in the same terraform plan or apply.

Resolves #CES-781


infra/modules/azure_role_assignments/modules/key_vault/variables.tf
```diff
@@ -7,6 +7,8 @@ variable "key_vault" {
   description = "A list of key vault role assignments"
   type = list(object({
     name                = string
+    id                  = optional(string, null)
+    supports_rbac       = optional(bool, null)
```
gunzip, line 11:
```suggestion
    has_rbac_support       = optional(bool, null)
```

as a geneaìreal rule starts boolean var names with `has` or `is`

infra/modules/azure_role_assignments/modules/apim/variables.tf
```diff
@@ -9,6 +9,7 @@ variable "apim" {
     name                = string
     resource_group_name = string
     role                = string
+    id                  = optional(string, null)
```
Krusty93, line 12: What is?

Krusty93:
Does it make sense to keep name/resource group pairs in variables?

christian-calabrese:
> Does it make sense to keep name/resource group pairs in variables?

It's pretty much the same idea as described [here](https://github.com/pagopa/dx/pull/304#discussion_r1965166247). 
Do you think we can get rid of the possibility to just specify the name of the service when it's already created?

Krusty93:
> > Does it make sense to keep name/resource group pairs in variables?
> 
> It's pretty much the same idea as described [here](https://github.com/pagopa/dx/pull/304#discussion_r1965166247). Do you think we can get rid of the possibility to just specify the name of the service when it's already created?
> 
> PS: Actually, I should check if they are still needed when specifying the resource id. Maybe we can make them optional and check that at least one is populated

I think we can remove name/rg pairs in both contexts (new/existing) as the id is always needed to the `scope` property, while the name/rg pair is useful only for `data`. I suggest to have a look at it before doing changes by the way

christian-calabrese:
> > > Does it make sense to keep name/resource group pairs in variables?
> > 
> > 
> > It's pretty much the same idea as described [here](https://github.com/pagopa/dx/pull/304#discussion_r1965166247). Do you think we can get rid of the possibility to just specify the name of the service when it's already created?
> > PS: Actually, I should check if they are still needed when specifying the resource id. Maybe we can make them optional and check that at least one is populated
> 
> I think we can remove name/rg pairs in both contexts (new/existing) as the id is always needed to the `scope` property, while the name/rg pair is useful only for `data`. I suggest to have a look at it before doing changes by the way

@Krusty93 currently, the code allows specifying either the target resource's name + resource group or its id.
We can surely remove it, but it needs the azurerm provider function `parse_resource_id`, hence azurerm 4.x.y.

Since it's a breaking change, I think we can do it anyways. 

Moreover, I saw that from version 3.114 it is possible to opt in for the provider functions feature ([source](https://registry.terraform.io/providers/hashicorp/azurerm/latest/docs/functions/parse_resource_id))

If you and the team agrees, I proceed removing the name and resource group name variables for simplicity.

.changeset/beige-buses-refuse.md
```diff
@@ -0,0 +1,5 @@
+---
+"azure_role_assignments": major
+---
+
```
gunzip, line 4: can we elaborate about why the change is breaking and add a migration path for users?

gunzip (approved):
lgtm, let's see what others think @pagopa/engineering-team-devex

.changeset/beige-buses-refuse.md
```diff
@@ -0,0 +1,44 @@
...
+Role assignments can now be created contextually with identities and target resources
+
+**Migration guide**:
+Remove all name and resource_group_name parameters from the variables and use the id gotten from a resource, module or data source as follows:
```
kin0992, line 8: **nitpick**: I'd format the parameters as code, so it is even more expressive in the changelog

I'm going to make a separate suggestion in case you want to add it directly

kin0992, line 8:
```suggestion
Remove all `name` and `resource_group_name` parameters from the variables and use the id gotten from a resource, module or data source as follows:
```

infra/modules/azure_role_assignments/modules/apim/locals.tf
```diff
@@ -1,9 +1,14 @@
 locals {
-  apims = distinct([for assignment in var.apim : { name = assignment.name, resource_group_name = assignment.resource_group_name }])
-
+  norm_apims = [for apim in var.apim : {
```
kin0992, line 2: **question**: what does `norm` stand for?

infra/modules/azure_role_assignments/modules/apim/providers.tf
```diff
@@ -0,0 +1,8 @@
+terraform {
+  required_providers {
+    azurerm = {
+      source  = "hashicorp/azurerm"
+      version = ">= 3.114, < 5.0"
```
Krusty93, line 5: Is `3.114` ok with the `provider::azurerm::parse_resource_id` function?

infra/modules/azure_role_assignments/modules/apim/locals.tf
```diff
@@ -1,9 +1,14 @@
...
+    name                = provider::azurerm::parse_resource_id(apim.id)["resource_name"]
```
Krusty93, line 3: Why do you still need `name` and `resource_group_name`?

infra/modules/azure_role_assignments/providers.tf
```diff
@@ -2,7 +2,7 @@ terraform {
   required_providers {
     azurerm = {
       source  = "hashicorp/azurerm"
-      version = ">= 3.110, < 5.0"
+      version = ">= 3.114, < 5.0"
```
Krusty93, line 5: **question**: Isn't this constraint enough for all the sub modules?

infra/modules/azure_role_assignments/tests/role.tftest.hcl
```diff
@@ -1,6 +1,7 @@
 provider "azurerm" {
   features {
   }
+  subscription_id = "d7de83e0-0571-40ad-b63a-64c942385eae"
```
Krusty93, line 4: **suggestion**: It is already set in CI, I don't think we need it here as well

infra/modules/azure_role_assignments/modules/apim/locals.tf
christian-calabrese: It stands for `normalized`, there's not precise explanation for the choice. It is a local used to insert and edit additional (required) information to the input variable

infra/modules/azure_role_assignments/modules/apim/providers.tf
christian-calabrese: I added [these lines](https://github.com/pagopa/dx/pull/315/files#diff-245ae644dba1367ed8d97631dc606a3a2f02dc5fb828e74493dcd16dd7271180R8-R12) in the migration guide to explain how to make it work for >= 3.114 azurerm versions

infra/modules/azure_role_assignments/providers.tf
christian-calabrese: Submodules can be instantiated on their own, so the providers definition shall be defined

.changeset/beige-buses-refuse.md
```diff
@@ -0,0 +1,50 @@
+---
+"azure_role_assignments": major
+---
+
+Role assignments can now be created contextually with identities and target resources
+
+**Migration guide**:
+When using the azurerm provider version >= 3.114.0, the beta opt-in feature must be enabled by setting the following environment variable:
```
gunzip, line 8: **question**: what's the "the beta opt-in feature" ?

christian-calabrese: I tried to explain it better in a next commit. The module makes use of features available from azurerm 4.0 version (provider defined functions). The beta opt-in feature allows to use them from 3.114.0 version as well.

It's a way to not break projects that don't support the 4.0 version of the provider yet.

infra/modules/azure_role_assignments/modules/apim/locals.tf
Krusty93: If they are only internal we can keep them I think. But I wouldn't use them for new submodules

Krusty93 (approved):

PR #316, opened 2025-02-26
Add logic to delete Azure resources upon TTL expiration

mamu0:
### List of changes

Now only for `azure_storage_account` module, as a proposed change, the `ModifiedOn` tag has been added, which is created during apply and set the date of last apply on terraform.
A Azure policy that add the same `ModifiedOn` tag on new resources only if environment is `Dev` and `ModifiedOn` is not `Terraform`.
A GitHub Workflow that search all Dev resources with `ModifiedOn` tag and delete them if `ModifiedOn+TTL` is today.

### Motivation and context

As previously proposed, a tag will be added to trigger the deletion of `DEV` resources.
With this tag, added via terraform and policy will be a trigger for the GitHub Workflow that check if the resources have reached the expiry date.

Possible Next steps:
- Slack notification 1 day before deletion
- ...

### Type of changes

- [ ] Add new resources

- [x] No

### Other information

Resolves: CES-525


gunzip:
maybe I'm missing something, but imho this will result in an odd setup where you have a tag 'expiration day' on every resource which defaults to some value which is not used where the related azure policy is not running. 

assume dev don't set up this intentionally. they'll see in the terraform state that their resources will 'expire' which is scary.

mamu0:
> maybe I'm missing something, but imho this will result in an odd setup where you have a tag 'expiration day' on every resource which defaults to some value which is not used where the related azure policy is not running.
> 
> assume dev don't set up this intentionally. they'll see in the terraform state that their resources will 'expire' which is scary. can't we handle this in some other way? (without an explicit expire tag within terraform modules)

I made a change to restrict tag creation to the `DEV` environment to avoid unnecessary alerts in other environments.

Since deletion will be triggered by an Action, an alternative approach is to add only a `CreationDate` tag to each resource and calculate the `ExpirationDate` within the Action.

Additionally, we could enforce a **policy** that prevents resource creation in `DEV` if the `CreationDate` tag is missing.

Another option is a policy that denies deletion of `DEV` resources until they reach their `ExpirationDate`. However, this might be less convenient for developers and should be carefully considered.

gunzip:
> I made a change to restrict tag creation to the DEV environment to avoid unnecessary alerts in other environments.
 
👍 

> Since deletion will be triggered by an Action, an alternative approach is to add only a CreationDate tag to each resource and calculate the ExpirationDate within the Action.

👍 

> Additionally, we could enforce a policy that prevents resource creation in DEV if the CreationDate tag is missing.

This will slow down development.

> Another option is a policy that denies deletion of DEV resources until they reach their ExpirationDate. However, this might be less convenient for developers and should be carefully considered.

👎

mamu0:
> This will slow down development.
> 
Actually, it shouldn’t be an issue since all new resources will have the tag added by Terraform. However, it would be useful to identify existing resources in `DEV` that don’t have the tag, as they won’t be triggered by the action.

But it’s not strictly necessary, nor something that needs to be done immediately or alongside other tasks, but it can be considered if needed.

gunzip:
> Actually, it shouldn’t be an issue since all new resources will have the tag added by Terraform

Most of dev resources are created manually, and we must expect that this will persist since there's no need to create development resources using IaC.

gunzip:
Anyway, if the policy that prevents the creation of new untagged resources works with a clear error even using Azure portal, we can "enforce a policy that prevents resource creation in DEV if the CreationDate tag is missing"

mamu0:
> Anyway, if the policy that prevents the creation of new untagged resources works with a clear error even using Azure portal, we can "enforce a policy that prevents resource creation in DEV if the CreationDate tag is missing"

Yes of course, if we decided to use it on azure portal we would see an error like this:
image
So it would be enough to choose a speaking title for the policy

.github/workflows/_cleanup_resources.yaml
```diff
@@ -0,0 +1,102 @@
+name: Cleanup Resources
```
gunzip, line 1: is it possible to add a description to this workflow?

.github/workflows/_cleanup_resources.yaml
```diff
@@ -0,0 +1,102 @@
...
+
+on:
+  workflow_call:
+    inputs:
+      environment:
+        description: Environment where search resources.
```
gunzip, line 7: **suggestion**: be more specific in descriptions

.github/workflows/_cleanup_resources.yaml
```diff
@@ -0,0 +1,102 @@
...
+        type: string
+        required: false
+        default: "dev"
+      ttl:
+        description: Time to live in days for resources.
```
gunzip, line 12: ditto

.github/workflows/_cleanup_resources.yaml
```diff
@@ -0,0 +1,102 @@
...
+        id: resources
+        run: |
+          if [ "${{ inputs.environment }}" == "dev" ]; then
+            environment="Dev"
+          elif [ "${{ inputs.environment }}" == "uat" ]; then
+            environment="Uat"
+          elif [ "${{ inputs.environment }}" == "prod" ]; then
+            environment="Prod"
+          else
+            echo "Invalid environment. Please provide valid environment."
```
gunzip, line 55: **issue**: list valid environment in message

infra/modules/azure_storage_account/locals.tf
```diff
@@ -52,4 +52,8 @@ locals {
       principal_id        = try(coalesce(var.customer_managed_key.user_assigned_identity_id, azurerm_storage_account.this.identity[0].principal_id), "")
     } : {}
   }
+
+  tags = var.environment.env_short == "d" ? merge(var.tags, {
```
gunzip, line 56: **question**: why do we need this is the tag is applied by the policy below?

gunzip (changes requested):

infra/modules/azure_storage_account/locals.tf
mamu0: As discussed in the relevant RFC, if we don’t add it in Terraform, there will always be a drift since it can be modified via the portal. Additionally, the policy only applies tags during creation, meaning they won’t be recreated or updated if deleted or modified. With Terraform, the tag is enforced and updated on every apply.
Therefore, collaboration between policy and Terraform is needed to cover both cases: resources created via the **portal** and those created with one of the DX modules in **Terraform**.

mamu0:
@pagopa/engineering-team-cloud-eng @pagopa/engineering-team-devex, does the logic I’ve proposed in this **PR** work for you?
Before proceeding, I’d like your input on a couple of points:
- Is it fine to add that _tag_ to every Terraform module in **dx**, or are there any we should exclude from this logic?
- Do you think **Slack notifications** about expiring resources one day in advance are a good way to alert the teams? Now they look like this:
image

.github/workflows/_cleanup_resources.yaml
```diff
@@ -0,0 +1,166 @@
...
+              echo ${row} | base64 --decode | jq -r ${1}
+            }
+
+            modifiedOn=$(_jq '.modifiedOn')
+            resourceId=$(_jq '.id')
+
+            # Expiry Date (ModifiedOn + TTL)
+            expiryDate=$(date -d "$modifiedOn + $ttl days" +%Y-%m-%d)
+
+            if [[ "$expiryDate" == "$today" ]]; then
```
christian-calabrese, line 96: **suggestion**: Shall we check if expiryDate is equals or **greater** than today? This would help in cases where the workflow fails for some reason leaving resources undeleted
```suggestion
            if [[ "$expiryDate" -ge "$today" ]]; then
```

infra/modules/azure_storage_account/locals.tf
christian-calabrese, line 56: **question**: I think the ModifiedOn tag may be useful in the other environments (uat, prod) as well. Of course, not the deletion process, but just an additional information that can be visually compared with the modified date we can get from the azure portal.

This way, it is possible to understand whether the resource has been manually edited

.github/workflows/_cleanup_resources.yaml
mamu0: Unfortunately, this is a comparison between two strings, so `-ge` won’t work in this case

infra/modules/azure_storage_account/locals.tf
mamu0: I agree, it could be useful across all environments. Additionally, there’s no risk of deleting resources in **PROD** or **UAT** because both the **action** and the **policy** have the `environment` as a configurable parameter.

.github/workflows/_cleanup_resources.yaml
mamu0: That’s a great observation! I’ll incorporate it, but since it’s a string comparison, `<` and `==` should be used; otherwise, it will result in an error.

mamu0:
**blocked**: The main issues are:

- No reliable way to prevent drift in the `plan`, as the date would change with every `apply`, the only solution found was to ignore changes for the entire `tag` block.
- No viable solution to correctly add `tags` to sub-resources created by Terraform. Even with a _policy_ in place, `tags` would not be updated and could result in premature deletions.

PR #323, opened 2025-02-26
Set up custom warm-up in webapps modules

lucacavallaro:
Set `var.health_check_path` as custom warm-up path to constraint the swap operation only when the source slots is healthy.


gunzip:
you have to format hcl

lucacavallaro:
@Krusty93 in you opinion is ok to add these variables? there are already used by wallet and FIMS

Krusty93 (approved):

PR #328, opened 2025-02-28
Implementation of Azure FrontDoor CDN module

christian-calabrese:
Adds a new Terraform module for Azure Front Door CDN that provides content delivery capabilities with support for multiple origins, custom domains, and secure defaults (e.g. HTTPS-only). 

Resolves #CES-520


infra/modules/azure_cdn/cdn.tf
```diff
@@ -0,0 +1,81 @@
+resource "azurerm_cdn_frontdoor_profile" "this" {
+  name                = "${module.naming_convention.prefix}-cdnp-${module.naming_convention.suffix}"
```
Krusty93, line 2:
```suggestion
  name                = "${module.naming_convention.prefix}-afd-${module.naming_convention.suffix}"
```

infra/modules/azure_cdn/cdn.tf
```diff
@@ -0,0 +1,81 @@
...
+  resource_group_name = var.resource_group_name
+  sku_name            = "Standard_AzureFrontDoor"
+
+  tags = var.tags
+}
+
+resource "azurerm_cdn_frontdoor_endpoint" "this" {
+  name                     = "${module.naming_convention.prefix}-cdne-${module.naming_convention.suffix}"
```
Krusty93, line 10:
```suggestion
  name                     = "${module.naming_convention.prefix}-fde-${module.naming_convention.suffix}"
```

infra/modules/azure_cdn/cdn.tf
```diff
@@ -0,0 +1,81 @@
...
+  cdn_frontdoor_profile_id = azurerm_cdn_frontdoor_profile.this.id
+
+  tags = var.tags
+}
+
+resource "azurerm_cdn_frontdoor_origin_group" "this" {
+  name                     = "${module.naming_convention.prefix}-cdnog-${module.naming_convention.suffix}"
```
Krusty93, line 17:
```suggestion
  name                     = "${module.naming_convention.prefix}-afdog-${module.naming_convention.suffix}"
```

infra/modules/azure_cdn/cdn.tf
```diff
@@ -0,0 +1,81 @@
...
+  cdn_frontdoor_profile_id = azurerm_cdn_frontdoor_profile.this.id
+
+  health_probe {
```
Krusty93, line 20: **question**: what's the difference between `health_probe` and `load_balancing`?
[here](https://github.com/pagopa/io-infra/blob/0d32a5a261d7b5ac6f4d7492123f8270e6c9aec4/src/domains/ioweb-common/06_cdn_itn.tf#L79) they use the latter

infra/modules/azure_cdn/cdn.tf
```diff
@@ -0,0 +1,81 @@
...
+    request_type        = "HEAD"
+  }
+
+  load_balancing {}
+}
+
+resource "azurerm_cdn_frontdoor_origin" "this" {
+  for_each = var.origins
+
+  name                           = "${module.naming_convention.prefix}-${each.key}-cdnp-${module.naming_convention.suffix}"
```
Krusty93, line 33:
```suggestion
  name                           = "${module.naming_convention.prefix}-${each.key}-fdo-${module.naming_convention.suffix}"
```

infra/modules/azure_cdn/custom_domain.tf
```diff
@@ -0,0 +1,29 @@
...
+
+resource "azurerm_dns_a_record" "this" {
+  for_each            = { for custom_domain in var.custom_domains : custom_domain.host_name => custom_domain if custom_domain.dns.zone_name != null && custom_domain.dns.zone_resource_group_name != null }
+  name                = each.value.host_name == each.value.dns.zone_name ? "@" : trimsuffix(each.value.host_name, ".${each.value.dns.zone_name}")
+  zone_name           = each.value.dns.zone_name
+  resource_group_name = each.value.dns.zone_resource_group_name
+  ttl                 = 3600
+  target_resource_id  = azurerm_cdn_frontdoor_endpoint.this.id
+  tags                = var.tags
+}
```
Krusty93, line 29: **suggestion**: [here](https://github.com/pagopa/io-infra/blob/0d32a5a261d7b5ac6f4d7492123f8270e6c9aec4/src/domains/ioweb-common/06_cdn_itn.tf#L213) they set TXT as well for domain verification. I think we need it as well

infra/modules/azure_cdn/examples/basic/versions.tf
```diff
@@ -0,0 +1,34 @@
...
+  }
+}
+
+provider "azurerm" {
+  features {}
+}
+
+module "naming_convention" {
+  source  = "pagopa/dx-azure-naming-convention/azurerm"
+  version = "~> 0.0"
```
gunzip, line 24: **question**: why constraint to a minor?

infra/modules/azure_cdn/custom_domain.tf
```diff
@@ -0,0 +1,44 @@
...
+
+resource "azurerm_dns_txt_record" "validation" {
```
gunzip, line 31: **suggestion**: can we add a comment that explain this part in brief?

infra/modules/azure_cdn/custom_domain.tf
christian-calabrese: Done!

infra/modules/azure_cdn/core
Krusty93, line 1: ?

infra/modules/azure_cdn/outputs.tf
```diff
@@ -0,0 +1,14 @@
+output "id" {
```
Krusty93, line 1: Always add `name` and `resource_group_name` as they are the most used module outputs

infra/modules/azure_cdn/examples/basic/versions.tf
gunzip: It looks like TF does not follow semantic version rules when using '~>', as it allows major upgrades: version = ">=0, <1" 

So it's ok to use "~> 0.0"

infra/modules/azure_cdn/cdn.tf
christian-calabrese: Both of them are important, but have different objective:
- health_probe: define on which path, protocol and how often check for origins health
- load_balancing: define how to determine the origin to choose in a group according to latency and health. This is mostly important when origins are located in different regions around the globe ([doc](https://learn.microsoft.com/en-us/azure/frontdoor/routing-methods#lowest-latencies-based-traffic-routing))

They both have default values that i think are correct for our use cases

Krusty93 (approved):

PR #329, opened 2025-03-03
Create APIM logger when AI is enabled

### LINKED ISSUE #273: Application Insights on APIM module with tier `s` is not working
When connecting Application Insights to the APIM (APIM's tier equals to `s`), the Terraform plan fails because it didn't fine any logger resource.
The logger resource is going to be created when AI is enabled and `tier` is not `s`.

You can find more details about the error in [this PR](https://github.com/pagopa/dx-playground/pull/121#issuecomment-2639151596).

kin0992:
When tier is set to `s`, but AI is enabled, the plan fails because it assumes the APIM logger exists.
With this change, when AI is enabled, then the module creates an APIM logger as well

Closes #273 #CES-799

### DIFF
--- a/infra/modules/azure_api_management/monitor.tf
+++ b/infra/modules/azure_api_management/monitor.tf
@@ -1,5 +1,5 @@
 resource "azurerm_api_management_logger" "this" {
-  count = var.application_insights.enabled && var.tier != "s" ? 1 : 0
+  count = var.application_insights.enabled ? 1 : 0
 
   name                = "${local.apim.name}-logger"
   api_management_name = azurerm_api_management.this.name
@@ -11,7 +11,6 @@ resource "azurerm_api_management_logger" "this" {
       connection_string = var.application_insights.connection_string
     }
   }
-
 }
 
 resource "azurerm_monitor_metric_alert" "this" {
--- a/infra/modules/azure_api_management/tests/apim.tftest.hcl
+++ b/infra/modules/azure_api_management/tests/apim.tftest.hcl
@@ -57,13 +57,17 @@ run "apim_is_correct_plan" {
 
     subnet_id                     = run.setup_tests.subnet_id
     virtual_network_type_internal = true
-
   }
 
   # Checks some assertions
   assert {
     condition     = azurerm_api_management.this.sku_name == "Premium_1"
     error_message = "The APIM SKU is incorrect, have to be Premium_1"
+  }
+
+  assert {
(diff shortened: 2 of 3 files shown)


infra/modules/azure_api_management/monitor.tf
```diff
@@ -1,5 +1,5 @@
 resource "azurerm_api_management_logger" "this" {
-  count = var.application_insights.enabled && var.tier != "s" ? 1 : 0
+  count = var.application_insights.enabled ? 1 : 0
```
kin0992, line 2: **note**: in order to use Application Insights on APIM with tier `s` (like the one used by the playground), we need to create the `azurerm_api_management_logger` resource

mamu0:
I’m not sure I fully understand the goal of this PR. Is the intention to enable the creation of an AI even when the tier is `s`, or to prevent the error that occurs when AI creation is enabled for tier `s`?
If it’s the latter, the issue might be with the `logger_id` output, as it lacks a condition on the tier
```
output "logger_id" {
  value = var.application_insights.enabled && var.tier != "s" ? azurerm_api_management_logger.this[0].id : null
}
```

kin0992:
> I’m not sure I fully understand the goal of this PR. Is the intention to enable the creation of an AI even when the tier is `s`, or to prevent the error that occurs when AI creation is enabled for tier `s`? If it’s the latter, the issue might be with the `logger_id` output, as it lacks a condition on the tier
> 
> ```
> output "logger_id" {
>   value = var.application_insights.enabled && var.tier != "s" ? azurerm_api_management_logger.this[0].id : null
> }
> ```

@mamu0 my goal is to link AI to the APIM

I think we do not have reason to avoid the creation of a logger if the tier is `s`; if a requirement is not to allow the creation of that, then we should improve the module to prevent that from happening.

infra/modules/azure_api_management/tests/apim.tftest.hcl
```diff
@@ -57,14 +57,18 @@ run "apim_is_correct_plan" {
...
 
   # Checks some assertions
   assert {
     condition     = azurerm_api_management.this.sku_name == "Premium_1"
     error_message = "The APIM SKU is incorrect, have to be Premium_1"
   }
+
+  assert {
+    condition     = length(azurerm_api_management_logger.this) == 0
+    error_message = "The APIM logegr should not exist"
```
mamu0, line 70:
```suggestion
    error_message = "The APIM logger should not exist"
```

mamu0 (approved):

PR #334, opened 2025-03-06
Add docs for Private Endpoints creation across Azure subscriptions

Krusty93:
Add documentation to guide developers in Private Endpoint setup among two resources of different Azure subscriptions

Closes #CES-779


website/docs/infrastructure/peps-cross-subscription.md
```diff
@@ -0,0 +1,90 @@
+---
+sidebar_label: Connecting Resources Across Azure Subscription via Private Endpoints
+---
+
+# Connecting resources across Azure Subscription via Private Endpoints
+
+- [Connecting resources across Azure Subscription via Private Endpoints](#connecting-resources-across-azure-subscription-via-private-endpoints)
```
gunzip, line 7: **hint**: TOC aren't needed since you have the sidebar menu on the right

website/docs/infrastructure/peps-cross-subscription.md
```diff
@@ -0,0 +1,90 @@
...
+  - [Overview](#overview)
+  - [Step-by-Step Implementation](#step-by-step-implementation)
+    - [Define the Private Endpoint in the caller's Terraform codebase (Subscription B)](#define-the-private-endpoint-in-the-callers-terraform-codebase-subscription-b)
+    - [Request Approval from the Target Team](#request-approval-from-the-target-team)
+      - [Private Endpoint Approval mechanism](#private-endpoint-approval-mechanism)
+        - [Via Azure Portal](#via-azure-portal)
+        - [Via Azure CLI](#via-azure-cli)
+
+## Overview
```
gunzip, line 16: **issue**: you need to apply formatting using prettier, this should limit lines to 80 chars:
https://github.com/pagopa/dx/blob/main/website/.prettierrc

gunzip: @kin0992 is going to add a pre-commit task to handle this automatically

website/docs/infrastructure/peps-cross-subscription.md
Krusty93: Sure? The sidebar has the website content, TOC is about in-page content

website/docs/infrastructure/peps-cross-subscription.md
Krusty93: Including hidden links?

website/docs/infrastructure/peps-cross-subscription.md
Krusty93: Never seen that lol

website/docs/infrastructure/peps-cross-subscription.md
Krusty93: Done

website/docs/infrastructure/peps-cross-subscription.md
```diff
@@ -0,0 +1,97 @@
...
+    is_manual_connection           = true
+  }
+}
+```
+
+:::warning
+
+For resources in the same subscription, it is recommended to set the property
+`is_manual_connection` to `false` in order to skip the approval step as the
+owner of the two resources is often the same. However, in this case, the
```
gunzip, line 55: **question**: I cannot understand this statement. Should we set it to true or false'? (Which 'case'?) :-)

website/docs/infrastructure/peps-cross-subscription.md
```diff
@@ -0,0 +1,97 @@
...
+property must be set to `true`; otherwise, the approval mechanism will not work.
+
+:::
+
+:::info
+
+In this example, the `private_connection_resource_id` value is a hardcoded
+string. If you prefer to have a reference via Terraform `data` block, you and
```
gunzip, line 63: issue: 'you' should refer instead to some account that does something. ie.

```suggestion
string. If you prefer to have a reference via Terraform `data` block, the account that ... <add the needed action here, for example. that needs to run terraform apply...> and...
```

website/docs/infrastructure/peps-cross-subscription.md
```diff
@@ -0,0 +1,97 @@
...
+the Managed Identities federated with your GitHub workflows need to have the
+`Reader` role on the target resource (subscription A).
+
+:::
+
+### Request Approval from the Target Team
+
+Once the Terraform configuration is applied, the target team (Subscription A)
+will receive a connection request. The connection between the two resources will
```
gunzip, line 72: issue: "receive a connection request" on which channel? (how exactly?)

website/docs/infrastructure/peps-cross-subscription.md
```diff
@@ -0,0 +1,97 @@
...
+not work until the request is not approved.
+
+#### Private Endpoint Approval mechanism
+
+##### Via Azure Portal
```
gunzip, line 77: issue: I'd add that these are alternatives

gunzip (changes requested):

website/docs/infrastructure/peps-cross-subscription.md
Krusty93: As the previous example shows, it must be set to `true`. However, I want to highlight this property, because in general we set `false`, and by doing copy and paste is easy to lose track of its value

website/docs/infrastructure/peps-cross-subscription.md
Krusty93: It is shown below

PR #335, opened 2025-03-06
Add `pre-commit` hook to lint website pages

kin0992:
Make sure the `prettier` configuration is applied to the website's pages

Closes CES-823

### DIFF
--- a/.changeset/red-countries-retire.md
+++ b/.changeset/red-countries-retire.md
@@ -1,0 +1,5 @@
+---
+"docs": patch
+---
+
+Add `pre-commit` hook to lint website pages


gunzip (approved):

.pre-commit-config.yaml
```diff
@@ -10,6 +10,11 @@ repos:
         exclude: ^.*\/modules\/.*
         pass_filenames: true
         require_serial: true
+      - id: lint-website-pages
+        name: Lint Website Pages (Markdown)
+        entry: yarn dlx prettier --write
```
lucacavallaro, line 15: why are you using `dlx prettier` instead of a defined-script such as `format`?

lucacavallaro: Ok! I think in the future we should update these workflow to install dependencies, as we should treat `infra` as a normal monorepo workspace (`yarn format` will execute `prettier` on js workspaces and `terraform fmt` on tf workspaces)

gunzip: **question**: another option could be to have `yarn dlx prettier --write` as npm task instead of using the installed `prettier` package?

lucacavallaro: the installed `prettier` package is versioned and used by third-party integration like vsc to format the code or renovate to check for CVEs

lucacavallaro: @kin0992 note if we want to use `dlx` with `prettier` we should pin the version (`yarn dlx prettier@EXACT-VERSION --write`) since the rules evolves and are not retrocompatible

lucacavallaro:
Why we want to format the `website` package and not other packages like `@pagopa/eslint-config` ? I would exec a monorepo-wide task to make sure that all workspaces are formatted

gunzip:
> Why do we want to format the website package specifically, and not other packages like @pagopa/eslint-config? I would suggest running a monorepo-wide task to ensure that all workspaces are formatted

I suspect that Terraform automatically formats its generated help documentation, which may result in misaligned content.

gunzip:
Why draft?

kin0992:
> Why draft?

because this is not the proper way to accomplish the goal. perhaps we can close this PR. what do you think @gunzip?

gunzip:
agree

kin0992:
not the way to reach the goal

PR #336, opened 2025-03-06
Use the new pagopa-dx GitHub organization to host dx tf modules

christian-calabrese:
Resolves #CES-804

### DIFF
--- a/infra/scripts/add-module.sh
+++ b/infra/scripts/add-module.sh
@@ -8,12 +8,12 @@ NC="\033[0m" # No color
 
 # Function to display usage instructions
 usage() {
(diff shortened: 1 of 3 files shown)


Krusty93 (approved):

infra/scripts/add-module.sh
```diff
@@ -40,8 +39,7 @@ if [ -z "$DESCRIPTION" ]; then
   usage
 fi
 
-DX_PREFIX="dx"
-SUBREPO_NAME="terraform-$PROVIDER-${DX_PREFIX}-${MODULE_NAME}"
+SUBREPO_NAME="terraform-$PROVIDER-${MODULE_NAME}"
```
kin0992, line 42: **question**: why a variable is wrapped inside the `{}`, while another one is not?

Could be possible to use the same syntax?

mamu0 (approved):

gunzip (approved):

PR #344, opened 2025-03-11
Add resource_id attribute when creating the APIM logger

kin0992:
This helps the navigation on the Azure Portal, because it links the resource

Closes #CES-839


kin0992:
This PR is backward compatible: it works if the `id` property is not set.

Before this PR, when the logger is created, it is not _linked_ to the AI instance:
Screenshot 2025-03-11 at 11 45 46

Adding the `resoucre_id` property, if you click on the logger's name, you can go to the AI page related to the logger:
Screenshot 2025-03-11 at 11 48 41

gunzip (approved):
good catch!

PR #345, opened 2025-03-11
Create azurerm_api_management_diagnostic resource

kin0992:
This enables AI to collect APIM logs

### DIFF
--- a/infra/modules/azure_api_management/monitor.tf
+++ b/infra/modules/azure_api_management/monitor.tf
@@ -14,6 +14,22 @@ resource "azurerm_api_management_logger" "this" {
   }
 }
 
+resource "azurerm_api_management_diagnostic" "this" {
+  count = var.application_insights.enabled ? 1 : 0
+
+  identifier               = "applicationinsights"
+  api_management_name      = azurerm_api_management.this.name
+  resource_group_name      = azurerm_api_management.this.resource_group_name
+  api_management_logger_id = azurerm_api_management_logger.this.id
+
+  always_log_errors         = true
+  http_correlation_protocol = "W3C"
+  sampling_percentage       = "100.0"
+  verbosity                 = "information"
+
+  depends_on = [azurerm_api_management_logger.this]
+}
+
 resource "azurerm_monitor_metric_alert" "this" {
   for_each = var.tier != "s" ? var.metric_alerts : {}
 
--- a/infra/modules/azure_api_management/README.md
+++ b/infra/modules/azure_api_management/README.md
@@ -56,6 +56,7 @@ module "apim" {
 |------|------|
 | [azurerm_api_management.this](https://registry.terraform.io/providers/hashicorp/azurerm/latest/docs/resources/api_management) | resource |
 | [azurerm_api_management_certificate.this](https://registry.terraform.io/providers/hashicorp/azurerm/latest/docs/resources/api_management_certificate) | resource |
+| [azurerm_api_management_diagnostic.this](https://registry.terraform.io/providers/hashicorp/azurerm/latest/docs/resources/api_management_diagnostic) | resource |
 | [azurerm_api_management_logger.this](https://registry.terraform.io/providers/hashicorp/azurerm/latest/docs/resources/api_management_logger) | resource |
 | [azurerm_api_management_policy.this](https://registry.terraform.io/providers/hashicorp/azurerm/latest/docs/resources/api_management_policy) | resource |
 | [azurerm_management_lock.this](https://registry.terraform.io/providers/hashicorp/azurerm/latest/docs/resources/management_lock) | resource |


infra/modules/azure_api_management/monitor.tf
```diff
@@ -14,6 +14,20 @@ resource "azurerm_api_management_logger" "this" {
   }
 }
 
+resource "azurerm_api_management_diagnostic" "this" {
+  count = var.application_insights.enabled ? 1 : 0
+
+  identifier               = "applicationinsights"
```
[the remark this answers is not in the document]
kin0992: https://github.com/hashicorp/terraform-provider-azurerm/issues/29050

infra/modules/azure_api_management/monitor.tf
```diff
@@ -14,6 +14,20 @@ resource "azurerm_api_management_logger" "this" {
...
+  api_management_name      = azurerm_api_management.this.name
+  resource_group_name      = azurerm_api_management.this.resource_group_name
+  api_management_logger_id = azurerm_api_management_logger.this[0].id
+
+  always_log_errors         = true
+  http_correlation_protocol = "W3C"
+  verbosity                 = "information"
+  sampling_percentage       = var.application_insights.sampling_percentage
+}
```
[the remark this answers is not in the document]
kin0992: I'd start with a minimal configuration and then, if we receive feedback related to the customization of the requests and responses, we could improve the module later.
What do you think? @pagopa/engineering-team-devex

infra/modules/azure_api_management/variables.tf
```diff
@@ -191,21 +191,28 @@ variable "lock_enable" {
 
 variable "application_insights" {
   type = object({
-    enabled           = bool
-    connection_string = string
-    id                = optional(string, null)
+    enabled             = bool
+    connection_string   = string
+    id                  = optional(string, null)
+    sampling_percentage = optional(number, 100)
```
gunzip, line 197: **suggestion**: make this mandatory and avoid to default in order to avoid 100% sampling by default.

gunzip (changes requested):

infra/modules/azure_api_management/monitor.tf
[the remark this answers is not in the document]
gunzip: what's the difference between `azuremonitor` and `applicationinsights`?

infra/modules/azure_api_management/monitor.tf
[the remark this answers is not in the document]
gunzip: **question**: does this have performance impact? what happens if I enable application insights without the `azurerm_api_management_diagnostic` resource?

infra/modules/azure_api_management/monitor.tf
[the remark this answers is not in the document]
kin0992: @gunzip honestly, I don't know. I couldn't fine a lot on the documentation and the only test I did with `azuremonitor`, using `${module.apim.id}/loggers/azuremonitor` as `api_management_logger_id` didn't work.

My goal was to compare the data ingested to AI with both configuration, but at the moment the only one that works is the one with `applicationinsights`

infra/modules/azure_api_management/monitor.tf
```diff
@@ -14,6 +14,69 @@ resource "azurerm_api_management_logger" "this" {
...
+
+  always_log_errors         = true
+  http_correlation_protocol = "W3C"
+
+  verbosity           = var.application_insights.verbosity
+  sampling_percentage = var.application_insights.sampling_percentage
+}
+
+// Collect diagnostic logs and metrics to a Log Analytics workspace
+resource "azurerm_monitor_diagnostic_setting" "apim" {
```
kin0992, line 33: **question**: how do we handle existing resources that are not in the terraform state?

here are some details:
- in order to send tracing information to AI, we need the `azurerm_api_management_diagnostic` with `applicationinsights` identifier
- in order to send metrics and logs to azure monitor (in the `AzureDiagnostic` and `AzureMetrics` tables), we need to create both the `azurerm_monitor_diagnostic_setting` and `azurerm_api_management_diagnostic` resources
- now, here is the issue I faced (and keep in mind to test this I created and deleted these resources many times): during the apply terraform fails with this error 
```
Error: A resource with the ID "<subscriptionId>/resourceGroups/dx-d-itn-test-rg-01/providers/Microsoft.ApiManagement/service/dx-d-itn-playground-pg-apim-01/diagnostics/azuremonitor" already exists - to be managed via Terraform this resource needs to be imported into the State. Please see the resource documentation for "azurerm_api_management_diagnostic" for more information.
```
I cannot find that resources anywhere in the Azure Portal. Do you have some hints @pagopa/engineering-team-devex?

infra/modules/azure_api_management/monitor.tf
```diff
@@ -14,6 +14,69 @@ resource "azurerm_api_management_logger" "this" {
...
+      category_group = length(var.monitoring.logs.groups) > 0 ? enabled_log.value : null
+      category       = length(var.monitoring.logs.categories) > 0 ? enabled_log.value : null
+    }
+  }
+
+  # Add metrics if enabled
+  metric {
+    category = "AllMetrics"
+    enabled  = var.monitoring.metrics.enabled
+  }
```
kin0992, line 58: **note**: this is not in a `dynamic` block, because the APIM only has the `AllMetrics` category

infra/modules/azure_api_management/monitor.tf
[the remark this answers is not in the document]
kin0992: BTW, I'm going to move the `azurerm_api_management_diagnostic.applicationinsights` resource in a separate PR, so we can have a complete tracing of the requests within AI.

I'd keep this PR for the monitoring stuff, such as metrics and diagnostics that are persisted in dedicated tables.

infra/modules/azure_api_management/monitor.tf
gunzip: is this still relevant?

kin0992: yes, it is! yesterday I received the following updated from Microsoft Support:
- when a resource `azurerm_monitor_diagnostic_setting` is created, Azure will create an `azuremonitor` logger, the one that has the id ending with `/loggers/azuremonitor`
- the problem is that Azure can take a while to create it, so in the module we cannot reference its id, because it may not be present at the time terraform makes the apply

I am going to ask Microsoft if it is possible to create that logger using a Terraform resource, of we do not have control on it.

kin0992: It looks like that Azure will create the resources required to accomplish our goal after the creation of the `azurerm_monitor_diagnostic_setting` resource.

So, once the monitoring is enabled, eventually we will see the logs in the Log Analytics Workspace.

infra/modules/azure_api_management/variables.tf
```diff
@@ -225,16 +223,89 @@ variable "application_insights" {
...
+      !var.monitoring.logs.enabled ||
+      (length(var.monitoring.logs.groups) == 0) != (length(var.monitoring.logs.categories) == 0)
+    )
+    error_message = "If logs are enabled, exactly one of 'logs.groups' or 'logs.categories' must be provided, but not both."
+  }
+
+  # Validate logs.groups values
+  validation {
+    condition = var.monitoring.logs.enabled == false || alltrue([
+      for group in var.monitoring.logs.groups : contains(data.azurerm_monitor_diagnostic_categories.apim.log_category_groups, group)
```
mamu0, line 295: This `validation` could cause issues for a new APIM instance with monitoring enabled, because the `data` wouldn’t exist yet during the `plan` phase

kin0992: Thanks for the catch, you are right! 🙏 

I've updated the validation.

.changeset/fair-goats-end.md
```diff
@@ -0,0 +1,5 @@
+---
+"azure_api_management": patch
+---
+
+Create APIM diagnostic resource, when AI is enabled, so the requests on the APIM are sent to Application Insights
```
kin0992, line 5:
```suggestion
Create APIM diagnostic setting resource, to forward logs to Log Analytics workspace
```

infra/modules/azure_api_management/variables.tf
```diff
@@ -225,16 +223,75 @@ variable "application_insights" {
...
   validation {
     condition     = var.application_insights.sampling_percentage >= 0 && var.application_insights.sampling_percentage <= 100
-    error_message = "Invalid `sampling_percentage` value provided. Valid values are between 0 and 100"
+    error_message = "Invalid 'sampling_percentage' value provided. Valid values are between 0 and 100"
   }
 
   validation {
     condition     = contains(["verbose", "information", "error"], var.application_insights.verbosity)
-    error_message = "Invalid `verbosity` value provided. Valid values are 'verbose', 'information', 'error'"
+    error_message = "Invalid 'verbosity' value provided. Valid values are 'verbose', 'information', 'error'"
```
Krusty93, line 231: We usually adopt double quotes with escape `\"`

kin0992: I updated them to follow the convention

mamu0 (approved):
Nice Job! 🚀

PR #351, opened 2025-03-13
Fix live terraform plan output in pipeline execution

BurnedMarshal:


BurnedMarshal:
It looks like this is not a solution. The pipeline still does not print a live output from the plan.

gunzip:
Does it works for you? I've tried this solution without success.

gunzip:
I've tried all solutions shared here without success: https://stackoverflow.com/questions/11337041/force-line-buffering-of-stdout-in-a-pipeline

The last one I haven't tried yet is using `script`

https://stackoverflow.com/a/11349234

BurnedMarshal:
> Does it works for you? I've tried this solution without success.

I've tried locally and it works, but in the pipeline it doesn't 😭 (It works on my machine scenario)
Executing locally without the `--line-buffered` option the output stacks until the terraform command complete.
Tested in [this](https://github.com/pagopa/io-infra/pull/1442) PR nothing changes

PR #357, opened 2025-03-14
Enable GitHub autolink reference

kin0992:
GitHub will automatically add a link to the Jira task id

Resolves CES-849


infra/repository/github_repository.tf
```diff
@@ -29,5 +29,12 @@ resource "github_repository" "this" {
       path   = "/"
     }
   }
+}
+
+resource "github_repository_autolink_reference" "jira" {
```
gunzip, line 34: **suggestion**: what about adding this here as well?

https://github.com/pagopa/dx/blob/2aa5329057f031c360d17dee462f974af5d1e30b/infra/modules/azure_github_environment_bootstrap/github_repository.tf#L1

gunzip (approved):

PR #359, opened 2025-03-17
Fix an issue that prevented AppSettings to be displayed in Terraform Plans

Krusty93:
By using Terraform >= 1.10.x, the AppSettings property was marked as sensitive and therefore no longer displayed.

By forcing the boolean local to not being sensitive, the "mark" is not propagated to the `app_settings` resource property, which then doesn't result sensitive anymore

Closes CES-740

### DIFF
--- a/infra/modules/azure_function_app/locals.tf
+++ b/infra/modules/azure_function_app/locals.tf
@@ -31,7 +31,7 @@ locals {
   }
 
   application_insights = {
-    enable = var.application_insights_connection_string != null
+    enable = nonsensitive(var.application_insights_connection_string) != null
   }
 
   storage_account = {
--- a/infra/modules/azure_function_app/locals.tf
+++ b/infra/modules/azure_function_app/locals.tf
@@ -31,7 +31,7 @@ locals {
   }
 
   application_insights = {
-    enable = nonsensitive(var.application_insights_connection_string) != null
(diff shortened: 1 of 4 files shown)


infra/modules/azure_function_app/locals.tf
```diff
@@ -31,7 +31,7 @@ locals {
   }
 
   application_insights = {
-    enable = var.application_insights_connection_string != null
+    enable = nonsensitive(var.application_insights_connection_string) != null
```
gunzip, line 34: isn't

```suggestion
    enable = nonsensitive(var.application_insights_connection_string != null)
```

don't we risk to expose the connection string otherwise?

Krusty93: The resulting boolean value is calculated at runtime, then it will not be stored anywhere.

I haven't verified, but I believe the difference between mine and your code is that whether the `enable` value is stored in the state file or not

gunzip: I'm unsure about marking a sensitive value as 'nonsensitive'

Krusty93: Yes still working

gunzip (approved):

gunzip:
we've moved docs inside `apps/website`

PR #360, opened 2025-03-17
Enable Application Insights tracing on APIM

kin0992:
We want to make it possible to track APIM requests within Application Insights and have a correlation with dependencies

Closes CES-853


infra/modules/azure_api_management/variables.tf
```diff
@@ -191,21 +191,35 @@ variable "lock_enable" {
...
-    enabled           = false
-    connection_string = null
-    id                = null
+    enabled             = false
+    connection_string   = null
+    id                  = null
+    sampling_percentage = 0
+    verbosity           = "error"
   }
   description = "Application Insights integration. The connection string used to push data"
```
gunzip, line 207: **issue**: description must be updated as well (ie. describing the allowed values for verbosity)

gunzip (changes requested):

gunzip (commented):
lgtm. I wonder if it would be better to use managed identities instead of connection string. let's see what @pagopa/engineering-team-devex think.

PR #361, opened 2025-03-17
Add pagopa-dx/azure terraform provider

mamu0:
This PR introduces the Azure Terraform provider to our repository and enhances the existing push-to-subrepo action to support both modules and providers.

#### Changes

- Added `azure` Terraform provider configuration and necessary files into `infra/provider`
- Modified the _push-modules_ into `push-to-subrepo` GitHub Action to detect and handle both module and provider types
- Added workflow to push provider into his subrepo


gunzip:
**suggestion**: split this PR into a 'hello world' provider (skeleton needed to publish the provider) and another PR with the naming convention business logic

mamu0:
> **suggestion**: split this PR into a 'hello world' provider (skeleton needed to publish the provider) and another PR with the naming convention business logic

I understand the reasoning behind splitting the PR, but I believe it would be an unnecessary extra step. Creating a separate “hello world” provider only to replace it immediately with the actual implementation adds no real value and would only slow down the process.

The current PR already includes all the necessary scaffolding to publish the provider. If there’s a concern about reviewing complexity, I can provide additional clarifications or split out specific parts if needed. However, a temporary placeholder provider seems redundant.

Let me know if there’s a specific concern that we can address differently!

infra/provider/internal/provider/function_resource_name.go
```diff
@@ -0,0 +1,237 @@
...
+func (f *resourceNameFunction) Metadata(ctx context.Context, req function.MetadataRequest, resp *function.MetadataResponse) {
+	resp.Name = "resource_name"
+}
+
+func (f *resourceNameFunction) Definition(ctx context.Context, req function.DefinitionRequest, resp *function.DefinitionResponse) {
+	resp.Definition = function.Definition{
+		Summary:     "Return Azure dx resources naming convention",
+		Description: "Given a name, a resource name, an instance number and a resource type, returns the Azure dx resources naming convention.",
+
+		// L'ordine di inserimento è questo
```
Krusty93, line 27:
```suggestion
```

Krusty93 (approved):

PR #369, opened 2025-03-18
Add Keep Alive documentation

mamu0:
This PR add guide for the **Keep Alive** `action` and `workflow`

#### Changes

- Added documentation page for the Keep Alive action and workflow

Resolves: CES-191


gunzip:
**suggestion**: since all actions are used into some workflow, what about documenting only workflows? actions code should be self-documenting imho.

mamu0:
> **suggestion**: since all actions are used into some workflow, what about documenting only workflows? actions code should be self-documenting imho.

That’s a great suggestion, thanks!
However, I have a doubt about this: is it okay to document **non-reusable** workflows by presenting them as configuration templates?
For example, the one in this _PR_ is the **keep-alive workflow**, which runs only in **DX**, and we are not providing a reusable workflow but rather the `action` that teams need to implement as a step in their own scheduled workflow

gunzip:
I'd document action through workflow examples anyway when we don't provide an explict reusable workflow. So it will still go within the 'workflow' subcategory (using an example instead of an actual wf).

gunzip (approved):

PR #378, opened 2025-03-19
Add Azure AppService Plan Terraform module

Krusty93:
Add Azure AppService Plan Terraform module, for internal use only

Closes CES-861


infra/modules/azure_app_service_plan/tests/setup/main.tf
```diff
@@ -0,0 +1,28 @@
...
+    Environment    = "Dev"
+    Owner          = "DevEx"
+    Source         = "https://github.com/pagopa/dx/blob/main/infra/modules/azure_app_service_plan/tests"
+    ManagementTeam = "Developer Experience"
+    Test           = "true"
+    TestName       = "Create AppService Plan for automated tests"
+  }
+}
+
+resource "azurerm_resource_group" "plan_test" {
```
mamu0, line 23: If it’s not necessary to create a new resource group, you could use a data source with the resource group `dx-d-itn-test-rg-01`

mamu0 (approved):

mamu0:
**Question:** Since this plan is the same one that can optionally be created in the `App Service` module, could it be removed from the module so that it’s created separately?

Krusty93:
> **Question:** Since this plan is the same one that can optionally be created in the `App Service` module, could it be removed from the module so that it’s created separately?

That's the intention, but it must be done in two separate PRs as we need this module available on registry first

infra/modules/azure_app_service_plan/tests/setup/main.tf
Krusty93: I think sharing the resource group leads to [this kind of errors](https://github.com/pagopa/dx/actions/runs/13948146029/attempts/1?pr=373)?

mamu0: Since this is a `plan`, it shouldn’t happen.
The idea is to avoid creating a resource group during testing just to run a plan inside it, but if resources need to be created within, with an `apply`, then it’s fine to create a dedicated resource group.

In general, to avoid that issue without adding new resources, we need to ensure that test resources have unique names, but that’s a separate discussion

PR #387, opened 2025-03-24
Update API Management module with fixed zones configuration, PIP support, `xl` tier and new naming logic

mamu0:
This pull request enhances the API Management module with several new features and improvements:

- `Premium_2` **l** tier: Updated tier to support `Premium_2` instead of `Premium_1`
- Zones: Updated availability zones configuration to support 2 and 3 zones based on tier `l` and `xl`
- Public IP support: Implemented optional Public IP configuration variable for more flexible network setups
- `Premium_3` **xl** tier: Added support for the XL capacity tier within the Premium_3 SKU offering
- Improved naming logic: Modified the resource naming approach to better align with naming conventions used

Resolves: CES-881


infra/modules/azure_api_management/tests/apim.tftest.hcl
```diff
@@ -129,3 +129,68 @@ run "plan_with_invalid_parameters" {
...
+    environment = {
+      prefix          = "dx"
+      env_short       = "d"
+      location        = "italynorth"
+      domain          = "modules"
+      app_name        = "test"
+      instance_number = "01"
+    }
+
+    tags = {
```
Krusty93, line 146: **suggestion**: we can you output tags from setup module to avoid repetitions

infra/modules/azure_api_management/variables.tf
```diff
@@ -105,6 +116,25 @@ variable "enable_public_network_access" {
...
+
+variable "private_dns_zone_resource_group_name" {
+  type        = string
+  description = "(Optional) The resource group name of the private DNS zone. This is only required when resource group name is different from the VNet resource group."
+  default     = null
+}
+
+variable "create_network_security_group" {
+  type        = bool
+  description = "Create a network security group for the API Management"
```
Krusty93, line 133: **question**: in what cases it is not needed?

infra/modules/azure_api_management/locals.tf
```diff
@@ -1,15 +1,20 @@
...
+    zones          = length(var.zones_override) > 0 ? var.zones_override : ["1", "2", "3"]
     sku_name = lookup(
       {
-        "s" = "Developer_1",
-        "m" = "Standard_1",
-        "l" = "Premium_1"
+        "s"  = "Developer_1",
+        "m"  = "Standard_1",
+        "l"  = "Premium_1",
+        "xl" = "Premium_2"
```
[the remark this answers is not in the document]
mamu0: Yes, I made a change regarding this. Previously, `l` was incorrectly set to `Premium_1` with 3 zones, but now:

- `l` is `Premium_2` with 2 zones (to align with the current APIM configuration in ITN).
- `xl` is `Premium_3` with 3 zones.

infra/modules/azure_api_management/tests/apim.tftest.hcl
```diff
@@ -76,6 +67,11 @@ run "apim_is_correct_plan" {
     condition     = length(azurerm_api_management_logger.this) > 0
     error_message = "The APIM logger does not exist"
   }
+
+  assert {
+    condition     = length(azurerm_api_management.this.zones) == 2 && contains(azurerm_api_management.this.zones, "1") && contains(azurerm_api_management.this.zones, "2")
+    error_message = "The APIM zones are incorrect, they should be ['1', '2']"
```
gunzip, line 73: question: I cannot understand this condition which looks like:

```suggestion
    condition     = azurerm_api_management.this.zones == ["1", "2"]
    error_message = "The APIM zones are incorrect, they should be ['1', '2']"
```

can't zones just be hardcoded in case?

infra/modules/azure_api_management/variables.tf
mamu0: In general, I believe that can always be created. However, the APIM in ITN currently doesn’t have one, so to avoid creating a new resource, I made it conditional with the default set to true

(but If needed, we can change it to always create the `nsg` and also provision the new resource for the APIM in ITN)

infra/modules/azure_api_management/tests/apim.tftest.hcl
mamu0: This causes the test to fail even if the configuration is correct because the state does not list the zones exactly as in the expected list, making verification unreliable

infra/modules/azure_api_management/variables.tf
Krusty93: Italian APIM has a network security group, it is mandatory with platform stv2. You can find it [here](https://github.com/pagopa/io-infra/blob/8e010eedcc044883b3eb3fa81f6710a9e660fbfc/src/common/_modules/apim/networking.tf#L14)

infra/modules/azure_api_management/locals.tf
```diff
@@ -1,15 +1,20 @@
 locals {
+  prefix = var.environment.app_name != "apim" ? module.naming_convention.prefix : var.environment.domain != null ? "${module.naming_convention.project}-${var.environment.domain}" : module.naming_convention.project
```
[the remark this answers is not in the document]
mamu0: I added a simple description, but the main reason for that _prefix_ is to use the module for the existing **APIM ITN**, which has a non-standard name (without an `app_name`).

infra/modules/azure_api_management/variables.tf
mamu0: thanks a lot! I couldn’t identify it through the state. I’ll revert it to the previous setup and import the existing APIM

gunzip (approved):

infra/modules/azure_api_management/variables.tf
```diff
@@ -45,6 +45,7 @@ variable "autoscale" {
   type = object(
     {
       enabled                       = bool
+      legacy_name                   = optional(string, "")
```
gunzip, line 48: why a default if it's optional?

PR #392, opened 2025-03-26
Add support for Next.js standalone app deployment

lucacavallaro:


.github/workflows/web_app_deploy.yaml
```diff
@@ -71,6 +71,18 @@ jobs:
         run: |
           set -e
 
+          # generate node_modules folder excluding devDependencies
+          yarn workspaces focus --production
+
+          if [ -e "next.config.js" ] && grep -q 'output: "standalone"' "next.config.js"; then
```
gunzip, line 77: question: doesn't this suffice? https://github.com/pagopa/dx-playground/blob/main/apps/to-do-webapp/package.json#L18

lucacavallaro: with the new change you no longer need:
1. to add `.next` in `packageJson.files`
2. to setup pre-build script like https://github.com/pagopa/dx-playground/blob/main/apps/to-do-webapp/package.json#L12-L14

the pipeline does everything by itself, the `"build"` script can also be refactored to simply `"next build"`

the rationale behind our workflow design is that you should not configure nothing if you are using our recommended tools

PR #402, opened 2025-03-28
Use resource name and resource group name in azure_role_assignments

christian-calabrese:
The last major version of azure_role_assignments (1.0.0) removes the data sources causing plan issues when the role assignments' scopes are not yet created.

As a consequence, the resource name and resource group name had been removed as input variables cause computed from the resource id.

Since the resource name and resource group name are used as resource keys in for_each, Terraform’s internal handling still causes issues by ignoring the resource ID until creation.:
image

Based on this, the fix is to reintroduce the name and resource_group_name and build the resource id from them.


Krusty93 (approved):
**thought**: PR is ok to me. We can also consider to set an input variable for the subscription rather than adding a data (and related API call) for each sub module

mamu0:
I see that several modules have been modified, wouldn't it be better to put a patch `changeset`?

christian-calabrese:
> **thought**: PR is ok to me. We can also consider to set an input variable for the subscription rather than adding a data (and related API call) for each sub module

I agree, it would be nice to avoid the data, but we should add an additional variable. What do you think is better?

Krusty93:
> > **thought**: PR is ok to me. We can also consider to set an input variable for the subscription rather than adding a data (and related API call) for each sub module
> 
> I agree, it would be nice to avoid the data, but we should add an additional variable. What do you think is better?

As this is a major and the subscription id is already being used everywhere, I suggest to add the variable

Krusty93 (approved):

PR #413, opened 2025-04-01
Remove `naming_convention` inside App Service modules in favor of `pagopa-dx/azure` provider function

mamu0:
This PR removes all instances of the `naming_convention` module from the App Service and App Service Exposed modules. The functionality has been replaced with the new naming convention function provided by the `pagopa-dx/azure` provider.

Resolves: CES-914


infra/modules/azure_app_service/networking.tf
```diff
@@ -1,11 +1,11 @@
 resource "azurerm_private_endpoint" "app_service_sites" {
-  name                = "${module.naming_convention.prefix}-app-pep-${module.naming_convention.suffix}"
+  name                = provider::dx::resource_name(merge(local.naming_config, { resource_type = "app_private_endpoint" }))
```
Krusty93, line 2: **suggestion**: A local string can be used as the value is the same

Krusty93 (approved):

PR #421, opened 2025-04-02
Remove `naming_convention` inside Cosmos module in favor of `pagopa-dx/azure` provider function

mamu0:
This PR removes all instances of the naming_convention module from the Cosmos module. The functionality has been replaced with the new naming convention function provided by the pagopa-dx/azure provider.

Resolves: CES-918

### DIFF
--- a/infra/modules/azure_cosmos_account/locals.tf
+++ b/infra/modules/azure_cosmos_account/locals.tf
@@ -1,4 +1,16 @@
 locals {
+  naming_config = {
+    prefix      = var.environment.prefix,
+    environment = var.environment.env_short,
+    location = tomap({
+      "italynorth" = "itn",
+      "westeurope" = "weu"
+    })[var.environment.location]
+    domain          = var.environment.domain,
+    name            = var.environment.app_name,
+    instance_number = tonumber(var.environment.instance_number),
+  }
+
   primary_location = var.primary_geo_location.location == null ? var.environment.location : var.primary_geo_location.location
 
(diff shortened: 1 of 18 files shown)


infra/modules/azure_cosmos_account/locals.tf
```diff
@@ -1,4 +1,16 @@
 locals {
+  naming_config = {
+    prefix      = var.environment.prefix,
+    environment = var.environment.env_short,
+    location = tomap({
```
Krusty93, line 5: What if we move this logic (location -> short location) into the provider?

gunzip: sounds good, but I won't do that in the PR

mamu0: #429

infra/modules/azure_cosmos_account/networking.tf
```diff
@@ -2,13 +2,13 @@
 # Private endpoints
 #
 resource "azurerm_private_endpoint" "sql" {
-  name                = "${module.naming_convention.prefix}-cosno-pep-${module.naming_convention.suffix}"
+  name                = provider::dx::resource_name(merge(local.naming_config, { resource_type = "cosmos_private_endpoint" }))
```
Krusty93, line 5: **suggestion**: add this string to a local to reuse the same value below

Krusty93 (approved):

PR #433, opened 2025-04-03
Update provider with Null value support for domain parameter

mamu0:
Update the provider to fix the handling of the domain parameter in the `resource_name` function. Previously, the domain parameter could not be explicitly defined; using an empty string (`""`) was allowed, but passing `null` would result in an error. With this update, the domain parameter now supports `null` values without causing errors.

Resolves: CES-934

### DIFF
--- a/infra/provider/internal/provider/function_resource_name.go
+++ b/infra/provider/internal/provider/function_resource_name.go
@@ -26,9 +26,10 @@ func (f *resourceNameFunction) Definition(ctx context.Context, req function.Defi
 
 		Parameters: []function.Parameter{
 			function.MapParameter{
-				Name:        "configuration",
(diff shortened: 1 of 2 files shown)


infra/provider/internal/provider/function_resource_name.go
```diff
@@ -235,13 +234,14 @@ func (f *resourceNameFunction) Run(ctx context.Context, req function.RunRequest,
 		return
 	}
 
-	// Check if domain is provided
-	if domain != "" {
+	// Check if domain is provided and not null
+	domain, domainExists := configuration["domain"]
```
gunzip, line 238: you also con use:

```go
parts := make([]string, 0, 4)
parts = append(parts, prefix, environment, location)
if domainExists && !domain.IsNull() && domain.ValueString() != "" {
    parts = append(parts, domain.ValueString())
}
result = strings.Join(parts, "-")
```

PR #438, opened 2025-04-03
Add typescript as dependency

kin0992:
This removes the yarn warning '@pagopa/eslint-config@workspace:packages/eslint-config [db966] doesn't provide typescript (paf7fa), requested by typescript-eslint.' and prepare the package to be ready to the migration to the new major of typescript-eslint package.

PR #439, opened 2025-04-03
Create azure-tracing package

kin0992:
Instrument ESM application to send logs to Application Insights using the Azure Monitor Open Telemetry library

Closes CES-794 depends on #438


packages/azure-tracing/src/index.ts
```diff
@@ -0,0 +1,33 @@
...
+ * import { init } from "@azure/monitor-opentelemetry";
+ *
+ * init();
+ * @param instrumentations the list of instrumentations to register with the Azure Monitor.
+ */
+export const init = (instrumentations: readonly Instrumentation[]) => {
+  useAzureMonitor();
+
+  registerInstrumentations({
+    instrumentations: [new UndiciInstrumentation(), ...instrumentations],
```
gunzip, line 29: **suggestion**: consider configuring undici instrumentation as shared in the docs: https://pagopa.github.io/dx/docs/articles/azure-tracing#example-integration-with-app-service

packages/azure-tracing/src/azure-functions.mts
```diff
@@ -0,0 +1,26 @@
...
+import { metrics, trace } from "@opentelemetry/api";
+import { registerInstrumentations } from "@opentelemetry/instrumentation";
+import { UndiciInstrumentation } from "@opentelemetry/instrumentation-undici";
+
+useAzureMonitor();
+
+registerInstrumentations({
+  instrumentations: [
+    new UndiciInstrumentation(),
+    new AzureFunctionsInstrumentation({ enabled: true }),
```
gunzip, line 20: suggestion: I won't include this until they fix [this behavior](https://github.com/Azure/azure-functions-nodejs-opentelemetry/issues/8)

You can obtain the same without verbose debug logging using this commented 
snippet here https://pagopa.github.io/dx/docs/articles/azure-tracing#example-integration-with-app-service

packages/azure-tracing/README.md
```diff
@@ -0,0 +1,65 @@
...
+yarn add @pagopa/azure-tracing
+```
+
+## Getting Started
+
+### Instrumenting Azure Functions
+
+Currently, [ECMAScript Modules (ESM) support in OpenTelemetry is still experimental](https://github.com/open-telemetry/opentelemetry-js/blob/966ac176af249d86de6cb10feac2306062846768/doc/esm-support.md),
+which makes direct instrumentation of Azure Functions a bit tricky.
+To work around this, you can preload the instrumentation logic at runtime using the `NODE_OPTIONS` environment variable.
```
gunzip, line 20: **question**: you can or you must?

packages/azure-tracing/README.md
```diff
@@ -0,0 +1,65 @@
...
+
+```bash
+yarn add @pagopa/azure-tracing
+```
+
+## Getting Started
+
+### Instrumenting Azure Functions
+
+Currently, [ECMAScript Modules (ESM) support in OpenTelemetry is still experimental](https://github.com/open-telemetry/opentelemetry-js/blob/966ac176af249d86de6cb10feac2306062846768/doc/esm-support.md),
```
gunzip, line 18: **issue**: it's not clear if the package supports both ESM and CJS

gunzip (dismissed):

packages/azure-tracing/src/azure-monitor.ts
```diff
@@ -0,0 +1,55 @@
...
+ */
+export const initAzureMonitor = (
+  instrumentations: readonly Instrumentation[] = [],
+) => {
+  useAzureMonitor();
+
+  registerInstrumentations({
+    instrumentations: [
+      // instrument native node fetch
+      new UndiciInstrumentation({
```
gunzip, line 33: **thought**: we can export this custom configuration in its own module (ie. `azure-undici-instrumentation.ts`)

packages/azure-tracing/src/azure-functions.mts
```diff
@@ -0,0 +1,62 @@
...
+      .transform((value) => value / 100),
+  },
+});
+
+useAzureMonitor({
+  azureMonitorExporterOptions: {
+    connectionString: env.AI_CONNECTION_STRING,
+  },
+  enableLiveMetrics: true,
+  samplingRatio: env.AI_SAMPLING_PERCENTAGE,
```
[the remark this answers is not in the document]
gunzip: For Node.js App Services, it is preferable NOT to use the default Application Insights integration, and therefore NOT to use the default environment variable name `APPLICATIONINSIGHTS_CONNECTION_STRING`.

However, for Azure Functions, you _need_ to enable integration with the host using the `APPLICATIONINSIGHTS_CONNECTION_STRING` environment variable.

If we use a different name here (e.g., `AI_CONNECTION_STRING`), we can enable both scenarios. For Azure Functions only, we should then set `AI_CONNECTION_STRING` = `APPLICATIONINSIGHTS_CONNECTION_STRING` in the Terraform module configuration.

If we use `APPLICATIONINSIGHTS_CONNECTION_STRING` here instead, it will force the default integration even on App Services. 

Another option would be to let user pass both arguments.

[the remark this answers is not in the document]
kin0992: that's why I'd use the `APPINSIGHTS_CONNECTION_STRING` variable instead and the user will take care to set that variable in the terraform configuration

packages/azure-tracing/tsup.config.ts
```diff
@@ -0,0 +1,18 @@
...
+
+export default defineConfig({
+  bundle: true,
+  clean: true,
+  dts: true,
+  entry: [
+    "src",
+    "!src/azure/monitor/start-from-env.ts",
+    "!src/azure/monitor/env.ts",
+    "!src/azure/opentelemetry/azure-undici-instrumentation.ts",
```
kin0992, line 11: **note**: these are excluded from the bundle.

we do not need to expose those files

gunzip: can the undici configuration be useful in case of tracing within nextjs?

kin0992: In Next.js application, it is possible to [add instrumentation using the `initAzureMonitor` function, as done in the playground](https://github.com/pagopa/dx-playground/pull/153/files#diff-2756329cd991d4158a3451478792c29a4de4acafca7ddea3064c7206909868b4R4).

The `initAzureMonitor` already instrument undici, so I don't think, at the moment, it is necessary to export it

packages/azure-tracing/README.md
kin0992: It should be.
In the readme, they said
> Much of OpenTelemetry JS documentation is written assuming the compiled application is run as CommonJS. For more details on ECMAScript Modules vs CommonJS, refer to [esm-support](https://github.com/open-telemetry/opentelemetry-js/blob/main/doc/esm-support.md).

So I guess it is possible to use that package in a "standard" way, like importing a file at the beginning of the main file, while for ESM it is required to load it on start-up

gunzip: I'd add this to docs

kin0992: Improved the README

packages/azure-tracing/src/azure-functions.mts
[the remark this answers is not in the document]
kin0992: @gunzip, I added the `APPINSIGHTS_CONNECTION_STRING` environment variable in the Terraform modules

.changeset/three-numbers-knock.md
```diff
@@ -0,0 +1,10 @@
+---
+"azure_app_service": patch
+"azure_app_service_exposed": patch
+"azure_function_app": patch
+"azure_function_app_exposed": patch
+---
+
+Add `APPINSIGHTS_CONNECTION_STRING` as an environment variable when application insights is enabled.
+
+The `APPINSIGHTS_CONNECTION_STRING` is used by the `@pagopa/azure-tracing`.
```
kin0992, line 10:
```suggestion
The `APPINSIGHTS_CONNECTION_STRING` is used by the `@pagopa/azure-tracing` package.
```

packages/azure-tracing/package.json
```diff
@@ -0,0 +1,57 @@
...
+    "@pagopa/eslint-config": "workspace:^",
+    "@tsconfig/node20": "^20.1.5",
+    "@types/node": "^20.12.2",
+    "eslint": "8.57.0",
+    "shx": "^0.4.0",
+    "tsup": "^8.4.0",
+    "typescript": "~5.2.2"
+  },
+  "scripts": {
+    "clean": "shx rm -rf dist",
```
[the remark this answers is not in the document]
kin0992: you are right, thanks for the suggestion!

infra/modules/azure_app_service/app_service.tf
```diff
@@ -46,6 +46,8 @@ resource "azurerm_linux_web_app" "this" {
     local.application_insights.enable ? {
       # https://learn.microsoft.com/en-us/azure/azure-functions/functions-app-settings#applicationinsights_connection_string
       APPLICATIONINSIGHTS_CONNECTION_STRING = var.application_insights_connection_string
+      # Add environment variable used by the `@pagopa/azure-tracing` package
+      APPINSIGHTS_CONNECTION_STRING = var.application_insights_connection_string
```
gunzip, line 50: **issue**: why use two different variables with the same value? We probably should leave app service / functions users the choice to enable or not the default integration with AI. Removing `APPINSIGHTS_CONNECTION_STRING` from the module would let them use the tracing package without enabling the default integration (which is useful in case of app services).

PR #445, opened 2025-04-04
Remove unsupported characters from Container App module

Krusty93:
Remove unsupported characters from secret names

Close CES-927

### DIFF
--- a/.changeset/silent-wombats-mix.md
+++ b/.changeset/silent-wombats-mix.md
@@ -1,0 +1,5 @@
+---
+"azure_container_app": patch
+---
+
+Replace some unsupported characters from secret names


mamu0:
Wouldn't it be simpler to add a `validation` rule in the `secrets` variable to prevent names with underscores (`_`)? This would enforce the naming convention directly at the source

gunzip (approved):

Krusty93:
> **suggestion**: Wouldn't it be simpler to add a `validation` rule in the `secrets` variable to prevent names with underscores (`_`)? This would enforce the naming convention directly at the source
> 
> ```hcl
> variable "secrets" {
>   type = list(object({
>     name                = string
>     key_vault_secret_id = string
>   }))
>   default     = []
>   description = "Key Vault secret references to be used in all the containers of this Container App."
> 
>   validation {
>     condition     = alltrue([for secret in var.secrets : !strcontains(secret.name, "_")])
>     error_message = "The 'name' field in secrets must not contain underscores (_)."
>   }
> }
> ```

It would be, but I don't want to introduce another variable for the secret names (env vars accept any chars, secrets don't)

PR #450, opened 2025-04-07
Auto generate container name from image for Container Apps

Krusty93:
This PR addresses an issue regarding the container name assignment. In fact, if not provided through the proper variable, the container name was obtained from the image name and tags. However, the latter contains unsupported characters (e.g. `/`, `:`). This PR removes these characters and instead, get the image name only.

Close CES-941

### DIFF
--- a/infra/modules/azure_container_app/tests/containerapp.tftest.hcl
+++ b/infra/modules/azure_container_app/tests/containerapp.tftest.hcl
@@ -24,14 +24,7 @@ run "container_app_is_correct_plan" {
   command = plan
 
   variables {
-    environment = {
-      prefix          = "dx"
-      env_short       = "d"
-      location        = "italynorth"
-      domain          = "modules"
-      app_name        = "test"
-      instance_number = "01"
-    }
+    environment = run.setup_tests.environment
 
     tags = run.setup_tests.tags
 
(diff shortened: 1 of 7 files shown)


infra/modules/azure_container_app/tests/containerapp.tftest.hcl
```diff
@@ -79,6 +71,11 @@ run "container_app_is_correct_plan" {
     error_message = "The container app image is not correct"
   }
 
+  assert {
+    condition     = azurerm_container_app.this.template[0].container[0].name == "nginx"
+    error_message = "The container app container name auto-generated is wrong"
```
gunzip, line 76: **issue**: can we add info about why is wrong and how to make it right?

gunzip (changes requested):

infra/modules/azure_container_app/tests/containerapp.tftest.hcl
Krusty93: Consider this is the output for unit tests, so every value except for the desired one is wrong. There's a comment regarding what the value shall be on [the line where the logic is defined](https://github.com/pagopa/dx/pull/450/files#diff-baefbf9b8f147fb113a978b79fa84b1ecd4d503873869d556125bcb410ad8840)

mamu0 (approved):

gunzip (approved):

PR #458, opened 2025-04-08
Fix eventhub name

mamu0:
Resolves: CES-942

### DIFF
--- a/.changeset/lemon-radios-knock.md
+++ b/.changeset/lemon-radios-knock.md
@@ -1,0 +1,5 @@
+---
+"azure_event_hub": patch
+---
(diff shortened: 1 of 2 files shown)


.changeset/lemon-radios-knock.md
```diff
@@ -0,0 +1,5 @@
+---
+"azure_event_hub": patch
```
gunzip, line 2: isn't this a breaking change?

mamu0: this fix the braking change

Krusty93 (approved):

PR #462, opened 2025-04-09
Add Documentation for Storage Account module

mamu0:
Add documentation inside Storage account module and updated variables description.

Resolves: CES-185


infra/modules/azure_storage_account/variables.tf
```diff
@@ -46,19 +46,19 @@ variable "customer_managed_key" {
...
 }
 
 variable "force_public_network_access_enabled" {
   type        = bool
-  description = "(Optional) Whether the Storage Account permits public network access or not. Defaults to false."
+  description = "Allows public network access. Defaults to 'false'."
   default     = false
 }
 
 variable "access_tier" {
```
gunzip, line 59: **suggestion**: out of scope for this PR, but we could add a backlog task about a possible refacgoring here. this is currently a [leaky abstraction](https://en.wikipedia.org/wiki/Leaky_abstraction). we should not expose the `access_tier` directly but hiding it under some use-case tiering, for example: "archive", "realtime", "test", ...

gunzip (approved):

infra/modules/azure_storage_account/variables.tf
mamu0: It might be a good idea to simplify the user experience, I’ll create a task for it. Thanks for the suggestion!

PR #486, opened 2025-04-11
Container App Environment module exports all details of the user-assigned managed identity

Krusty93:
Close CES-949

### DIFF
--- a/.changeset/open-mammals-post.md
+++ b/.changeset/open-mammals-post.md
@@ -1,0 +1,9 @@
+---
+"azure_container_app_environment": major
+---
+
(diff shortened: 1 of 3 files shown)


christian-calabrese (approved):

.changeset/open-mammals-post.md
```diff
@@ -0,0 +1,9 @@
+---
+"azure_container_app_environment": major
```
mamu0, line 2: Maybe a `minor` version would be enough here? There’s no new feature, it’s just an output that has changed

PR #501, opened 2025-04-15
Add workflow template for Docker builds

Krusty93:
Close CES-957


.github/workflows/docker_image_build.yaml
```diff
@@ -37,58 +37,27 @@ on:
```
mamu0, line 37: To comply with the naming convention, could you change the name of the workflow? You could use something like `build-docker-image-v1.yaml`

mamu0: @Krusty93 the validation workflow has been updated, if you merge from `main`, it won’t fail anymore 👍

mamu0 (approved):

PR #513, opened 2025-04-22
Add provider resource to retrieve first available subnet CIDR in Azure VNet

mamu0:
This PR introduces a new resource named `dx_available_subnet_cidr` in the `pagopa-dx/azure` provider.

Functionality:

- Identifies the first available CIDR block within a specified Azure Virtual Network.
- Accepts:
  - `virtual_network_id`
  - `prefix_length`

Output:

-  Returns the first unallocated subnet CIDR block matching the requested prefix length.

Resolves: CES-908


infra/provider/internal/provider/datasource_available_subnet_cidr.go
```diff
@@ -0,0 +1,315 @@
...
+
+// availableSubnetCidrDataSourceModel describes the data source data
+type availableSubnetCidrDataSourceModel struct {
+	VirtualNetworkID types.String `tfsdk:"virtual_network_id"`
+	PrefixLength     types.Int64  `tfsdk:"prefix_length"`
+	CidrBlock        types.String `tfsdk:"cidr_block"`
+}
+
+func (d *availableSubnetCidrDataSource) Metadata(ctx context.Context, req datasource.MetadataRequest, resp *datasource.MetadataResponse) {
+	resp.TypeName = req.ProviderTypeName + "_available_subnet_cidr"
```
mamu0, line 43: **question**: about naming conventions for the data source, is `dx_available_subnet_cidr` acceptable to everyone? @pagopa/engineering-team-devex
Or would it be preferable to avoid using `dx_` at the beginning to make it more generic?

gunzip (commented):
very good! what about using a resource instead of data source? in this way the value is saved into the state during plan and applied during apply. in case of conflicts it would give an error. similar semantic as https://registry.terraform.io/providers/hashicorp/random/latest/docs

mamu0:
> very good! what about using a resource instead of data source? ...

Thank you for the suggestion! 🚀 
For now, I’ve added a resource that calculates the new CIDR block, functioning similarly to the data source. The only limitation is that the CIDR value won’t be known during the plan phase. Additionally, if you want to create multiple subnets, you’ll need to use this resource multiple times, setting `depends_on` to ensure the subnets are created sequentially.

infra/provider/internal/provider/datasource_available_subnet_cidr.go
Krusty93: To me is ok to remove the prefix as the provider name already contains `dx`

mamu0: I’ve realized that the resource type name must actually start with the provider prefix, so it’s required to begin with `dx_`, sorry

infra/provider/docs/resources/available_subnet_cidr.md
```diff
@@ -0,0 +1,83 @@
...
+  virtual_network_name = azurerm_virtual_network.main.name
+  address_prefixes     = [dx_available_subnet_cidr.next_cidr_2.cidr_block]
+}
+```
+
+## Schema
+
+### Required
+
+- `prefix_length` (Number) The desired prefix length for the new subnet CIDR (e.g., 24 for a /24 subnet). Must be larger than the VNet prefix and smaller or equal to 29.
```
Krusty93, line 65: Why smaller than 29?

mamu0: i've tried to create a new subnet on Azure Portal to be sure about this, and the max limit is 29, min is 16
image

Krusty93 (approved):

PR #518, opened 2025-04-23
Docusaurus preview

christian-calabrese:


lucacavallaro:
Have you already considered using [Azure Static Web Apps](https://azure.microsoft.com/en-us/products/app-service/static#Pricing-4)? This service already includes support for preview environments (meaning less custom code for us), and by leveraging it, we can create a solution that's reusable by other teams with similar use cases.

I think as DX team we should use the same solutions we make available to others.

christian-calabrese:
> Have you already considered using [Azure Static Web Apps](https://azure.microsoft.com/en-us/products/app-service/static#Pricing-4)? This service already includes support for preview environments (meaning less custom code for us), and by leveraging it, we can create a solution that's reusable by other teams with similar use cases.
> 
> I think as DX team we should use the same solutions we make available to others.

I agree on the fact that we should use reusable strategies to be provided to teams-

For now, it's just a test. If we agree on using azure/aws services for publishing the documentation and blog, we should think of the domain change (from pagopa.github.io/dx/* to something else) and effort.

PR #523, opened 2025-04-29
Set permissions on GitHub internal workflow jobs

Krusty93:
Close CES-976

### DIFF
--- a/.github/workflows/_cron-terraform-drift-detection.yaml
+++ b/.github/workflows/_cron-terraform-drift-detection.yaml
@@ -9,6 +9,8 @@ jobs:
   drift_detection:
     uses: ./.github/workflows/infra_drift_detection.yml
     name: Drift Detection
+    permissions:
+      contents: read
     secrets: inherit
     with:
       environment: 'dev'
--- a/.github/workflows/_release-bash-modules-to-subrepo.yaml
+++ b/.github/workflows/_release-bash-modules-to-subrepo.yaml
@@ -15,6 +15,8 @@ concurrency:
 jobs:
   detect-modules:
     runs-on: ubuntu-latest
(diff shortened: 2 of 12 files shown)


christian-calabrese:
**note**: previously, in workflows where the permissions block was not explicitly defined, [permissive access](https://docs.github.com/en/actions/security-for-github-actions/security-guides/automatic-token-authentication#permissions-for-the-github_token) was automatically granted according to dx repository configuration (see screenshot below).
image

If any of these workflows rely on permissions beyond `contents: read`, they may fail, since [omitting permissions now defaults to no access](https://docs.github.com/en/actions/security-for-github-actions/security-guides/automatic-token-authentication#modifying-the-permissions-for-the-github_token) (see the note section).

That said, I think that since all affected workflows are internal, we can safely adjust their permissions incrementally as needed.

Krusty93:
> **note**: previously, in workflows where the permissions block was not explicitly defined, [permissive access](https://docs.github.com/en/actions/security-for-github-actions/security-guides/automatic-token-authentication#permissions-for-the-github_token) was automatically granted according to dx repository configuration (see screenshot below). image
> 
> If any of these workflows rely on permissions beyond `contents: read`, they may fail, since [omitting permissions now defaults to no access](https://docs.github.com/en/actions/security-for-github-actions/security-guides/automatic-token-authentication#modifying-the-permissions-for-the-github_token) (see the note section).
> 
> That said, I think that since all affected workflows are internal, we can safely adjust their permissions incrementally as needed.

Agree, for this reason I've updated only internal workflows. As testing all workflows is tedious, I prefer a try/error approach as none of these workflows is critical

christian-calabrese (approved):

PR #531, opened 2025-04-30
Add workflow template to deploy Container Apps using multiple mode

Krusty93:
Add a workflow template to deploy a Container App by using Canary deployment.

The workflow builds the Docker image and push it to the GitHub registry. Tagging and labels are consistent with OCI specs.

## Workflow

1. Build and push the Docker image to the registry using the `docker-build-push`
   action
2. Determine the Container App’s revision mode: **Single** or **Multiple**
3. In **Single** mode, update the existing revision with the new image, routing
   100% of traffic immediately
4. In **Multiple** mode (canary rollout):
   1. Create a new revision with the updated image
   2. Wait until the new revision reports **Healthy** status
   3. If a script to verify the application health* is configured, shift traffic
      incrementally based on its JSON output:
      - Use a `swap` flag to cut over immediately (100% traffic to new revision)
      - Use `nextIncrementPercentage` and `afterMs` to schedule gradual rollouts
   4. If no script is provided, perform a full switch (100% to new revision)
5. After successful switch, deactivate the old revision
6. In case of failure, the new revision is deactivated, while the old revision
   remains active

*currently, only typescript (via npm or yarn) is supported

Close CES-958


.github/actions/docker-build-push/action.yaml
```diff
@@ -17,11 +17,16 @@ inputs:
     required: false
     default: ${{ github.repository }}
     description: Docker image name, defaults to repository name
-  image_description:
+  apply_github_tags_to_image_tags:
+    type: boolean
+    required: false
+    default: false
+    description: Use GitHub tags for image tagging
```
gunzip, line 24: **question**: what does it mean?

deploy-azure-containerapp-v1.yaml
```diff
@@ -0,0 +1,306 @@
...
+        description: Docker image name, defaults to repository name
+      docker_image_description:
+        type: string
+        required: true
+        description: A description to use as image label
+      docker_image_authors:
+        type: string
+        required: false
+        default: PagoPA
+        description: Authors names to use as image label
```
gunzip, line 27: issue: add the expected format (ie. "comma delimited")

deploy-azure-containerapp-v1.yaml
```diff
@@ -0,0 +1,306 @@
...
+      build_args:
+        description: List of build arguments to use for Dockerfile build, given in env=value format.
```
gunzip, line 29: **issue**: comma delimited?

deploy-azure-containerapp-v1.yaml
```diff
@@ -0,0 +1,306 @@
...
+        type: string
+        required: false
+      build_platforms:
+        type: string
+        required: false
+        default: linux/amd64
+        description: Image runtime platform, supports multiple values
```
gunzip, line 36: question: what does it mean "multiple values"? comma separated list of... ?

deploy-azure-containerapp-v1.yaml
```diff
@@ -0,0 +1,306 @@
...
+
+      - name: Allow AZ CLI Extensions
+        run: |
+          az config set extension.use_dynamic_install=yes_without_prompt
+
+      - name: Set Default Resource Group
+        run: |
+          az configure --defaults group=${{ env.RESOURCE_GROUP_NAME }}
+
+      # - name: Get Log Analytics Workspace ID
```
gunzip, line 139: issue: avoid committing commented out code

deploy-azure-containerapp-v1.yaml
```diff
@@ -0,0 +1,306 @@
...
+            echo "error: New revision did not become healthy within the timeout period"
+            exit 1
+          fi
+
+      - name: Check for New Revision Errors and Increment Traffic
+        if: ${{ steps.get_revision_mode.outputs.revision_mode == 'Multiple' && steps.create_new_revision.outputs.skip_remaining_steps == 'false' }}
+        run: |
+          new_revision_name=${{ steps.create_new_revision.outputs.new_revision_name }}
+
+          for percentage in 5 10 30 60 90 100
```
gunzip, line 251: issue: in the absence of a script that drives the increment percentage imho the pipeline should just do a blue/green deployment without any logic.

deploy-azure-containerapp-v1.yaml
Krusty93: The docker build action [requires this format](https://github.com/docker/build-push-action#inputs). Requiring commas and then removing them would make things far more complicated

deploy-azure-containerapp-v1.yaml
Krusty93: I specified to use a comma separated list as [required](https://docs.docker.com/reference/cli/docker/buildx/build/#platform)

deploy-azure-containerapp-v1.yaml
Krusty93: I do not agree on this one, as the deployment mode is chosen at the [infrastructure level](https://registry.terraform.io/modules/pagopa-dx/azure-container-app/azurerm/latest#input_revision_mode). The workflow should support both, even though monitor capabilities are not provided (yet). I can still see advantages in using this deployment model though, as alarms can be triggered before the entire customer base is served by the new version

gunzip: In my opinion, the pipeline shouldn't have any inherent knowledge of the desired deployment workflow. The user script should manage the incremental rollout based on parameters **known to the service developer**. If no script is provided, the default behavior should be a blue/green deployment, which aligns with the current approach for applications and is preferable in certain scenarios (e.g., when a rapid rollout is desired).

.github/actions/docker-build-push/action.yaml
Krusty93: I was convinced to have read somewhere that the tag was taken from the GitHub release page rather than the git tag, but I cannot find the reference anymore.
I've updated the code removing this variable and tagging the image with the git tag, if available. The output is something similar to the following screenshot, to allow clients to pin a version exploiting semver:

image

.github/actions/docker-build-push/action.yaml
```diff
@@ -60,39 +60,57 @@ runs:
...
+        labels: |
+          org.opencontainers.image.title=${{ env.IMAGE_NAME }}
           org.opencontainers.image.description=${{ env.DESCRIPTION }}
           org.opencontainers.image.authors=${{ env.AUTHORS }}
           org.opencontainers.image.url=github.com/pagopa/${{ env.REPOSITORY }}
           org.opencontainers.image.source=https://github.com/${{ env.IMAGE_NAME }}
 
     - name: ${{ inputs.push_to_registry && 'Build and Push Image' || 'Build Image' }}
       uses: docker/build-push-action@471d1dc4e07e5cdedd4c2171150001c434f0b7a4 # v6.15.0
-      if: ${{ inputs.push_to_registry }}
```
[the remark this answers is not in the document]
Krusty93: The input is used as parameter to activate/disable the push at line 95

deploy-azure-containerapp-v1.yaml
gunzip: in other words, I woulnd't implement any complex logic within the pipeline itself. if you do you prevent fast deploys forcing all users to use an incremental rollout based on fixed percentage (when sometimes it's desiderable to just have a blue/green setup or it's perfectly fine to go from %5 to 100%)

mamu0: Maybe an hybrid approach could be to **default** to a `canary` rollout with predefined percentages, while allowing users to override the behavior via input parameters, choosing between canary and blue/green, and optionally customizing the rollout steps.

For example:
```
inputs:
  ...
  deployment_strategy:
    description: "Deployment strategy: canary or blue-green"
    required: false
    default: "canary"
  rollout_percentages:
    description: "Optional: custom rollout percentages for canary strategy"
    required: false
    default: ""
  ...
```

gunzip: what's the advantage  of adding complexity (parameters) within the pipeline instead of delegaing configuration to an external script?

Krusty93: I have updated the code to run external NPM scripts. We can evaluate soon to evolve json schemas, produce SDKs, or support more languages

.github/workflows/release-azure-containerapp-v1.yaml
```diff
@@ -0,0 +1,316 @@
...
+        description: Name of the container app to deploy
+      resource_group_name:
+        type: string
+        required: true
+        description: Name of the resource group of the container app
+      environment:
+        description: Environment where the image will be deployed
+        type: string
+        required: true
+      npm_script:
```
gunzip, line 49: thoughts: why npm script? cant we just have a command line here? in order to let users chose the language they prefer.

.github/workflows/release-azure-containerapp-v1.yaml
```diff
@@ -0,0 +1,316 @@
...
+            --revision-weight $new_revision_name=100
+
+            exit 0
+          fi
+
+          currentPercentage=0
+
+          while true; do
+
+            output="$(npm run -s "$NPM_SCRIPT" $currentPercentage)"
```
mamu0, line 249: Shouldn’t there be a step like this for using npm?
```
      - name: Node Setup
        uses: pagopa/dx/.github/actions/node-setup@main
```

.github/workflows/release-azure-containerapp-v1.yaml
mamu0: I think this could be a good idea, but it would require an additional input to specify the chosen script language.
(A set of supported languages should also be defined in advance to correctly configure the script start command)

gunzip: what about having just one input like `script_language` and derive the script name from this value?

.github/workflows/release-azure-containerapp-v1.yaml
```diff
@@ -0,0 +1,367 @@
...
+
+      - name: (DEBUG) Checkout DX
+        uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2
+        if: ${{ steps.get_revision_mode.outputs.revision_mode == 'Multiple' && steps.create_new_revision.outputs.skip_remaining_steps == 'false' }}
+        with:
+          repository: pagopa/dx
+          ref: CES-958-definire-una-pipeline-di-cd-per-le-container-app
+          path: dx
+
+      - name: (DEBUG) Build yarn
```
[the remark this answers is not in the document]
Krusty93: I'm building a script to test if the workflow is able to run it and has required roles to query log analytics workspace

.github/workflows/release-azure-containerapp-v1.yaml
```diff
@@ -0,0 +1,367 @@
...
+
+      - name: Create New Container App Revision
+        id: create_new_revision
+        # if: ${{ steps.get_revision_mode.outputs.revision_mode == 'Multiple' }}
+        env:
+          CURRENT_REVISION: ${{ steps.get_current_revision.outputs.current_revision_name }}
+          SHORT_SHA: ${{ steps.get_commit_sha.outputs.short_sha }}
+          REVISION_MODE: ${{ steps.get_revision_mode.outputs.revision_mode == 'Multiple' }}
+        run: |
+          new_revision=$(az containerapp revision copy \
```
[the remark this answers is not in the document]
gunzip: how do you setup different values for this new revision which is cloned ?

.github/workflows/release-azure-containerapp-v1.yaml
Krusty93: Workflow updated!

.github/workflows/release-azure-containerapp-v1.yaml
```diff
@@ -0,0 +1,376 @@
...
+              echo "::notice::Swapping all traffic to next revision."
+
+              az containerapp ingress traffic set \
+                --name "$CONTAINER_APP_NAME" \
+                --revision-weight "$CURRENT_REVISION_NAME"=0 \
+                --revision-weight "$NEW_REVISION_NAME"=100
+              break
+            fi
+
+            nextIncrementPercentage=$(echo "$output" | jq -r '.nextIncrementPercentage')
```
gunzip, line 319:
```suggestion
            nextPercentage=$(echo "$output" | jq -r '.nextIncrementPercentage')
```

if you call it with 'increment' it looks like the value only contains the delta (increment) while if I understand well it represents the whole percentage.

infra/modules/azure_container_app/container_app.tf
```diff
@@ -142,5 +142,11 @@ resource "azurerm_container_app" "this" {
     }
   }
 
+  lifecycle {
```
gunzip, line 145: suggestion: I'd add a comment about why this is needed

.github/workflows/release-azure-containerapp-v1.yaml
```diff
@@ -0,0 +1,376 @@
...
+    permissions:
+      contents: read
+      id-token: write
+      attestations: write
+      packages: write
+
+    steps:
+
+      - uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2
+        name: Checkout
```
gunzip, line 79: thought: what about adding `fetch-depth:1`?

gunzip (commented):
what about adding` set -euo` on every step?

.github/workflows/release-azure-containerapp-v1.yaml
```diff
@@ -0,0 +1,376 @@
...
+
+          echo "ca_id=$container_app_id" >> $GITHUB_OUTPUT
+
+      - name: Check for New Revision Errors and Increment Traffic
+        if: ${{ steps.get_revision_mode.outputs.revision_mode == 'Multiple' && steps.create_new_revision.outputs.skip_remaining_steps == 'false' }}
+        env:
+          SCRIPT_NAME: ${{ inputs.script_name }}
+          NEW_REVISION_NAME: ${{ steps.create_new_revision.outputs.new_revision_name }}
+          CURRENT_REVISION_NAME: ${{ steps.get_current_revision.outputs.current_revision_name }}
+          FUNCTION_APP_NAME: "io-p-itn-auth-lollipop-func-02"
```
gunzip, line 259: suggestion: mark all the debug code in order to let reviewers tell it from the real one

.github/workflows/release-azure-containerapp-v1.yaml
Krusty93: Time saving is minimal to me

.github/workflows/release-azure-containerapp-v1.yaml
Krusty93: Updated

.github/workflows/release-azure-containerapp-v1.yaml
gunzip: I guess it depends on the repo size and history

apps/website/docs/pipelines/release-container-app.md
```diff
@@ -0,0 +1,77 @@
...
+This documentation covers the Container App Release reusable workflow defined in
+[release-azure-containerapp-v1.yaml](https://github.com/pagopa/dx/blob/main/.github/workflows/release-azure-containerapp-v1.yaml).
+
+## How It Works
+
+The workflow performs the following steps:
+
+1. Build and push the Docker image to the registry using the `docker-build-push`
+   action
+2. Determine the Container App’s revision mode: **Single** or **Multiple**
```
gunzip, line 17: suggestion: link to the related MS documentation

apps/website/docs/pipelines/release-container-app.md
```diff
@@ -0,0 +1,77 @@
...
+3. In **Single** mode, update the existing revision with the new image, routing
+   100% of traffic immediately
+4. In **Multiple** mode (canary rollout):
+   1. Create a new revision with the updated image
+   2. Wait until the new revision reports **Healthy** status
```
gunzip, line 22: question: how? it's some user defined endpoint with custom code?

apps/website/docs/pipelines/release-container-app.md
```diff
@@ -0,0 +1,77 @@
...
+   3. If a script to verify the application health is configured, shift traffic
+      incrementally based on its JSON output:
+      - Use a `swap` flag to cut over immediately (100% traffic to new revision)
+      - Use `nextPercentage` and `afterMs` to schedule gradual rollouts
```
gunzip, line 26: suggestion: I'd just link the section below, ie. 

"To implement a canary deployment, you can set up [a custom script](<link here>) within your pipeline that guides the deployment process using its output"

apps/website/docs/pipelines/release-container-app.md
```diff
@@ -0,0 +1,77 @@
...
+5. After successful switch, deactivate the old revision
+6. In case of failure, the new revision is deactivated, while the old revision
+   remains active
+
+### Implementing a Canary Test Script
+
+For canary deployments, add an NPM script to test the new version and drive
+workflow behaviour.
+
+The following schemas are supported:
```
gunzip, line 37:
```suggestion
The following schemas for the script output are supported:
```

apps/website/docs/pipelines/release-container-app.md
```diff
@@ -0,0 +1,77 @@
...
+
+```json
+{
+  "swap": true
+}
+```
+
+Use this schema to immediate cutover to the new version.
```
gunzip, line 45:
```suggestion
Output such payload to immediate cutover to the new version.
```

apps/website/docs/pipelines/release-container-app.md
```diff
@@ -0,0 +1,77 @@
...
+Instead, for gradual rollout use:
+
+```json
+{
+  "nextPercentage": 50,
+  "afterMs": 30000
+}
+```
+
+- `nextPercentage`: percentage of traffic for the next iteration.
```
gunzip, line 58:
```suggestion
- `nextPercentage`: percentage of users traffic that will flow to the new revision in the next iteration of the script.
```

apps/website/docs/pipelines/release-container-app.md
```diff
@@ -0,0 +1,77 @@
...
+{
+  "swap": true
+}
+```
+
+Use this schema to immediate cutover to the new version.
+
+- `swap`: when `true`, switches all traffic to the new revision immediately.
+
+Instead, for gradual rollout use:
```
gunzip, line 49:
```suggestion
Instead, for gradual rollout output:
```

apps/website/docs/pipelines/release-container-app.md
```diff
@@ -0,0 +1,77 @@
...
+      - Use `nextPercentage` and `afterMs` to schedule gradual rollouts
+   4. If no script is provided, perform a full switch (100% to new revision)
+5. After successful switch, deactivate the old revision
+6. In case of failure, the new revision is deactivated, while the old revision
+   remains active
+
+### Implementing a Canary Test Script
+
+For canary deployments, add an NPM script to test the new version and drive
+workflow behaviour.
```
gunzip, line 35:
```suggestion
workflow behaviour. This script can check the results of integration tests and/or monitoring metrics before outputting a payload that influence the workflow.
```

something alike.

apps/website/docs/pipelines/release-container-app.md
Krusty93: Added

apps/website/docs/pipelines/release-container-app.md
Krusty93: Added but don't think it's quite useful

apps/website/docs/pipelines/release-container-app.md
Krusty93: added

PR #536, opened 2025-05-05
Disable local authentication for Cosmos DB accounts

mamu0:
This PR disables local authentication for Cosmos DB accounts by setting `local_authentication_disabled = true`. This security enhancement aligns with Technology policies.

Resolves: CES-980

### DIFF
--- a/.changeset/rotten-taxes-slide.md
+++ b/.changeset/rotten-taxes-slide.md
@@ -1,0 +1,5 @@
+---
+"azure_cosmos_account": minor
+---
+
+Added `local_authentication_disabled = true` to Azure Cosmos DB module
--- a/infra/modules/azure_cosmos_account/cosmos.tf
+++ b/infra/modules/azure_cosmos_account/cosmos.tf
@@ -7,6 +7,7 @@ resource "azurerm_cosmosdb_account" "this" {
   automatic_failover_enabled    = true
   key_vault_key_id              = var.customer_managed_key.enabled ? var.customer_managed_key.key_vault_key_id : null
   public_network_access_enabled = var.force_public_network_access_enabled
+  local_authentication_disabled = true
 
   geo_location {
     location          = local.primary_location
--- a/infra/modules/azure_cosmos_account/tests/cosmos.tftest.hcl
+++ b/infra/modules/azure_cosmos_account/tests/cosmos.tftest.hcl
@@ -85,6 +85,11 @@ run "cosmos_is_correct_plan" {
   }
 
   assert {
+    condition     = azurerm_cosmosdb_account.this.local_authentication_disabled == true
+    error_message = "The Cosmos DB account must have local authentication disabled"
+  }
+
+  assert {
     condition     = azurerm_cosmosdb_account.this.consistency_policy[0].consistency_level == "BoundedStaleness"
     error_message = "The Cosmos DB consistency level must be Bounded Staleness"
   }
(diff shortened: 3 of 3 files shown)


.changeset/rotten-taxes-slide.md
```diff
@@ -0,0 +1,5 @@
+---
+"azure_cosmos_account": minor
```
mamu0, line 2: Probably this should be a major

gunzip: question: based on what?

infra/modules/azure_cosmos_account/cosmos.tf
```diff
@@ -7,6 +7,7 @@ resource "azurerm_cosmosdb_account" "this" {
   automatic_failover_enabled    = true
   key_vault_key_id              = var.customer_managed_key.enabled ? var.customer_managed_key.key_vault_key_id : null
   public_network_access_enabled = var.force_public_network_access_enabled
+  local_authentication_disabled = true
```
gunzip, line 10: question: do we have instances configured with local authentication?

.changeset/rotten-taxes-slide.md
mamu0: I’m not sure if key-based local authentication has ever been used with a Cosmos DB created with DX module. If it hasn’t been used, then disabling it won’t cause any breaking changes, so I believe a minor version bump is appropriate. However, if it has been used, then it would indeed be a breaking change.
@pagopa/engineering-team-devex

infra/modules/azure_cosmos_account/cosmos.tf
mamu0: By default, local_authentication_disabled is set to false, so any Cosmos DB created using the DX module can use local authentication. However, I don’t know if any team has actually taken advantage of it.

gunzip:
does this PR prevents accessing cosmosdb though a connection string as well?

.changeset/rotten-taxes-slide.md
Krusty93: We did try to disable local auth, and data from Azure Portal is still accessible. However, roles need to be assigned

mamu0: Here the minor PR #549

mamu0:
> does this PR prevents accessing cosmosdb though a connection string as well?

Yes, this PR prevents accessing Cosmos DB via connection strings, as disabling local authentication disables key-based access according to the [Microsoft documentation](https://learn.microsoft.com/en-us/azure/cosmos-db/nosql/how-to-disable-key-based-authentication?tabs=csharp&pivots=azure-interface-cli)

gunzip:
I recognize the importance of preventing Cosmos DB connections that use connection strings, and that’s what we aim to achieve. However, since this PR is disruptive—given that product teams typically rely on connection strings (to my knowledge)—we should wait for a bulk migration to managed identities before merging it.

.changeset/rotten-taxes-slide.md
mamu0: @pagopa/engineering-team-devex 
PR #549 has been merged! Before making the next change, do you think it makes sense to keep the `authorized_teams` variable **optional**? I was thinking that making it **mandatory** could complicate future developments of the module, since the `azure_role_assignments` module already has the ability to create these roles.
So, in view of potential future use with a stack or a catalog-like tool, it might not be ideal to release a **major** version just to remove this feature.

kin0992: I'd do a two steps update, which seems to be the one proposed by @Krusty93:
- introduce an optional variable (so we guarantee backward compatibility)
- then, after some advertisement of this change, we could set a deadline by which that variable becomes mandatory (bumping to a new major version)

mamu0: Yes, that’s exactly the plan to follow! What I wrote was just a personal reflection on whether to make `authorized_teams` mandatory or not, but regardless, it will be a **major** release because local auth will be disabled

Krusty93: Any update on this?

.changeset/rotten-taxes-slide.md
```diff
@@ -0,0 +1,15 @@
+---
+"azure_cosmos_account": patch
```
Krusty93, line 2:
```suggestion
"azure_cosmos_account": major
```

PR #537, opened 2025-05-07
Make storage account public in CDN basic example

christian-calabrese:


mamu0 (approved):

christian-calabrese:
> why this change? Wouldn't it be better to access the storage account only though the CDN?

Unfortunately, [private link connection from the CDN to the origin](https://learn.microsoft.com/en-us/azure/frontdoor/understanding-pricing#pricing-model-comparison) is available only for the premium SKU, but we use Standard

PR #539, opened 2025-05-07
Add module details in resources tags

christian-calabrese:
This PR introduces consistent tagging for resources created via DX modules by adding DXModuleName and DXModuleVersion. These tags help identify which module and version were used, making it easier to troubleshoot and understand resource origins.

Resolves CES-985


infra/modules/azure_api_management/locals.tf
```diff
@@ -1,4 +1,5 @@
 locals {
+  tags = merge(var.tags, { DXModuleVersion = try(jsondecode(file("${path.module}/package.json")).version, "unknown"), DXModuleName = try(jsondecode(file("${path.module}/package.json")).name, "unknown") })
```
mamu0, line 2: What about creating a tag that indicates the module is from **DX**, without putting “_DX_” in the tag key? For example, we could use `ModuleSource=DX`, and then have `ModuleName` and `ModuleVersion` as the other tags

christian-calabrese: Yes that's a good idea! @pagopa/engineering-team-devex what do you prefer?

Krusty93: Agree with @mamu0

PR #540, opened 2025-05-07
Disable public access to blobs in Storage Account module

mamu0:
This PR adds `allow_nested_items_to_be_public = false` to the Azure Storage Account module to ensure the `allowBlobPublicAccess` property is set to `false`. This change aligns DX module with Technology policies by preventing any blob container or blob from being configured with public access.

This is a security enhancement that ensures all storage accounts managed through this module will have public access to blobs disabled by default, requiring proper authentication for all blob access.

Resolves: CES-837

### DIFF
--- a/.changeset/blue-buckets-carry.md
+++ b/.changeset/blue-buckets-carry.md
@@ -1,5 +1,7 @@
 ---
-"azure_storage_account": minor
+"azure_storage_account": major
 ---
 
-Set `allow_nested_items_to_be_public = false` in Azure Storage Account module to disable public access to blobs
+**BREAKING CHANGE:** 
+
+Public access to blobs within containers is now disabled by default in the Azure Storage Account module. If public blob access is explicitly required, this setting must be overridden with variable `force_public_network_access_enabled`.


.changeset/blue-buckets-carry.md
```diff
@@ -0,0 +1,5 @@
+---
+"azure_storage_account": minor
+---
+
+Set `allow_nested_items_to_be_public = false` in Azure Storage Account module to disable public access to blobs
```
gunzip, line 5:
```suggestion
BREAKING: prevent public acces for Azure Storage Account by default
```

gunzip (changes requested):

infra/modules/azure_storage_account/storage_account.tf
```diff
@@ -1,13 +1,14 @@
...
-  public_network_access_enabled = var.force_public_network_access_enabled
+  name                            = provider::dx::resource_name(merge(local.naming_config, { resource_type = "storage_account" }))
+  resource_group_name             = var.resource_group_name
+  location                        = var.environment.location
+  account_kind                    = "StorageV2"
+  account_tier                    = local.tier_features.account_tier
+  account_replication_type        = local.tier_features.replication_type
+  access_tier                     = var.access_tier
+  public_network_access_enabled   = var.force_public_network_access_enabled
+  allow_nested_items_to_be_public = false
```
gunzip, line 11: suggestion: we must check if someone is using this for some use case

gunzip: For example: websites that do not have the "static website" feature enabled on their Azure Storage account and where content is not stored in the `$web` container, such as `cgnonboardingportal-p-cdnendpoint-storage` and `iopitncgnpeassetsst01`, the `allow_nested_items_to_be_public` setting must be set to `true`.

mamu0: Thanks for the suggestion! I’ve used the existing variable `force_public_network_access_enabled` for `allow_nested_items_to_be_public` as well. This way, if the storage account is public, everything will be configured accordingly, and the same applies if it’s private.

.changeset/blue-buckets-carry.md
```diff
@@ -0,0 +1,5 @@
+---
+"azure_storage_account": minor
```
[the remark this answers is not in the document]
mamu0: Since the value is now configured using a variable, should we also treat this as a `major` change?

infra/modules/azure_storage_account/storage_account.tf
gunzip: Regarding the io-sign use case, consider that when using SAS key you should keep

```
allow_nested_items_to_be_public = false
public_network_access_enabled   = true
```

 https://github.com/pagopa/io-sign/blob/main/infra/resources/prod/storage.tf#L14

mamu0: I understand, this depends on whether we want to enforce a stricter standard or keep things more configurable.
We can go ahead with a new variable as you suggested, I don’t have a strong opinion on this, both are valid options

mamu0: Also, in the specific case of `io-sign`, they’re not using this module, so they wouldn’t be affected

.changeset/blue-buckets-carry.md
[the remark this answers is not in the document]
gunzip: no if the variable is optional and the default is the same as the previous version. yes othwerwise.

infra/modules/azure_storage_account/storage_account.tf
Krusty93: I would discourage the use of SAS tokens (and keys in general) when possible, so I'd stick with passwordless authentication as the only available method. In case of `io-sign`, it might be a 20% case where they need to use the resource instead of the module

gunzip:
idea: what about adding into readme that this module must not be used to feed a CDN?

Krusty93:
> idea: what about adding into readme that this module must not be used to feed a CDN?

Agree, with a link to the CDN module

mamu0:
> idea: what about adding into readme that this module must not be used to feed a CDN?

I added this info to the README @gunzip @Krusty93

gunzip (approved):

PR #549, opened 2025-05-09
Add Cosmos DB SQL role assignment capability to module

mamu0:
This PR adds the capability to assign SQL roles to principals in the Azure Cosmos DB module. 

The change allows users to provide a list of principal IDs that will be automatically assigned the "Cosmos DB Built-in Data Reader" role on a specified database scope. 

Resolves: CES-990


infra/modules/azure_cosmos_account/variables.tf
```diff
@@ -147,3 +147,15 @@ variable "alerts" {
     error_message = "When alerts are enabled, all thresholds must be set."
   }
 }
+
+variable "authorized_principal_ids" {
```
gunzip, line 151: shoulnd't we do this using the [role assignement module](https://github.com/pagopa/dx/tree/main/infra/modules/azure_role_assignments)?

infra/modules/azure_cosmos_account/locals.tf
```diff
@@ -43,4 +43,12 @@ locals {
...
 
   private_endpoint_name = provider::dx::resource_name(merge(local.naming_config, { resource_type = "cosmos_private_endpoint" }))
+
+  # "Cosmos DB Built-in Data Reader/Cotributor" IDs
+  cosmos_db_data_reader_role_id      = "${azurerm_cosmosdb_account.this.id}/sqlRoleDefinitions/00000000-0000-0000-0000-000000000001"
+  cosmos_db_data_contributor_role_id = "${azurerm_cosmosdb_account.this.id}/sqlRoleDefinitions/00000000-0000-0000-0000-000000000002"
+
+  # Convert the list of authorized principal IDs into a map with static keys
+  reader_principal_ids_map = { for i, pid in var.authorized_principal_ids.readers : i => pid }
+  writer_principal_ids_map = { for i, pid in var.authorized_principal_ids.writers : i => pid }
```
christian-calabrese, line 53: Have you tried this on a new identity created in the same plan? The terraform error about ids not available until apply is always around the angle 🥲

infra/modules/azure_cosmos_account/variables.tf
mamu0: Yes, that module assigns those two roles, but since it’s not mandatory, if PR #536 is applied and the users of this Cosmos DB don’t have the roles, they won’t be able to access the database. The discussion where we talked about it is [here](https://github.com/pagopa/dx/pull/536#discussion_r2079394436)

gunzip: Isn't this variable optional as well? What's the difference?

mamu0: @Krusty93  already mentioned this in the discussion I linked above.
To sum it up, this `minor` adds the variable just to give teams some time to integrate it into their projects.
Then, in the PR #536, the same variable will be updated to make it mandatory, ensuring a smooth transition.

infra/modules/azure_cosmos_account/roles.tf
```diff
@@ -0,0 +1,21 @@
...
+resource "azurerm_cosmosdb_sql_role_assignment" "principal_role_assignments_writer" {
+  for_each = local.writer_principal_ids_map
+
+  resource_group_name = var.resource_group_name
+  account_name        = azurerm_cosmosdb_account.this.name
+  role_definition_id  = local.cosmos_db_data_contributor_role_id
+  principal_id        = each.value
+
+  scope = azurerm_cosmosdb_account.this.id
+}
```
lucacavallaro, line 21: I don't like this choice to make cosmos dependent on its consumers

I would move this `azurerm_cosmosdb_sql_role_assignment` resources into compute modules.

infra/modules/azure_cosmos_account/roles.tf
```diff
@@ -0,0 +1,21 @@
+resource "azurerm_cosmosdb_sql_role_assignment" "principal_role_assignments_reader" {
+  for_each = local.reader_principal_ids_map
+
+  resource_group_name = var.resource_group_name
+  account_name        = azurerm_cosmosdb_account.this.name
+  role_definition_id  = local.cosmos_db_data_reader_role_id
+  principal_id        = each.value
+
+  scope = azurerm_cosmosdb_account.this.id
```
Krusty93, line 9: **question**: is this id correct? Shouldn't be `/dbs/*` only?

mamu0: Maybe you are right, i've to test it, thank you

mamu0: i've done some tests, and it’s not possible to assign a scope that ends with `*`, it expects a scope in the full ID format. By assigning the Cosmos DB account ID as the scope, it should automatically include all databases within it

Krusty93: I believe [you are right](https://github.com/pagopa/dx/blob/e676023817e740fe6ee965ecc6747801df1f1990/infra/modules/azure_role_assignments/modules/cosmos/locals.tf#L8)

Krusty93 (approved):

PR #551, opened 2025-05-09
Introduce the Service Bus Namespace module

Krusty93:
New module to provision Service Bus Namespace using Premium and Standard SKUs only. It sticks to 1 partition and 1 message units to avoid unnecessary complexity. They can be added in future without breaking changes

Code includes docs, examples and tests.

Close CES-997

### DIFF
--- a/infra/modules/azure_service_bus_namespace/README.md
+++ b/infra/modules/azure_service_bus_namespace/README.md
@@ -17,7 +17,7 @@ This Terraform module deploys an Azure Service Bus namespace. It supports `Stand
 | m    | Low-load production environments and basic features usage    | Despite a firewall, it is publicly available on internet. No VNet integration |
 | l    | High-load production environments and all features available | Access via Private Endpoints only                                             |
 
-**WARNING**: It is strongly encouraged to use `Premium` SKU (default) due to security concerns of `Standard` SKU.
+**WARNING**: It is strongly encouraged to use the `Premium` SKU (default) due to the mentioned security concerns of the `Standard` SKU.
 
 ## Usage Example
 


infra/modules/azure_service_bus_namespace/README.md
```diff
@@ -0,0 +1,91 @@
...
+- **Private Endpoint Integration**: (`Premium` SKU only) Creates a private DNS A record for the container app, enabling secure internal communication.
+
+## Tiers and Configurations
+
+| Tier | Description                                                  | Security                                                                      |
+|------|--------------------------------------------------------------|-------------------------------------------------------------------------------|
+| m    | Low-load production environments and basic features usage    | Despite a firewall, it is publicly available on internet. No VNet integration |
+| l    | High-load production environments and all features available | Access via Private Endpoints only                                             |
+
+**WARNING**: It is strongly encouraged to use `Premium` SKU (default) due to security concerns of `Standard` SKU.
```
gunzip, line 20: **suggestion**: can we link a reference to clarify what "security concerns" means? to give users an hints to decide about the trade off.

infra/modules/azure_service_bus_namespace/locals.tf
```diff
@@ -0,0 +1,30 @@
...
+      "m" = "Standard",
+      "l" = "Premium"
+    },
+    var.tier,
+    "Premium"
+  )
+
+  capacity       = local.sku_name == "Premium" ? 1 : 0
+  partitions     = local.sku_name == "Premium" ? 1 : 0
+  default_action = local.sku_name == "Premium" ? "Allow" : "Deny" # Using "Deny" for Premium SKU breaks the provider validation
```
gunzip, line 25: question: is this a TODO? in case, I'd link the related Jira issue. is there any issue on the provider github repo?

gunzip (commented):
very good!

infra/modules/azure_service_bus_namespace/locals.tf
Krusty93: No, it is only a provider weird behavior

infra/modules/azure_service_bus_namespace/README.md
Krusty93: I am referring to the `Security` column in the previous table
Would you like to make it more explicit?

gunzip, line 20:
```suggestion
**WARNING**: It is strongly encouraged to use the `Premium` SKU (default) due to the mentioned security concerns of the `Standard` SKU.
```

gunzip (approved):

infra/modules/azure_service_bus_namespace/locals.tf
Krusty93: I don't know if it could be filed as bug or not. When you set private endpoints replacing the firewall (available in `Premium` SKU only), the `default_action` variable is useless as the firewall is disabled.
However, provider doesn't allow to set `Deny`, probably because the firewall is actually disabled. More tests would be useful to understand the behavior of the control plane's APIs behind the provider. But honestly, I don't think it does worth doing it though

PR #557, opened 2025-05-12
Remove caching in terraform apply workflow

christian-calabrese:


christian-calabrese:
The terraform plan GitHub Action currently does not cache the .terraform folder, while the terraform apply action does. This inconsistency can lead to mismatches in the versions of Terraform modules used between the two steps.

For instance, in [this workflow run](https://github.com/pagopa/io-wallet/actions/runs/14972791655), terraform plan used version 0.3.1 of the Function App module, whereas terraform apply reused the cached version 0.2.9.

Thanks to the newly introduced ModuleVersion tag, we were able to identify this long-standing issue. This PR addresses the problem by removing the caching step that causes additional issues than saved time (around 20 seconds in the cited workflow).

.github/workflows/infra_apply.yaml
```diff
@@ -155,14 +155,6 @@ jobs:
           bundle_name: tfplan
           file_path: ${{ needs.tf_plan.outputs.working_dir }}
 
-      - name: Retrieve Terraform Modules from Cache
```
gunzip, line 158: question: what if we have a plan with some module version and an apply with another?

christian-calabrese: This is the issue we are solving by removing caching. Do you think there may be nitpicks that we didn't consider?
([More info here](https://github.com/pagopa/dx/pull/557#issuecomment-2872788824))

gunzip (approved):

PR #558, opened 2025-05-12
Add Terraform Prod Config

Krusty93:
This PR introduces the Terraform configuration for production environments.

As already mentioned in some RFCs, we are starting to feel the need to prepare resources as a prerequisite for DX adoption by PagoPA products. The preparation of these resources, such as custom roles or policies, has so far been done ad-hoc in the IO repositories. However, since we are expanding, it is good to centralize all the necessary resources so as to support new products more quickly. It starts with adding the `PagoPA Opex Dashboards Contributor` custom role, currently used by IO only and defined in `io-infra` repository (should be removed though)

Close CES-1002

Depends on https://github.com/pagopa/eng-azure-governance/pull/176


gunzip (changes requested):
let's talk about this a little bit more before going on :-)

infra/resources/prod/main.tf
```diff
@@ -0,0 +1,50 @@
...
+
+provider "azurerm" {
+  alias = "prod-io"
+  features {
+  }
+  subscription_id     = local.prod-io
+  storage_use_azuread = true
+}
+
+resource "azurerm_role_definition" "pagopa_opex_contributor" {
```
gunzip, line 32: idea: can we enable the custom role for every subscription? (using a wildcard scope)

infra/resources/prod/main.tf
```diff
@@ -0,0 +1,50 @@
...
+  name        = "PagoPA Opex Dashboards Contributor"
```
gunzip, line 33:
```suggestion
  name        = "DX Opex Dashboards Contributor"
```

infra/resources/prod/main.tf
Krusty93: Unfortunately we don't have enough privileges to write roles at management group level

PR #559, opened 2025-05-13
Add Service Bus support to Role Assignment modules

Krusty93:
Close CES-1005


infra/modules/azure_role_assignments/variables.tf
```diff
@@ -128,3 +128,17 @@ variable "apim" {
 
   default = []
 }
+
+variable "service_bus" {
+  description = "A list of role assignments for Azure Service Bus, specifying the namespace, resource group, role, and queue names or subscriptions with their topics."
```
gunzip, line 133: suggestion: I'd try to be more explicit. For example:

1. what's the `description` field about ?
2. what do keys and values in the `subscriptions` field represent?
3. which `roles` are supported?

gunzip: What about renaming the field to something like 'role_description'? It's not immediate for the casual reader to correlate `description` with the role since it appears inside the `service_bus` block.

infra/modules/azure_role_assignments/variables.tf
```diff
@@ -128,3 +128,27 @@ variable "apim" {
 
   default = []
 }
+
+variable "service_bus" {
+  description = <<EOT
+  A list of role assignments for Azure Service Bus, specifying the namespace, resource group, and role.
+  For queues and topics, list the names. For subscriptions, pair the related topic and the subscription in a map object.
```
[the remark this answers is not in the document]
Krusty93: Good catch, probably we need to switch to a `map(list(string))`

infra/modules/azure_role_assignments/variables.tf
Krusty93: We have the same field name in all variables of the module, I suggest to do not breaking the coherence for a single resource type

infra/modules/azure_role_assignments/variables.tf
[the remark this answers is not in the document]
Krusty93: Added: https://github.com/pagopa/dx/pull/559/commits/65bc4fc04fb7e3dfa1c620db9fcaafc92fa19c65

gunzip (approved):

mamu0 (approved):
LGTM

PR #566, opened 2025-05-16
Enhance keep-alive action

mamu0:
This PR improves the repository keep-alive action to properly handle protected branches and repository access restrictions.

**Why:**
Our repository keep-alive action was failing with 403 errors in repositories with protected branches or restricted access. This change ensures that repositories can be kept active without manual intervention, preventing GitHub from automatically disabling scheduled workflows after 60 days of inactivity.

Resolves: CES-991


.github/workflows/_cron-bash-keep-alive-repository copy.yaml
```diff
@@ -8,6 +8,12 @@ name: Keep Alive
 
 on:
   workflow_dispatch:
+    inputs:
+      bypass:
+        description: "Bypass the 55 days check"
```
gunzip, line 13:
```suggestion
        description: "Perform the action even if the last commit was less than 55 days ago."
```

gunzip (approved):

PR #574, opened 2025-05-21
Revert TLS from 1.3 to 1.2 for App Service and Function App modules

mamu0:
This pull request reverts the TLS protocol version from 1.3 back to 1.2 for our App Service and Function App infrastructure modules. The change was necessary due to compatibility issues discovered with certain client applications that don't fully support TLS 1.3 yet.

Added new `tls_version` variable to allow flexible selection of TLS version.

Resolves: CES-1024

### DIFF
--- a/.changeset/plain-bees-sit.md
+++ b/.changeset/plain-bees-sit.md
@@ -1,8 +1,8 @@
 ---
-"azure_function_app_exposed": minor
-"azure_app_service_exposed": minor
-"azure_function_app": minor
-"azure_app_service": minor
+"azure_function_app_exposed": patch
+"azure_app_service_exposed": patch
+"azure_function_app": patch
+"azure_app_service": patch
 ---
 
 Revert from TLS 1.3 to TLS 1.2 for app service and function app modules, added new variable named tls_version with default 1.2
--- a/.changeset/plain-bees-sit.md
+++ b/.changeset/plain-bees-sit.md
@@ -5,4 +5,6 @@
 "azure_app_service": patch
 ---
 
-Revert from TLS 1.3 to TLS 1.2 for app service and function app modules, added new variable named tls_version with default 1.2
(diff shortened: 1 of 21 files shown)


.changeset/plain-bees-sit.md
```diff
@@ -0,0 +1,8 @@
+---
+"azure_function_app_exposed": minor
```
gunzip, line 2: **question**: why not patch? I'd expect this fixes the latest minor

.changeset/plain-bees-sit.md
```diff
@@ -0,0 +1,8 @@
...
+"azure_app_service_exposed": minor
+"azure_function_app": minor
+"azure_app_service": minor
+---
+
+Revert from TLS 1.3 to TLS 1.2 for app service and function app modules, added new variable named tls_version with default 1.2
```
gunzip, line 8: suggestion: I'd add a link to some reference to explain why this is needed now

.changeset/plain-bees-sit.md
mamu0: because there is a new variable, but since the latest version is not used much I suppose they could also make a patch

kin0992: I agree with `patch` too, so we do not have a bugged minor version if somebody references it with `~> X.Y`.
Moreover, it is true that there is a new variable, but it has a default, so it doesn't change the behaviour of the users of these modules

kin0992 (approved):
Besides the comments of @gunzip, which I agree with those, looks good to me

.changeset/plain-bees-sit.md
gunzip: if it's not mandatory it could be a patch

mamu0: Do you have any specific references I could include?

gunzip: https://learn.microsoft.com/en-us/answers/questions/2125989/azure-api-management-throwing-error-in-connecting

PR #582, opened 2025-05-26
Add support to Azure Function with programming model v3

kin0992:
To ensure the legacy Azure Function programming model, requests are traced, we need to wrap the handler function in an OpenTelemetry wrapper that manipulates the context to have correlated requests on Application Insights.

Closes CES-1016

### DIFF
--- a/.changeset/great-tips-jump.md
+++ b/.changeset/great-tips-jump.md
@@ -1,0 +1,5 @@
+---
+"@pagopa/azure-tracing": minor
+---
+
+Enhance support for legacy Azure Function (v3) to ensure the requests are properly correlated.
--- a/.changeset/great-tips-jump.md
+++ b/.changeset/great-tips-jump.md
@@ -3,3 +3,21 @@
 ---
 
 Enhance support for legacy Azure Function (v3) to ensure the requests are properly correlated.
+
+## Usage
+
+For legacy Azure Functions, you can use the wrap the Azure Function within the `withOtelContextFunctionV3` to make sure the requests are properly correlated.
+
+```typescript
+import { AzureFunction, Context as FunctionContext } from "@azure/functions"; // "@azure/functions": "^3"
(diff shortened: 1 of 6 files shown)


packages/azure-tracing/README.md
```diff
@@ -68,6 +68,29 @@ registerAzureFunctionHooks(app);
 ...
 ```
 
+### Instrumenting Azure Functions v3 Handlers
+
+For Azure Functions using the v3 programming model, you can use the `withOtelContextFunctionV3` helper function to wrap your handlers and ensure OpenTelemetry context propagation. This function works with the v3 `Context` object structure.
```
gunzip, line 73: question: do the same consideration for ESM modules apply? (ie. you still have to set `NODE_OPTIONS` in case)

kin0992: yes, that is the suggested way to integrate the tracing package to have full correlation between requests.

Should I explicit it in another way?

.changeset/great-tips-jump.md
```diff
@@ -0,0 +1,23 @@
...
+
+## Usage
+
+For legacy Azure Functions, you can use the wrap the Azure Function within the `withOtelContextFunctionV3` to make sure the requests are properly correlated.
+
+```typescript
+import { AzureFunction, Context as FunctionContext } from "@azure/functions"; // "@azure/functions": "^3"
+import createAzureFunctionHandler from "@pagopa/express-azure-functions/dist/src/createAzureFunctionsHandler.js";
+
+import { withOtelContextFunctionV3 } from "@pagopa/azure-tracing/azure-functions/v3"; // from version ^0.4.0
```
kin0992, line 15:
```suggestion
import { withOtelContextFunctionV3 } from "@pagopa/azure-tracing/azure-functions/v3"; // "@pagopa/azure-tracing": "^0.4"
```

PR #384, opened 2025-03-20
Module to deploy self hosted runner on AWS CodeBuild

christian-calabrese:
Resolves #CES-879

PR #583, opened 2025-05-26
Implementation of the bootstrap module for AWS projects

christian-calabrese:
Resolves CES-754
Depends on #384


infra/modules/aws_github_environment_bootstrap/iam_roles_app.tf
```diff
@@ -0,0 +1,31 @@
...
+    domain        = var.environment.domain
+    name          = "app-github-cd"
+    resource_type = "iam_role"
+  }))
+  description = "Role to assume to run CD for applications via GitHub Actions"
+
+  assume_role_policy = data.aws_iam_policy_document.github_assume_role_policy.json
+}
+
+resource "aws_iam_role_policy_attachment" "app_cd_admin" {
```
mamu0, line 28: **question**: At the moment, I see that both the `app` and `infra` have the same policies assigned, so they are currently identical. It’s not urgent, but would it make sense to assign them different privileges? I think the `infra` setup is already fine, but for the `app`, we could possibly reduce access, especially for the `cd`, by limiting it to just compute-related services.

christian-calabrese: Absolutely right! Shall we just include Lambda and ECS?

christian-calabrese: Done!

mamu0: Yes, Lambda and ECS should be enough!

infra/modules/aws_github_environment_bootstrap/README.md
```diff
@@ -0,0 +1,218 @@
...
+    runs-on: ubuntu-latest
+    environment: infra-cd
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v4
+
+      - name: Configure AWS credentials
+        uses: aws-actions/configure-aws-credentials@v4
+        with:
+          role-to-assume: ${{ secrets.INFRA_CD_ROLE_ARN }}
```
Krusty93, line 53: Why `INFRA_CD_ROLE_ARN`? Should not be `ROLE_ARN` as it has already the environment boundaries?

PR #592, opened 2025-05-29
Improve Terraform Drift Detection workflow template documentation

Krusty93:
Close CES-1046


gunzip (approved):
do we need to add the required permission in the calling workflow example?

Krusty93:
> do we need to add the required permission in the calling workflow example?

Wondered the same. Don't know, they are already specified internally. If we need to change them in the future, the risk is to raise the need to make a breaking release because of them. I don't think it does worth (at the moment)

PR #594, opened 2025-05-29
Bootstrap devex prod subscription

christian-calabrese:
Resolves CES-1042
Depends on #664

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -32,7 +32,8 @@ repos:
           - --hook-config=--create-file-if-not-exist=true
         exclude: |
           (?x)^(
-            infra/modules/.*?
+            infra/modules/.*?|
+            providers/.*?
             )$
       - id: terraform_tflint
         args:


infra/core/_modules/core_infra/network.tf
```diff
@@ -0,0 +1,24 @@
+module "cidrs" {
```
gunzip, line 1: question: what's the difference between using this and out provider method?

infra/core/prod/main.tf
```diff
@@ -0,0 +1,7 @@
+module "core" {
+  source = "../_modules/core_infra"
+
+  environment          = local.environment
+  virtual_network_cidr = "10.52.0.0/16"
```
gunzip, line 5: question: why it's hardcoded?

infra/modules/azure_github_environment_bootstrap/github_environments_cd.tf
```diff
@@ -1,5 +1,5 @@
-resource "github_repository_environment" "infra_prod_cd" {
-  environment = "infra-prod-cd"
+resource "github_repository_environment" "infra_cd" {
```
gunzip, line 1: question: are these breaking changes?

infra/modules/azure_github_environment_bootstrap/github_environments_ci.tf
```diff
@@ -1,5 +1,5 @@
-resource "github_repository_environment" "infra_prod_ci" {
-  environment = "infra-prod-ci"
+resource "github_repository_environment" "infra_ci" {
```
gunzip, line 1: question: are these breaking changes?

infra/modules/github_environment_bootstrap/github_repository.tf
```diff
@@ -5,6 +5,8 @@ resource "github_repository" "this" {
```
gunzip, line 5: question: are these breaking changes?

infra/core/_modules/core_infra/network.tf
christian-calabrese: This is a static module that does not check the availability of the CIDR. Therefore, it doesn't suggest you the first available space.

On the other hand, our provider method needs to know the ID of the VNet to do so.
Since the VNet is new and completely empty, the "cidrs" module is better suitable

infra/modules/azure_github_environment_bootstrap/github_environments_ci.tf
christian-calabrese: We added the moved blocks to avoid causing a breaking

PR #601, opened 2025-06-03
Upgraded modules lock pre-commit script

christian-calabrese:
Fix a bug causing inconsistent locking due to differences in the version of the `sort` command
Implement extensive logging in JSON format
Enhance lock files adding the module's version and name

Resolves CES-1055

### DIFF
--- a/infra/scripts/lock-modules.sh
+++ b/infra/scripts/lock-modules.sh
@@ -152,10 +152,11 @@ function get_module_registry_url() {
   local version=$(echo "$module_info" | jq -r '.Version // "latest"')
 
   # Check if it's a registry module
-  if [[ "$source" == registry.terraform.io/* ]]; then
-    # Convert to URL
-    local url="https://registry.terraform.io/modules/${source#registry.terraform.io/}/$version"
-    echo "$url"
+  if [[ "$source" == $REGISTRY_URL/* ]]; then
+    # Convert to terraform registry URL
+    local base_url="https://${REGISTRY_URL}/modules"
+    local module_path="${source#${REGISTRY_URL}/}"
+    echo "${base_url}/${module_path}/${version}"
   else
     echo "Not a registry module: $source"
     return 1
--- a/infra/scripts/lock-modules.sh
+++ b/infra/scripts/lock-modules.sh
@@ -15,8 +15,6 @@ if [[ "${PRECOMMIT_DEBUG:-0}" -eq 1 ]]; then
     set -x
 fi
 
-# Initialize JSON results array
-declare -a json_results=()
 declare  module_results='[]'
 
 # Logging functions for different severity levels
@@ -87,8 +85,6 @@ function add_module_result() {
             "prevhash": $prevhash
         }')
 
-    # Add to stdout JSON output
-    json_results+=("$json_entry")
     echo "$json_entry" >> "$JSON_OUTPUT_FILE"
 }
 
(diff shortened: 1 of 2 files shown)


infra/scripts/lock-modules.sh
```diff
@@ -42,26 +47,97 @@ function info() {
...
+            "name": $modulefoldername,
+            "status": $status,
+            "version": $version,
+            "message": $message,
+            "hash": $hash,
+            "prevhash": $prevhash
+        }')
+
+    # Add to stdout JSON output
+    json_results+=("$json_entry")
```
gunzip, line 91: question: where the `json_results` is used?

infra/scripts/lock-modules.sh
```diff
@@ -95,43 +175,68 @@ function process_module() {
     local -r module_name=$(basename "$module_path")
```
gunzip, line 175: question: what if we have two modules with the same basename but different orgs?

infra/scripts/lock-modules.sh
christian-calabrese: It was used to also print logs in JSON format in a first iteration. I'll remove it!

infra/scripts/lock-modules.sh
christian-calabrese: I'm not sure about what you mean by orgs, but the lock file is placed in each root module (e.g. src/domains/ioweb-common in io-infra), so entries are always decoupled since you can't have duplicated module names.
Or you mean the registry namespace?

gunzip: Yep, I mean the registry namespace.

christian-calabrese: Nice addition indeed! I just added the url of the module in the lock file
e.g.: registry.terraform.io/pagopa-dx/azure-storage-account/azurerm

christian-calabrese: Amazingly enough the url is not actually working, but that's what you can get from the .terraform/modules/modules.json file created by tearraform

I'll wrangle it a little bit to make it work

infra/scripts/lock-modules.sh
```diff
@@ -2,18 +2,21 @@
...
+readonly SCRIPT_NAME=$(basename "$0")                          # Get the script name for logging
+readonly JSON_OUTPUT_FILE="${LOCK_MODULES_JSON_OUTPUT_FILE:-}" # Optional file path for JSON output. Set it to the name or path of the JSON file to produce (e.g. lock_output.json)
 
 # Enable debug mode if PRECOMMIT_DEBUG environment variable is set to 1
 # set -x enables printing of each command before execution
 if [[ "${PRECOMMIT_DEBUG:-0}" -eq 1 ]]; then
     set -x
 fi
 
+declare  module_results='[]'
```
mamu0, line 18: It seems that this variable is never updated

mamu0 (approved):
Nice work! :shipit:

infra/scripts/lock-modules.sh
```diff
@@ -42,26 +45,126 @@ function info() {
...
+
+  # Extract Source and Version
+  local source=$(echo "$module_info" | jq -r '.Source')
+  local version=$(echo "$module_info" | jq -r '.Version // "latest"')
+
+  # Check if it's a registry module
+  if [[ "$source" == registry.terraform.io/* ]]; then
+    # Convert to URL
+    local url="https://registry.terraform.io/modules/${source#registry.terraform.io/}/$version"
+    echo "$url"
```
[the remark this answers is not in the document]
christian-calabrese: Good enhancement, pushed!

PR #602, opened 2025-06-05
Generate provenance attestation when publishing npm packages

kin0992:
Closes CES-1058


.github/workflows/_release-changesets-pull-request.yaml
```diff
@@ -37,3 +38,4 @@ jobs:
         env:
           GITHUB_TOKEN: ${{ secrets.BOT_GH_TOKEN }}
           NPM_TOKEN: ${{ secrets.NPM_TOKEN }}
+          NPM_CONFIG_PROVENANCE: true
```
kin0992, line 41: **note**: [as documented here](https://docs.npmjs.com/generating-provenance-statements/#using-third-party-package-publishing-tools), this should be enough

PR #604, opened 2025-06-05
Fix leaky abstraction for Storage Account `access_tier` variable

mamu0:
This PR introduces new values for the `access_tier` variable to explicitly define the access tier for the storage account. The change improves clarity and flexibility in workload-specific configurations.

**Minor Change:**
Teams should update the module configuration to include the new values, and they should update the version.

Resolves: CES-943

### DIFF
--- a/infra/modules/azure_storage_account/README.md
+++ b/infra/modules/azure_storage_account/README.md
@@ -23,12 +23,12 @@ This Terraform module provisions an Azure Storage Account with optional configur
 
 ## Access Tier Comparison
 
-| Access Tier       | Description                                               | Retrieval Speed | Cost Efficiency | Use Case            |
-|-------------------|-----------------------------------------------------------|-----------------|-----------------|---------------------|
-| `frequent`        | Frequent access with immediate retrieval.                 | High            | Moderate        | Real-time workloads |
-| `infrequent`      | Infrequent access with slightly delayed retrieval.        | Moderate        | High            | Analytical data     |
-| `rare`            | Rare access with immediate retrieval.                     | Low             | Very High       | Archival storage    |
-| `performance`     | SSD-based high-speed access for demanding workloads.      | Very High       | Low             | High-performance    |
+| Access Tier       | Description                                               | Storage Costs | Access Costs |  Use Case                                                            |
+|-------------------|-----------------------------------------------------------|---------------|--------------|----------------------------------------------------------------------|
+| `frequent`        | Frequent access with immediate retrieval.                 | High          | Low          | Real-time workloads, data for web/mobile apps                        |
+| `infrequent`      | Infrequent access with slightly delayed retrieval.        | Low           | High         | Short-term backups, Disaster Recovery, less frequently accessed logs |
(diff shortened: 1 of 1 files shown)


.changeset/odd-tools-invite.md
```diff
@@ -0,0 +1,31 @@
...
+
+# BREAKING CHANGE
+
+The breaking change introduces a new variable `use_case` to define the access tier for the storage account.
+
+The change was made to improve clarity and flexibility in configuring storage account access tiers. By explicitly setting the use_case variable, consumers can better align their configurations with specific workload requirements.
+
+To update your code, use an higher version and add the `use_case` variable to your module configuration instead of old one `access_tier`. Choose one of the following valid options:
+
+- realtime → frequent access (Hot)
```
gunzip, line 13: suggestion: if we know in advance some use case where these tiers arealready used I'd add some examples (ie. "users file, authentication logs, forensic data, ...")

mamu0: Thanks for the feedback! it makes sense, and I agree with you.
This PR is meant as a starting point to introduce the concept, so we can rename use_case values, and also variable name if can be confusing.
Since access tiers may support multiple use cases, it's tricky to summarize that in a single word 🤔

mamu0: Maybe we can change them with something like

- `frequent_access` -> Hot
- `infrequent_access` -> Cool
- `rare_access` -> Cold
- `performance_critical` -> Premium

And return to `access_tier` variable name

gunzip: maybe access_tier = frequent | infrequent | rare | immediate ?

gunzip: it's ok `performance` or `priority`

gunzip (commented):
suggestion: `performance` -> `performant` (it aligns with `frequent`, ...)

mamu0:
> suggestion: `performance` -> `performant` (it aligns with `frequent`, ...)

I see your point, but I think "performance" fits better here since it's a standard noun like the other options. "Performant" isn't commonly used in English and might sound a bit off.
if we're looking for alternatives, maybe something like "high_performance", "fast", or "premium" (as the base terraform resource) could work too.

infra/modules/azure_storage_account/locals.tf
```diff
@@ -28,6 +28,20 @@ locals {
 
   tier_features = local.tiers[var.tier]
 
+  access_tier_map = {
+    frequent    = "Hot"
+    infrequent  = "Cool"
+    rare        = "Cold"
+    performance = "Premium"
```
Krusty93, line 35: **suggestion**: I'm afraid the majority of consumer will chose `performance`, but it makes sense only for file share (and nobody uses them). I would remove it then

mamu0: at the moment it is already possible to choose that tier, but with the official name which is `Premium`, do we want to remove it anyway?

mamu0:
I'm closing this PR for now as we're currently having broader discussions about how to best manage tiering. The outcomes of those discussions will lead to a new PR with the appropriate changes and an updated approach.

Thanks for your understanding!

PR #614, opened 2025-06-09
Add packages write permissions to GitHub workflow

Krusty93:
Close CES-1068

### DIFF
--- a/infra/modules/github_environment_bootstrap/workflow_permissions.tf
+++ b/infra/modules/github_environment_bootstrap/workflow_permissions.tf
@@ -1,0 +1,19 @@
+resource "null_resource" "workflow_permissions" {
+  triggers = {
+    repo = github_repository.this.full_name
+  }
+
+  provisioner "local-exec" {
+    when    = create
+    command = <<EOT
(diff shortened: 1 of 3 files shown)


infra/modules/github_environment_bootstrap/workflow_permissions.tf
```diff
@@ -0,0 +1,19 @@
+resource "null_resource" "workflow_permissions" {
+  triggers = {
+    repo = github_repository.this.full_name
+  }
+
+  provisioner "local-exec" {
+    when    = create
```
mamu0, line 7: **suggestion:** I'm not sure if it makes sense in this specific case (since these settings might already be the default), but you could consider adding another `local-exec provisioner` for the **destroy** phase to revert this operation. E.g.:
```hcl
  provisioner "local-exec" {
    when    = destroy
    command = <<EOT
      gh api --method=PUT "repos/${self.triggers.repo}/actions/permissions/workflow" \
        -F can_approve_pull_request_reviews=false \
        -F default_workflow_permissions="read"
    EOT
  }
```

mamu0 (approved):

PR #619, opened 2025-06-10
Add module to simplify the alert setup for Service Bus DLQs

Krusty93:
A new Terraform module to simplify the provisioning of Azure monitor alerts over Service Bus DLQs.

The Service Bus Namespace module was not used because this code will be managed by the owner of the entity set as target, while the namespace is potentially shared among multiple teams

Close CES-1043


.changeset/fluffy-cases-greet.md
```diff
@@ -0,0 +1,5 @@
+---
+"azure_service_bus_alerts": patch
```
gunzip, line 2:
```suggestion
"azure_service_bus_alerts": minor
```

infra/modules/azure_service_bus_alerts/variables.tf
```diff
@@ -0,0 +1,108 @@
...
+    severity        = optional(string, "Warning")
+    auto_mitigate   = optional(bool, true)
+    check_every     = optional(string, "PT15M")
+    lookback_period = optional(string, "PT30M")
+    enable          = optional(bool, true)
+  })
+
+  default     = null
+  description = <<EOF
+"Configure alert over messages moved in dead-letter queue. If set to null, no alert will be created
```
gunzip, line 28: question: what's the diffence vs `alerts_on_dlq_messages` ?

gunzip: It's not clear from the description (which is the same) when to use one vs the other.

Krusty93: Updated

gunzip (approved):

infra/modules/azure_service_bus_alerts/example/simple/main.tf
```diff
@@ -0,0 +1,60 @@
...
+
+module "sbns_alert" {
+  source = "../../"
+
+  environment = local.environment
+
+  alerts_on_active_messages = {
+    entity_names = [
+      azurerm_servicebus_queue.example.name,
+      azurerm_servicebus_topic.example.name
```
giamma1295, line 50: When using subscriptions, DLQs are tied to the subscription, could there be a case where a message is correctly processed in one subscription but fail in another subscription(so present in its DLQ), in this case how would the module be configured, passing the subscription's name instead of the topic one?

Krusty93: Unfortunately, the metric at subscription level is not exported, so topic-level should be an aggregated of all DLQs under the same topic.
In summary, a custom tool such as a Function App is needed to monitor DLQs at subscription level

I think I already have somewhere in my private repos some code to do that

PR #627, opened 2025-06-11
Add GitHub Runner image and publish workflow

mamu0:
This PR adds Dockerfile for a GitHub self-hosted runner image and a workflow to build and publish it, and performs a minor update to the Terraform module by setting a default value for the `container_app_environment` variable to enable image selection.

Resolves: CES-992


gunzip:
What's the user's advantage against the former runner?

.github/workflows/_release-docker-self-hosted-runner.yaml
```diff
@@ -0,0 +1,128 @@
...
+  schedule:
+    - cron: '0 8 1 */2 *' # Every 2 months on the 1st day at 08:00 UTC
+  push:
+    branches:
+      - main
+    paths:
+      - 'self-hosted-runner/**'
+
+jobs:
+  found_sha:
```
Krusty93, line 26: **question**: Why two jobs instead of a single one?

self-hosted-runner/entrypoint.sh
```diff
@@ -0,0 +1,75 @@
...
+printf "ℹ️ Runner Name: %s\n\t" "$RUNNER_NAME"
+printf "ℹ️ Working Directory: %s\n\t" "$WORK_DIR"
+
+if [ ! -f ".runner" ]; then
+  echo "Configuring the runner..."
+  ./config.sh \
+    --url "$REPO_URL" \
+    --token "$RUNNER_TOKEN" \
+    --name "$RUNNER_NAME" \
+    --work "$WORK_DIR" \
```
Krusty93, line 58: **question**: What's supposed to do?

self-hosted-runner/entrypoint.sh
```diff
@@ -0,0 +1,75 @@
...
+      -H 'X-GitHub-Api-Version: 2022-11-28' \
+      "$REGISTRATION_TOKEN_API_URL" \
+      | jq -r '.token')"
+    
+    if [[ -z "$RUNNER_TOKEN" || "$RUNNER_TOKEN" == "null" ]]; then
+      echo "ERROR: Failed to generate RUNNER_TOKEN"
+      exit 1
+    fi
+
+    export RUNNER_TOKEN
```
Krusty93, line 27: **question**: do you think other processes might need to access these variables?

self-hosted-runner/Dockerfile
```diff
@@ -0,0 +1,51 @@
...
+    npm cache clean --force && \
+    rm -rf /root/.npm
+
+# Test
+RUN node --version && \
+    npm --version && \
+    az --version && \
+    aws --version
+
+COPY entrypoint.sh /entrypoint.sh
```
Krusty93, line 46: **question**: why this COPY?

Krusty93 (commented):
Impressive job!!

self-hosted-runner/entrypoint.sh
mamu0: They are the parameters that `config.sh` needs to stand up the runner:

- `--url` tells it where to register (GitHub repo URL).
- `--token` proves the container is allowed to register.
- `--name` sets the runner’s display name in GitHub.
- `--work` points to the workspace directory for jobs.

In short, they form the minimal configuration required to initialise the GitHub runner non-interactively.

.github/workflows/_release-docker-self-hosted-runner.yaml
mamu0: I think it makes the flow of the **Action** more readable this way

self-hosted-runner/entrypoint.sh
```diff
@@ -0,0 +1,75 @@
...
+  export LABELS=""
+fi
+
+printf "ℹ️ Configuring GitHub Runner for %s\n\t" "$REPO_URL"
+printf "ℹ️ Runner Name: %s\n\t" "$RUNNER_NAME"
+printf "ℹ️ Working Directory: %s\n\t" "$WORK_DIR"
+
+if [ ! -f ".runner" ]; then
+  echo "Configuring the runner..."
+  ./config.sh \
```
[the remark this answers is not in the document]
mamu0: you are right! I update it, thanks

self-hosted-runner/entrypoint.sh
mamu0: Currently i've configured 2 possibilities:

1. **Direct token** – pass RUNNER_TOKEN if you already have a short-lived token (handy for local testing).
2. **Token on the fly** – supply the API URL plus a GitHub PAT so the entry-point can fetch the token automatically, just like the current runner does.

The token is only used for the one-time registration handshake and the PAT is needed only to retrieve that token (in fact if you already have RUNNER_TOKEN, the PAT isn’t needed).
For safety, the script unsets the PAT right after use, so it doesn’t linger as an environment variable.

self-hosted-runner/Dockerfile
Krusty93: Sorry, I wasn't clear enough. I meant, why `COPY entrypoint.sh /entrypoint.sh` and not `COPY entrypoint.sh ./entrypoint.sh`?

self-hosted-runner/entrypoint.sh
Krusty93: I was referring to `--work` only here. Is it needed or just cosmetics? Because it is not set in the old Dockerfile

gunzip:
Can you elaborate more about "better lifecycle management" @Krusty93 ?

Krusty93:
> Can you elaborate more about "better lifecycle management" @Krusty93 ?

We do not have to maintain backward compatibility nor installing tools we don't need because this image is scoped to our toolchain and lifecycle, and not used by other departments. We have more freedom to make changes and customise it according our needs. It uses changeset for versioning, and the automatic image tag update put in place by @mamu0 here is definitely better of what's done in https://github.com/pagopa/github-self-hosted-runner-azure

.github/workflows/_release-docker-self-hosted-runner.yaml
```diff
@@ -0,0 +1,128 @@
...
+# This workflow publishes the latest GitHub Self-Hosted Runner image to the GitHub Container Registry.
+# It checks for the latest official Runner image digest, updates the Dockerfile if necessary,
+# and builds and pushes the new image to the registry.
+# The workflow is scheduled to run every 2 months on the 1st day at 08:00 UTC.
+# It can also be triggered manually for testing purposes using input 'test_run'.
+
+on:
+  workflow_dispatch:
+    inputs:
+      test_run:
```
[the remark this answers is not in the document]
mamu0: what you think about `dry_run` rather than `publish`? Since the workflow normally pushes images and this makes clear that it’s running in a non-publishing mode

self-hosted-runner/Dockerfile
```diff
@@ -0,0 +1,52 @@
+FROM ghcr.io/actions/actions-runner:latest AS base
+
+# === Stage 1: Use base Runner and install tools ===
+ARG RUNNER_DIGEST
+LABEL base_runner_digest=${RUNNER_DIGEST}
+
+USER root
+RUN apt-get update && apt-get install -y --no-install-recommends \
```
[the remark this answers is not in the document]
mamu0: `jq` il already present in the base image `ghcr.io/actions/actions-runner`

self-hosted-runner/entrypoint.sh
Krusty93: Where these scripts use RUNNER_TOKEN? Because the old image didn't export them

lucacavallaro (approved):

PR #634, opened 2025-06-13
Add support for CJS

kin0992:
Closes CES-1071


.changeset/swift-webs-occur.md
```diff
@@ -0,0 +1,7 @@
+---
+"@pagopa/azure-tracing": patch
+---
+
+Add support for CJS
+
+With this release, the `@pagopa/azure-tracing` package now supports CommonJS (CJS) module format, allowing it to be used in environments that do not support ES modules.
```
gunzip, line 7: question: do we need to upgrade the readme as well?

PR #647, opened 2025-06-20
Scaffold CLI application

kin0992:
Create the scaffold for the DX CLI

Closes CES-1085


apps/cli/package.json
```diff
@@ -0,0 +1,37 @@
+{
+  "name": "@pagopa/dx-cli",
+  "version": "0.0.0",
+  "type": "module",
+  "private": true,
```
kin0992, line 5: **note**: at the moment it is private.

I don't know yet how we would like to provide the CLI; if we distribute that as an npm package, then we will remove this field.

apps/cli/tsup.config.ts
```diff
@@ -0,0 +1,13 @@
+import { defineConfig } from "tsup";
+
+export default defineConfig({
+  banner: {
+    js: "#!/usr/bin/env node",
+  },
```
kin0992, line 6: **note**: this adds the shabang in the compiled files, without the need to add it in the `.ts` file

PR #652, opened 2025-06-23
Add `doctor` command

kin0992:
Verify the monorepo has the `npm scripts` used by the DX workflows.

Closes CES-1092


apps/cli/src/adapters/yarn/__tests__/index.test.ts
```diff
@@ -0,0 +1,58 @@
...
+  it("should return an empty array", async () => {
+    (execa as unknown as ReturnType<typeof vi.fn>).mockRejectedValueOnce({
+      exitCode: 1,
+      stdout: "aStdout",
+    });
+
+    const nodeReader = makeNodeReader();
+    const scripts = await nodeReader.getScripts(directory);
+
+    expect(scripts.length).toStrictEqual(0);
```
kin0992, line 55:
```suggestion
```
redundant

apps/cli/src/adapters/yarn/index.ts
```diff
@@ -0,0 +1,16 @@
...
+
+import { NodeReader, ScriptSchema } from "../../domain/node.js";
+
+export const makeNodeReader = (): NodeReader => ({
+  getScripts: async (cwd = process.cwd()) => {
+    const { exitCode, stdout } = await execa("yarn", ["run", "--json"], {
+      cwd,
+    });
+
+    return exitCode === 1
```
kin0992, line 12:
```suggestion
    return exitCode !== 0
```

apps/cli/src/commands/index.ts
lucacavallaro, line 1: `commands` should be moved into `adapters` folders (they are commander implementation details)

apps/cli/src/domain/node.ts
```diff
@@ -0,0 +1,37 @@
...
+
+const requiredRootScripts = ["code-review"];
+
+export const checkMonorepoScripts =
+  (monorepoDir: string) =>
+  async (dependencies: Pick<Dependencies, "logger" | "nodeReader">) => {
+    const { logger, nodeReader } = dependencies;
+
+    const scripts = await unwrapOrLogError(dependencies)(() =>
+      nodeReader.getScripts(monorepoDir),
```
lucacavallaro, line 27: `unwrapOrLogError` throws, but is not explicit - without looking its code I don't understand what is happening here

I suggest to use https://github.com/supermacro/neverthrow to make error more explicit (and track them at type level)

apps/cli/src/adapters/yarn/index.ts
```diff
@@ -0,0 +1,16 @@
+import { execa } from "execa";
+import * as process from "node:process";
+
+import { NodeReader, ScriptSchema } from "../../domain/node.js";
+
+export const makeNodeReader = (): NodeReader => ({
+  getScripts: async (cwd = process.cwd()) => {
+    const { exitCode, stdout } = await execa("yarn", ["run", "--json"], {
+      cwd,
+    });
```
lucacavallaro, line 10: Is running this command necessary? I think the scripts can be checked just by looking at the `script` section of `package.json` (without executing it or inferring the package manager)

apps/cli/src/adapters/yarn/index.ts
```diff
@@ -0,0 +1,16 @@
...
+
+    return exitCode === 1
+      ? []
+      : stdout.split("\n").map((line) => ScriptSchema.parse(JSON.parse(line)));
```
lucacavallaro, line 14: `JSON.parse` can throw (maybe we need to check it or to handle the error differently?)

apps/cli/src/domain/node.ts
```diff
@@ -0,0 +1,37 @@
...
+    );
+    requiredRootScripts.map((script) => {
+      const exists = scripts.some(({ name }) => name === script);
+      if (exists) {
+        logger.success(`Script "${script}" is present in the monorepo root`);
+      } else {
+        logger.error(`Script "${script}" is missing in the monorepo root`);
+      }
+    });
```
lucacavallaro, line 36: I am not sure, but maybe this check can be done with `zod` by parsing the `scripts` by the `requiredScript` schema (like `({ "code-review": z.string().nonempty()`)

apps/cli/src/index.ts
```diff
@@ -1,10 +1,24 @@
...
+const deps: Dependencies = {
+  logger: makeConsoleLogger(),
+  nodeReader: makeNodeReader(),
+  repositoryReader: makeRepositoryReader(),
+};
+
 const program = new Command();
 
 program
   .name("DX-CLI")
```
lucacavallaro, line 18: the name should be aligned by the `bin` script 
```suggestion
  .name("dx")
```

lucacavallaro (changes requested):
Additional note: since the logger is used everywhere, I suggest the use of https://logtape.org which allows us not to pass the `logger` variable explicitly to every function. It also activates only when needed and at the appropriate level, and is compatible with many external transports (console, Otel...)

apps/cli/src/domain/reporisoty.ts
lucacavallaro, line 1: typo in file name

apps/cli/src/domain/node.ts
```diff
@@ -0,0 +1,37 @@
+import { z } from "zod";
+
+import { Dependencies } from "./dependencies.js";
+import { unwrapOrLogError } from "./index.js";
+
+const ScriptName = z.string().brand<"ScriptName">();
+
+export const ScriptSchema = z.object({
```
lucacavallaro, line 8: nitpick: variable names should be in camelCase, so this schema should be renamed `scriptSchema`

apps/cli/src/domain/node.ts
gunzip: what about using fp-ts instead? I see we need DI too (`make` methods) and `Readers` can help.

kin0992: Honestly, I don't know.
I wouldn't mind keeping it simple without adding the `fp-ts` code; this project can also be maintained by someone approaching TypeScript as a beginner. With `fp-ts`, this is more complex to accomplish

apps/cli/src/adapters/node/codec.ts
```diff
@@ -0,0 +1,27 @@
+import { z } from "zod";
```
lucacavallaro, line 1: @kin0992 I am thinking we can start using `zod4`, the new API (https://zod.dev/v4). It should be already included in this NPM package (the import is a little bit different), the API is almost the same but is more performant and there are many "quality of life" upgrades

Give it a try!

apps/cli/src/domain/package-json.ts
```diff
@@ -0,0 +1,65 @@
...
+    }
+
+    const requiredScripts = packageJsonReader.getRootRequiredScripts();
+    const { isValid, missingScripts } = validateRequiredScripts(
+      scriptsResult.value,
+      requiredScripts,
+    );
+
+    if (isValid) {
+      logger.info("✅ Monorepo scripts are correctly set up");
```
lucacavallaro, line 58: here we are mixing domain (BL) layer with UI (Presentation) layer - these layers should be separated to make the domain reusable and unaware of the outside context

in my opinion, `checkMonorepoScript` should returns a `Result` object (custom made, or the result from neverthrow) that then will be adapted to the console output into the `adapters` folder

apps/cli/src/domain/package-json.ts
```diff
@@ -0,0 +1,65 @@
...
+      return ok();
+    }
+
+    const errorMessage = `Missing required scripts: ${missingScripts.join(", ")}`;
+    logger.error(`❌ ${errorMessage}`);
```
lucacavallaro, line 63: imho (to better follow cli conventions) we should treat this error as "non-recoverable error" and we should exit with an exit code different that 0 (e.g. `process.exit(1)`)

this would make this command "abstractions-friendly" (example: `yarn dx doctor` on `code-review` CI), without any other wrappers

lucacavallaro (changes requested):

.changeset/shiny-doors-own.md
```diff
@@ -0,0 +1,8 @@
+---
+"@pagopa/dx-cli": minor
+---
+
+Add `doctor` command
```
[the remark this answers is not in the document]
lucacavallaro: What about `assess`? We thought of this command to speed up our assessment process when new support requests come in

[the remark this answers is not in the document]
lucacavallaro: Yes agree, we can postpone this decision!

PR #655, opened 2025-06-25
Add Terraform MCP server in VSCode

gunzip:
Integrate Docker support for running the Terraform MCP server within the VSCode environment, enhancing development capabilities.

### DIFF
--- a/.devcontainer/templates/node/devcontainer-template.json
+++ b/.devcontainer/templates/node/devcontainer-template.json
@@ -1,6 +1,6 @@
 {
   "id": "node",
-  "version": "1.1.5",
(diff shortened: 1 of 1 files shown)


.devcontainer/templates/node/devcontainer-template.json
```diff
@@ -1,6 +1,6 @@
 {
   "id": "node",
-  "version": "1.0.5",
+  "version": "1.1.5",
```
kin0992, line 3: **suggestion**: if you bump the minor, then set the patch to `0`

gunzip: oops

PR #667, opened 2025-06-27
Add TypeScript package generator with PlopJS

gunzip:
Introduce a PlopJS generator for creating TypeScript workspaces in the DX monorepo.

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,9 +1,4 @@
 {
-  "github.copilot.chat.codeGeneration.instructions": [
-    {
-      "file": ".github/instructions/code.instructions.md"
-    }
-  ],
   "github.copilot.chat.commitMessageGeneration.instructions": [
     {
       "file": ".github/instructions/commit-message.instructions.md"
@@ -19,11 +14,6 @@
       "file": ".github/instructions/review.instructions.md"
     }
   ],
-  "github.copilot.chat.testGeneration.instructions": [
-    {
-      "file": ".github/instructions/test.instructions.md"
-    }
-  ],
   "editor.defaultFormatter": "esbenp.prettier-vscode",
   "editor.formatOnSave": true,
   "[terraform]": {
@@ -35,4 +25,5 @@
   "[go]": {
     "editor.defaultFormatter": "golang.go"
   },
+  "githubPullRequests.codingAgent.uiIntegration": true
 }


packages/typescript-generator/templates/typescript-package/package.json.hbs
```diff
@@ -0,0 +1,41 @@
...
+  },
+  "dependencies": {},
+  "devDependencies": {
+    "@pagopa/eslint-config": "workspace:^",
+    "@tsconfig/node20": "^20.1.5",
+    "@types/node": "^20.12.2",
+    "eslint": "8.57.0",
+    "tsup": "^8.5.0",
+    "typescript": "~5.8.3",
+    "vitest": "^1.6.0"
```
lucacavallaro, line 26: hardcoding these values is the best way to spread legacy dependencies and technical debt! (example: the latest version of `vitest` is `3.x`, I don't think we should suggest to use the `1.x`)

this section should be valued dinamically by the package manager, in a step that should happen before the code generation

example of the dep setup script:

```bash
yarn add -D @pagopa/eslint-config @tsconfig/node20 @types/node eslint@^8.0.0 typescript vitest
yarn add -DE prettier
```

packages/typescript-generator/templates/typescript-package/package.json.hbs
```diff
@@ -0,0 +1,41 @@
...
+  },
+  "scripts": {
+    "start": "node .",
+    "build": "tsup",
+    "dev": "tsup --watch",
+    "typecheck": "tsc --noEmit",
+    "lint": "eslint --fix src",
```
lucacavallaro, line 33: I would make the setup as easier as possible - IMHO there are no reason for pushing to our consumers a code bundler. `tsc` is easy to use and it is good enough for the majority of use case we have.

(I would include `esbuild/tsup` only for advanced setup)

```suggestion
    "build": "tsc",
    "build:watch": "tsc --watch",
    "typecheck": "tsc --noEmit",
    "lint": "eslint --fix src",
```

packages/typescript-generator/templates/typescript-package/tsconfig.json.hbs
```diff
@@ -0,0 +1,12 @@
...
+  "extends": "@tsconfig/node20/tsconfig.json",
+  "compilerOptions": {
+    "outDir": "./dist",
+    "rootDir": "./src",
+    "sourceMap": true,
+    "declaration": true,
+    "moduleResolution": "bundler",
+    "module": "ESNext"
+  },
+  "exclude": ["tsup.config.ts"]
```
lucacavallaro, line 11: if use `rootDir`, this field should be redundant

packages/typescript-generator/templates/typescript-package/tsconfig.json.hbs
```diff
@@ -0,0 +1,12 @@
+{
+  "extends": "@tsconfig/node20/tsconfig.json",
+  "compilerOptions": {
+    "outDir": "./dist",
+    "rootDir": "./src",
+    "sourceMap": true,
+    "declaration": true,
+    "moduleResolution": "bundler",
+    "module": "ESNext"
```
lucacavallaro, line 9: I would use the default values from `@tsconfig/node20`

packages/typescript-generator/templates/typescript-package/vitest.config.ts.hbs
```diff
@@ -0,0 +1,8 @@
+import { defineConfig } from "vitest/config";
+
+export default defineConfig({
+  test: {
+    globals: true,
+    environment: "node",
+  },
+});
```
lucacavallaro, line 8: IMHO enabling `globals` is a bad practice that will make the test more "implicit" and hard to understand

```suggestion
export default defineConfig({
  test: {
    environment: "node",
  },
});
```

packages/typescript-generator/package.json
```diff
@@ -0,0 +1,48 @@
...
+    "typescript": "~5.8.3"
+  },
+  "scripts": {
+    "build": "tsup",
+    "dev": "tsup --watch",
+    "typecheck": "tsc --noEmit",
+    "lint": "eslint --fix src",
+    "lint:check": "eslint src",
+    "format": "prettier --write .",
+    "format:check": "prettier --check ."
```
lucacavallaro, line 46: this script calls `prettier`, but `prettier` is not installed into this project

packages/typescript-generator/templates/typescript-package/package.json.hbs
lucacavallaro: @kin0992 yes for libraries I think is the best option, maybe we can try to use `esbuild` directly (`tsup` is just a thin wrapper of it) and add the latter to the technology radar

packages/typescript-generator/templates/typescript-package/package.json.hbs
```diff
@@ -0,0 +1,41 @@
+{
+  "name": "@pagopa/{{packageName}}",
+  "version": "0.1.0",
+  "type": "module",
+  "description": "{{description}}",
+  "repository": {
+    "type": "git",
+
+    "url": "git+https://github.com/pagopa/dx.git",
+    "directory": "packages/{{packageName}}"
```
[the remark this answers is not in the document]
gunzip: I've rewritten packages sharing a common configuration (`typescript-core-generator`) and adding an instance (`typescript-library-generator`) that generates the `tsup` version.

packages/typescript-generator/src/index.ts
```diff
@@ -0,0 +1,334 @@
...
+
+    const allDependencies = [...baseDependencies, ...additionalDependencies];
+
+    try {
+      // Sanitize dependency names to prevent shell injection
+      const sanitizedDependencies = allDependencies.map((dep) =>
+        dep.replace(/[^a-zA-Z0-9@/._-]/g, ""),
+      );
+
+      execSync(`yarn add -D ${sanitizedDependencies.join(" ")}`, {
```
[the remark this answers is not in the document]
lucacavallaro: @gunzip I changed my mind about how we should handle dependencies 😢 🙏 

Since we will migrate to `pnpm` soon, we will able to leverage catalogs to [centralize and share our constraints to dependencies](https://pnpm.io/catalogs). 

I also thought about the fact that some major versions of certain dependencies (vitest, typescript) can break our abstractions so maybe we need to control when our consume should use a specific major version.

So, in conclusion, I think we can just set good semantic version ranges that respect our constraints. 

The only dependency that does not work with semver ranges is `prettier` but for now we can just set the latest version and we will handle it later.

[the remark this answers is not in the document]
gunzip: we can safely replace yarn add with pnpm add and it will use range version from the catalog. what about this?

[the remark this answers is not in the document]
gunzip: This generator builds a nested project (workspace). I think it's cleaner to use `pnpm add` as it lets you avoid assumptions about what is and what is not into the top level catalog (which you must make if you generate a `package.json` using a template).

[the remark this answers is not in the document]
lucacavallaro: got it, but if you want to avoid dealing with the "monorepo skeleton", you should not run shell commands like `yarn add` or `pnpm add` because you can't assume the pm used without reading the `packageManager` field. `pnpm add` can fail if `pnpm` is not installed or if `packageManager` specifies a different pm.

the safest way to implement that is by valuing the `dependencies` field statically, without using the pm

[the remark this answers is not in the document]
gunzip: > valuing the dependencies field statically

This was the first version you suggested to change as this would pin dependencies to eventually obsolete version. What if we assume pnpm as the package manager? Our "monorepo skeleton generator" will configure it anyway as the default pm.

package.json
```diff
@@ -9,11 +9,13 @@
     "prettier": "catalog:",
     "turbo": "^2.5.4",
     "typescript": "catalog:",
-    "vitest": "catalog:"
+    "vitest": "catalog:",
+    "plop": "^4.0.1"
```
kin0992, line 13: **question**: did you happen to manually add it?

Usually, package managers list dependencies alphabetically.
If you need to install a dependency in the root `package.json` you can run `pnpm -w add -D plop`

packages/typescript-library-generator/package.json
```diff
@@ -0,0 +1,50 @@
+{
+  "name": "@pagopa/typescript-library-generator",
+  "version": "0.1.0",
```
kin0992, line 3: ditto

packages/typescript-library-generator/package.json
```diff
@@ -0,0 +1,50 @@
...
+  "dependencies": {
+    "@pagopa/typescript-generator": "workspace:^",
+    "inquirer": "catalog:",
+    "plop": "catalog:"
+  },
+  "devDependencies": {
+    "@pagopa/eslint-config": "workspace:^",
+    "@tsconfig/node22": "catalog:",
+    "@types/node": "catalog:",
+    "eslint": "catalog:",
```
kin0992, line 36: ditto

packages/typescript-library-generator/README.md
```diff
@@ -0,0 +1,76 @@
...
+```bash
+pnpm add -D @pagopa/typescript-library-generator
+```
+
+## Usage
+
+When using in the DX monorepo, you can run the generator directly from the root directory:
+
+```bash
+pnpm run turbo gen typescript-library-generator
```
kin0992, line 26: ditto

I think the right command should be `pnpm exec turbo gen typescript-library-package`.
If you prompt `typescript-library-generator`, the CLI is asking you to select the generator to use; options are `typescript-package` and `typescript-library-package`

kin0992 (changes requested):
also a rebase is required

packages/typescript-library-generator/README.md
kin0992: it's ok `pnpm turbo gen` as well

PR #676, opened 2025-07-02
Add guide to migrate Opex dashboards and alerts to a new Application Gateway

Krusty93:


apps/website/docs/infrastructure/migrate-opex.md
```diff
@@ -0,0 +1,113 @@
...
+```
+
+7. After applying the empty configuration, all existing Opex dashboards and
+   alerts will be deleted. You can now update your Opex configuration with the
+   new Application Gateway id and run the GitHub Actions workflow to recreate
+   dashboards and alerts.
+
+:::info
+
+Due to internal limitations, the Opex resources will always be created in the
```
gunzip, line 110: suggestion: may you clarify "internal limitations" (of what?)?

apps/website/docs/infrastructure/migrate-opex.md
```diff
@@ -0,0 +1,113 @@
...
+      source  = "hashicorp/azurerm"
+      version = "~> 4.0"
+    }
+  }
+
+  backend "azurerm" {
+    resource_group_name  = "<your-resource-group-name>"
+    storage_account_name = "<your-storage-account-name>"
+    container_name       = "<your-container-name>"
+    key                  = "<your-terraform-state-file-name>"
```
gunzip, line 51: suggestion: I'd make it clear here that these values must refer to the opex tf state only

apps/website/docs/infrastructure/migrate-opex.md
Krusty93: Values are hardcoded inside the application, but I don't think it's relevant for the reader

apps/website/docs/infrastructure/migrate-opex.md
gunzip: I see, but to ensure clarity and prevent reader confusion, it's always best to provide contextual hints precisely when the information is relevant. Otherwise, readers may be left wondering "how" until they reach a subsequent paragraph.

gunzip (approved):

PR #677, opened 2025-07-02
Fix Docusaurus build

gunzip:
Upgrade Docusaurus dependencies to the latest version to improve functionality and performance.

### DIFF
--- a/apps/website/package.json
+++ b/apps/website/package.json
@@ -15,11 +15,11 @@
     "typecheck": "tsc"
   },
   "dependencies": {
-    "@docusaurus/core": "^3.7.0",
-    "@docusaurus/plugin-client-redirects": "3.7.0",
-    "@docusaurus/plugin-content-docs": "^3.7.0",
-    "@docusaurus/preset-classic": "^3.7.0",
-    "@docusaurus/theme-common": "^3.7.0",
+    "@docusaurus/core": "3.8.1",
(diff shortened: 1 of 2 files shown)


Krusty93 (approved):

apps/website/package.json
```diff
@@ -28,12 +28,13 @@
...
-    "@docusaurus/theme-classic": "^3.7.0",
-    "@docusaurus/tsconfig": "^3.7.0",
-    "@docusaurus/types": "^3.7.0",
+    "@docusaurus/module-type-aliases": "3.8.1",
+    "@docusaurus/theme-classic": "3.8.1",
+    "@docusaurus/tsconfig": "3.8.1",
+    "@docusaurus/types": "3.8.1",
     "@types/node": "^20.12.2",
     "@types/react": "^18.0.38",
+    "image-size": "^2.0.2",
```
kin0992, line 37: **question**: is this dependency required by some other dependency we have?

kin0992:
I would also keep track of the changes (using changesets) in this project. Just a friendly reminder for future PR on the website 😄

apps/website/package.json
gunzip: nope

kin0992: so, why has it been introduced into this PR? 🤔

gunzip: it's required by docusaurus and we have to pin it to a particular release in order to avoid errors during the build

PR #678, opened 2025-07-03
Fix VPN Azure AD Application Access Issue

LoFrance:
# Fix VPN Azure AD Application Access Issue

## Problem
The VPN application was configured with a hardcoded application name that caused Azure AD authentication failures for users. Users were encountering the following error:

```
AADSTS50105: Your administrator has configured the application eng-d-app-vpn ('854cef16-360b-4751-9206-21b958321a83') to block users unless they are specifically granted ('assigned') access to the application. The signed in user 'lorenzo.franceschini@pagopa.it' is blocked because they are not a direct member of a group with access, nor had access directly assigned by an administrator.
```

This error occurred because the application display name was not properly configured to match the Azure AD application registration.

## Solution
Modified the Terraform VPN module to accept an external `display_name` variable that can be properly configured for the Azure AD application. This allows for:

- Flexible configuration of the application display name
- Proper alignment with Azure AD application registration
- Prevention of access blocking issues for authorized users

## Changes
- Updated VPN Terraform module to accept `display_name` as an input variable
- Modified the VPN object creation to use the externally provided display name
- Ensured proper mapping between Terraform configuration and Azure AD application

## Testing
- [ ] Verified that the VPN application now accepts the configurable display name
- [ ] Confirmed that users can successfully authenticate without AADSTS50105 errors
- [ ] Validated that the Azure AD application registration matches the configured display name

## Impact
This fix resolves authentication issues for VPN users and provides better configurability for Azure AD application management.

### DIFF
--- a/infra/modules/azure_core_infra/variables.tf
+++ b/infra/modules/azure_core_infra/variables.tf
@@ -42,16 +42,12 @@ variable "pep_subnet_cidr" {
 
 variable "vpn" {
   type = object({
-    cidr_subnet              = optional(string, "")
-    dnsforwarder_cidr_subnet = optional(string, "")
+    cidr_subnet              = string
+    dnsforwarder_cidr_subnet = string
+    display_name             = string
   })
-  description = "VPN configuration. Both 'cidr_subnet' and 'dnsforwarder_cidr_subnet' must be specified together or not at all."
+  description = "VPN configuration"
   default     = {}
-
-  validation {
-    condition     = (var.vpn.cidr_subnet == "" && var.vpn.dnsforwarder_cidr_subnet == "") || (var.vpn.cidr_subnet != "" && var.vpn.dnsforwarder_cidr_subnet != "")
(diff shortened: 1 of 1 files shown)


infra/modules/azure_core_infra/variables.tf
```diff
@@ -42,16 +42,12 @@ variable "pep_subnet_cidr" {
...
 variable "vpn" {
   type = object({
-    cidr_subnet              = optional(string, "")
-    dnsforwarder_cidr_subnet = optional(string, "")
+    cidr_subnet              = string
+    dnsforwarder_cidr_subnet = string
+    display_name             = string
   })
-  description = "VPN configuration. Both 'cidr_subnet' and 'dnsforwarder_cidr_subnet' must be specified together or not at all."
+  description = "VPN configuration"
```
gunzip, line 49: question: why this change?

LoFrance: My understanding is that VPN setup requires CIDR and DNS details and shouldn’t have blank values, but I’d defer to the network specialists for confirmation.

PR #681, opened 2025-07-04
Upgrade `@pagopa/eslint-config` to use Eslint 9

kin0992:
This will add the support for newer eslint version

Closes CES-610

### DIFF
--- a/.yarnrc.yml
+++ b/.yarnrc.yml
@@ -1,6 +1,1 @@
 preferReuse: true
-
-packageExtensions:
-  eslint-plugin-perfectionist@*:
-    dependencies:
-      "@typescript-eslint/types": "*"
--- a/packages/eslint-config/package.json
+++ b/packages/eslint-config/package.json
@@ -20,19 +20,19 @@
     "index.js"
   ],
   "dependencies": {
-    "@eslint/js": "^9.0.0",
-    "eslint-config-prettier": "^9.1.0",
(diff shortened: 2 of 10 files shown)


.yarnrc.yml
```diff
@@ -1,6 +1 @@
 preferReuse: true
-
-packageExtensions:
-  eslint-plugin-perfectionist@*:
-    dependencies:
-      "@typescript-eslint/types": "*"
```
kin0992, line 6: **note**: this is no longer required, since after a dependency bump, the dependency `@typescript-eslint/types` is already included

packages/eslint-config/package.json
```diff
@@ -20,19 +20,19 @@
     "index.js"
   ],
   "dependencies": {
-    "@eslint/js": "^9.0.0",
-    "eslint-config-prettier": "^9.1.0",
-    "eslint-plugin-perfectionist": "^2.9.0",
-    "eslint-plugin-prettier": "^5.1.3",
-    "eslint-plugin-vitest": "^0.5.3",
```
kin0992, line 27: **note**: this is the legacy package, replaced by the `@vitest/eslint-plugin` package

.yarnrc.yml
lucacavallaro: Fixed by PagoPA https://github.com/azat-io/eslint-plugin-perfectionist/pull/150 😎

lucacavallaro (approved):

PR #685, opened 2025-07-04
Fix bug of missing storage_account_id when assigning roles

christian-calabrese:
On specific blob containers, tables or queues


mamu0 (approved):
LGTM

PR #691, opened 2025-07-07
Verify `turbo` configuration using `doctor` command

kin0992:
Verify that `turbo` is configured.

Closes CES-1095


apps/cli/src/adapters/node/repository.ts
```diff
@@ -20,6 +20,21 @@ const findRoot = (dir: string): Result<string, Error> => {
...
+  if (existsSync(turboPath)) {
+    return ok(true);
+  }
+
+  return err(
+    new Error(
+      `${fileName} not found in repository root. Make sure to have Turbo configured for the monorepo.`,
+    ),
+  );
+};
```
lucacavallaro, line 35: I would also check if `turbo` is present in the root `packageJson` file, and it is the latest supported version (`2.x`, you can use the semver package to check that constraint)

apps/cli/src/adapters/node/__tests__/package-json.test.ts
```diff
@@ -1,102 +1,186 @@
...
   const directory = "/some/dir";
 
-  it("parses scripts from package.json", async () => {
-    const mockPackageJson = JSON.stringify({
-      name: "test-package",
-      scripts: {
-        build: "aScript",
-        "code-review": "aCodeReviewScript",
-      },
+  describe("getScripts", () => {
```
kin0992, line 24: **note**: just wrapped in a dedicated `describe`

apps/cli/src/adapters/semver/dependency-version-validator.ts
```diff
@@ -0,0 +1,18 @@
...
+  (): DependencyVersionValidator => ({
+    isValid: (dependency: Dependency, minVersion: string): boolean => {
+      // Check if the dependency version is valid semver or a valid range
+      if (!valid(dependency.version) && !valid(coerce(dependency.version))) {
+        return false;
+      }
+
+      return satisfies(dependency.version, `>=${minVersion}`);
+    },
+  });
```
kin0992, line 18: **note**: this has to be fixed.

it works for fixed versions, but not for `^2.4.5` like versions

kin0992: i'm going to fix it soon

apps/cli/src/domain/repository.ts
```diff
@@ -32,3 +34,66 @@ export const checkPreCommitConfig =
...
+    if (!turboDependency) {
+      return ok({
+        checkName,
+        errorMessage:
+          "Turbo dependency not found in devDependencies. Please add 'turbo' to your devDependencies.",
+        isValid: false,
+      });
+    }
+
+    return dependencyVersionValidator.isValid(
```
lucacavallaro, line 84: IMHO `semver` can be used directly as we use `zod`, without wrapping it in an adapter

I don't think in the future we'll have setups that will require different validator strategy swapped at runtime for the same checks.

apps/cli/src/adapters/node/__tests__/package-json.test.ts
```diff
@@ -1,102 +1,186 @@
...
+import { makeMockPackageJson } from "./data.js";
 
 vi.mock("node:fs/promises", () => ({
   readFile: vi.fn(),
 }));
 
+const mockReadFile = (content: string) =>
+  (readFile as unknown as ReturnType<typeof vi.fn>).mockResolvedValueOnce(
+    content,
+  );
```
lucacavallaro, line 15: maybe this can be simplified using `spyOn`

```js
import fs from "node:fs/promises";
// ....
const mockReadFile = vi.spyOn(fs, "readFile").mockResolvedValueOnce(content);
```

kin0992: Thanks! Changed in https://github.com/pagopa/dx/pull/691/commits/23aa318db6b48dfeb768aaf7bb1a515f9e75a5ba

lucacavallaro (approved):

PR #695, opened 2025-07-07
Make js_code_review compatible with npm and pnpm

lucacavallaro:
Depends on #694

### DIFF
--- a/.github/workflows/js_code_review.yaml
+++ b/.github/workflows/js_code_review.yaml
@@ -34,6 +34,7 @@ jobs:
       # Example: "turbo run typecheck format:check lint:check //#test:coverage"
       # See https://github.com/pagopa/io-fims/pull/41
       - name: Run code-review script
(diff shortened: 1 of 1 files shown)


lucacavallaro:
> **issue**: it looks like the setup node does not work properly.
> 
> take a look at this [code review run, that points to the branch of this PR](https://github.com/pagopa/dx-playground/actions/runs/16120699488/job/45485481784): the workflow shows the steps to install the dependencies using the proper packageManager, but the previous step, setup node, it looks it is not executing the action

Yes @kin0992, because this workflow `setup-node@main` which is the older version that does not call `detect-node-package-manager` and it is part of this change.

You can test the workflow locally following these steps:
- edit the js_code_workflow to reference `./.github/actions/setup-node` instead of `pagopa/dx/setup-node@main`
- run `js_code_review` via `act` 

   ```sh
   act -W .github/workflows/_validate-typescript-code-review.yaml workflow_dispatch
   ```

PR #702, opened 2025-07-07
Update Private Endpoint names in Function App module

Krusty93:
Close CES-1116


.changeset/thick-mammals-dress.md
```diff
@@ -0,0 +1,5 @@
+---
+"azure_function_app": major
+---
+
+Update names of storage account private endpoints
```
gunzip, line 5: in case of breaking changes remember to add here an alert and a description of the breaking change and the path to migrate to this new major

PR #659, opened 2025-06-26
Refactor PR Comment Action to TypeScript

gunzip:
Transition the PR Comment Manager from a YAML composite action to a TypeScript-based implementation for enhanced maintainability and organization. Update workflows to utilize the new local path for the action.

PR #711, opened 2025-07-08
Add action to manage incremental rollout in AppService workflow

Krusty93:
Close CES-1122


.github/actions/incremental-rollout-appservice/action.yml
```diff
@@ -0,0 +1,122 @@
+name: 'Incremental Rollout for App Service'
```
lucacavallaro, line 1: **question:** doesn't this action also work for function apps? the cli commands should be the same

.github/actions/incremental-rollout-appservice/action.yml
```diff
@@ -0,0 +1,122 @@
...
+        set -euo pipefail
+
+        currentPercentage=0
+        stagingSlot=staging
+        prodSlot=production
+
+        while true; do
+          case "$SCRIPT_LANGUAGE" in
+            "typescript (npm)")
+              output="$(npm run -s "$SCRIPT_NAME" $currentPercentage)"
```
lucacavallaro, line 39: **question:** what does the `-s` do? do we really need to support multiple `$SCRIPT_LANGUAGE`?

kin0992 (commented):
**question**: should we move this action under the [`actions` directory (as sibling of `apps`)](https://github.com/pagopa/dx/tree/main/actions/), as already done for other actions?

in this way we could version it (and keep track of changes using changesets)

.github/actions/incremental-rollout-appservice/action.yml
```diff
@@ -0,0 +1,122 @@
...
+      shell: bash
+      env:
+        RESOURCE_GROUP_NAME: ${{ inputs.resource_group_name }}
+        WEB_APP_NAME: ${{ inputs.web_app_name }}
+      run: |
+        az webapp deployment slot swap \
+          --resource-group "$RESOURCE_GROUP_NAME" \
+          --name "$WEB_APP_NAME" \
+          --slot staging \
+          --target-slot production
```
lucacavallaro, line 122: **issue:** this file should be moved to `actions/incremental-rollout-appservice`

lucacavallaro (changes requested):

kin0992:
> **question**: should we move this action under the [`actions` directory (as sibling of `apps`)](https://github.com/pagopa/dx/tree/main/actions/), as already done for other actions?
> 
> in this way we could version it (and keep track of changes using changesets)

if you agree with this, then the next step would be to address the "name" issue: take a look at https://github.com/pagopa/dx/pull/659#discussion_r2192445471

cc @pagopa/engineering-team-devex

.github/actions/incremental-rollout-appservice/action.yml
Krusty93: `-s` is the silent mode to limit logging
About `$SCRIPT_LANGUAGE`: https://github.com/pagopa/dx/pull/531#discussion_r2093575064

actions/incremental-rollout-appservice/action.yml
```diff
@@ -0,0 +1,122 @@
...
+  resource_group_name:
+    description: "Azure Resource Group name"
+    required: true
+  web_app_name:
+    description: "Azure Web App name"
+    required: true
+  script_language:
+    description: "Script language (typescript (npm) or typescript (yarn))"
+    required: true
+    default: "typescript (npm)"
```
mamu0, line 14: **suggestion**: You could use more user-friendly values for this variable, such as `npm` and `yarn`? Or maybe, if the language is always **TypeScript**, you might even remove this input entirely for example perform a check in the root directory where if a `yarn.lock` file is present, assume yarn; otherwise, default to npm.

mamu0, line 14:
```suggestion
```

If `required` is `true` the default will be not used

actions/incremental-rollout-appservice/action.yml
```diff
@@ -0,0 +1,122 @@
...
+            az webapp traffic-routing clear \
+              --resource-group "$RESOURCE_GROUP_NAME" \
+              --name "$WEB_APP_NAME"
+            exit 1
+          fi
+
+          swap=$(echo "$output" | jq -r '.swap // empty')
+          if [ "$swap" == "true" ]; then
+            echo "::warning::Swap to new version requested."
+            echo "::notice::Swap to staging requested. Moving all traffic to staging."
```
mamu0, line 64:
```suggestion
            echo "::notice::Swap requested. Shifting 100% traffic to staging slot."
```

actions/incremental-rollout-appservice/action.yml
Krusty93: The goal is to add support to other languages such as Java shortly. I don't think we can tie this logic to the mono repository setting. We haven't identified a multi-language package manager yet.
At the moment we only support typescript, but in next PRs we can easily add support to any language we want to. I do not think they have to be tied with the package manager  language. They're totally independent (I could use bash, Java or anything else I find friendly).

mamu0: I understand your point of view. However, I’d suggest making the input values more user-friendly (for example, using names like `npm`, `yarn`, `bash`, or `java`). Values like `typescript (npm)` aren’t very intuitive or easy to remember, so they should be clearly documented

gunzip: i guess that in case we should suppress any extranous turbo output, isn't it?

lucacavallaro: You are right 😢 What about calling just an endpoint from the service instead of local script? I was thinking that maybe this kind of scripts uses other backing services to assess the status, so it can be easier to implement on the service itself.

If we prefer a local script at this point I suggest requiring just a bash script that runs the appropriate commands.

gunzip: Ok for bash script, calling an endpoint makes thigs more complex since it needs appropriate rights (ie. to access an application insights instance), to already be online and to be unreachable from internet (as it may perform some heavy weight task). It's ok for me to just call a bash script.

actions/incremental-rollout-appservice/action.yml
```diff
@@ -0,0 +1,122 @@
...
+        SCRIPT_NAME: ${{ inputs.script_name }}
+        RESOURCE_GROUP_NAME: ${{ inputs.resource_group_name }}
+        WEB_APP_NAME: ${{ inputs.web_app_name }}
+      run: |
+        set -euo pipefail
+
+        currentPercentage=0
+        stagingSlot=staging
+        prodSlot=production
+
```
christian-calabrese, line 35: It may be nice to show the progress in the github step summary as done for the static analysis (more will follow)
```suggestion
        echo "## 🧪 Incremental Rollout Progress" >> "$GITHUB_STEP_SUMMARY"

```

actions/incremental-rollout-appservice/action.yml
```diff
@@ -0,0 +1,122 @@
...
+          swap=$(echo "$output" | jq -r '.swap // empty')
+          if [ "$swap" == "true" ]; then
+            echo "::warning::Swap to new version requested."
+            echo "::notice::Swap to staging requested. Moving all traffic to staging."
+            az webapp traffic-routing set \
+              --resource-group "$RESOURCE_GROUP_NAME" \
+              --name "$WEB_APP_NAME" \
+              --distribution staging=100,production=0
+            break
```
christian-calabrese, line 69:
```suggestion
            az webapp traffic-routing set \
              --resource-group "$RESOURCE_GROUP_NAME" \
              --name "$WEB_APP_NAME" \
              --distribution staging=100,production=0
            echo "🚀 Immediate swap requested. 100% traffic moved to staging." >> "$GITHUB_STEP_SUMMARY"
            break
```

actions/incremental-rollout-appservice/action.yml
```diff
@@ -0,0 +1,122 @@
...
+
+          exit_code=$?
+          echo "Script Output: $output"
+
+          if [ $exit_code -ne 0 ]; then
+            echo "::error::Monitoring script failed. Reverting traffic to production."
+            az webapp traffic-routing clear \
+              --resource-group "$RESOURCE_GROUP_NAME" \
+              --name "$WEB_APP_NAME"
+            exit 1
```
christian-calabrese, line 58:
```suggestion
            az webapp traffic-routing clear \
              --resource-group "$RESOURCE_GROUP_NAME" \
              --name "$WEB_APP_NAME"
            echo "❌ Canary script failed at ${currentPercentage}%. Reverting traffic." >> "$GITHUB_STEP_SUMMARY"
            exit 1
```

actions/incremental-rollout-appservice/action.yml
```diff
@@ -0,0 +1,122 @@
...
+
+          nextPercentage=$(echo "$output" | jq -r '.nextPercentage')
+          afterMs=$(echo "$output" | jq -r '.afterMs')
+
+          if [ -z "$nextPercentage" ] || [ -z "$afterMs" ]; then
+            echo "::error::Invalid output from script (missing nextPercentage or afterMs). Reverting traffic."
+            az webapp traffic-routing clear \
+              --resource-group "$RESOURCE_GROUP_NAME" \
+              --name "$WEB_APP_NAME"
+            exit 1
```
christian-calabrese, line 80:
```suggestion
            az webapp traffic-routing clear \
              --resource-group "$RESOURCE_GROUP_NAME" \
              --name "$WEB_APP_NAME"
            echo "❌ Invalid script output at ${currentPercentage}%. Reverting traffic." >> "$GITHUB_STEP_SUMMARY"
            exit 1
```

actions/incremental-rollout-appservice/action.yml
```diff
@@ -0,0 +1,122 @@
...
+          fi
+
+          echo "::notice::Next traffic percentage: $nextPercentage% to staging"
+          az webapp traffic-routing set \
+            --resource-group "$RESOURCE_GROUP_NAME" \
+            --name "$WEB_APP_NAME" \
+            --distribution staging=$nextPercentage,production=$((100-nextPercentage))
+
+          currentPercentage=$nextPercentage
+
```
christian-calabrese, line 90:
```suggestion
          currentPercentage=$nextPercentage
          echo "✅ Rolled out to ${currentPercentage}% traffic on staging" >> "$GITHUB_STEP_SUMMARY"

```

actions/incremental-rollout-appservice/action.yml
```diff
@@ -0,0 +1,122 @@
...
+          if [ "$nextPercentage" -ge 100 ]; then
+            echo "::notice::Successfully shifted 100% traffic to staging slot"
+            break
```
christian-calabrese, line 93:
```suggestion
            echo "::notice::Successfully shifted 100% traffic to staging slot"
            echo -e "\n## ✅ Rollout Completed\n🎉 Incremental rollout and slot swap completed successfully. 100% of traffic is now running on the new version in the production slot." >> "$GITHUB_STEP_SUMMARY"
            break
```

actions/swap-appsvc-slot/action.yml
```diff
@@ -0,0 +1,32 @@
...
+      env:
+        GITHUB_ACTION_PATH: ${{ github.action_path }}
+      run: echo "$GITHUB_ACTION_PATH" >> $GITHUB_PATH
+
+    - name: Incremental Rollout to Staging Slot
+      shell: bash
+      env:
+        RESOURCE_GROUP_NAME: ${{ inputs.resource_group_name }}
+        WEB_APP_NAME: ${{ inputs.web_app_name }}
+      run: rollout.sh "$RESOURCE_GROUP_NAME" "$WEB_APP_NAME"
```
gunzip, line 32: question: does this work even when calling the action from another repo?

gunzip (approved):
very good!

actions/swap-appsvc-slot/action.yml
lucacavallaro: Yes! the previous step adds the "GITHUB_ACTION_PATH" to the PATH of the runner :)

lucacavallaro (approved):

actions/swap-appsvc-slot/rollout.sh
```diff
@@ -0,0 +1,149 @@
...
+
+cat <<EOF > "$GITHUB_STEP_SUMMARY"
+Successfully swapped staging slot to production.
+
+> [!TIP]
+> You can enable the canary deployment by creating a script named \`canary-monitor.sh\` in the root of your repository.
+> Find more details in the [workflow documentation](https://pagopa.github.io/dx/docs/pipelines/release-azure-appsvc#canary-deployments).
+EOF
+
+
```
Krusty93, line 148:
```suggestion
```

actions/swap-appsvc-slot/rollout.sh
```diff
@@ -0,0 +1,149 @@
...
+  local y_axis
+
+  x_axis=$(jq -src [.[].time] events.json)
+  y_axis=$(jq -src [.[].canary] events.json)
+
+  cat <<EOF > "$GITHUB_STEP_SUMMARY"
+### Canary Deployment Rollout
+$message
+
+\`\`\`mermaid
```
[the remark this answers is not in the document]
lucacavallaro: We write a summary even when the canary fails, after `az webapp traffic-routing clear` with the same chart and the reason behind the failure.

https://github.com/pagopa/dx/pull/711/files/d40bbc0d5636f5ff5692e6bb0f72504c7d029944#diff-0c0b663e2534d63a8eeca790ea9dda54a7cd66113477eadd41c66bbe7695e930R51

PR #716, opened 2025-07-09
Verify workspaces configuration using `doctor` command

kin0992:
Closes CES-1094


apps/cli/src/adapters/node/fs/__tests__/file-reader.test.ts
```diff
@@ -68,28 +68,6 @@ describe("readFileAndDecode", () => {
   });
 });
 
-describe("fileExists", () => {
-  beforeEach(() => {
-    vi.clearAllMocks();
-  });
-
-  it("should return true when the path exists", async () => {
-    vi.spyOn(fs, "stat").mockResolvedValueOnce({} as Stats);
```
lucacavallaro, line 77: **note:** to do file system testing it can be useful to use [memfs, that allows to work with inmemory files using the same `fs` api](https://medium.com/nerd-for-tech/testing-in-node-js-easy-way-to-mock-filesystem-883b9f822ea4). It can be useful to avoid using `as` and making mocks hard to understand (this is not the case!)

apps/cli/src/adapters/node/json/index.ts
```diff
@@ -0,0 +1,6 @@
+import { Result } from "neverthrow";
+
+export const toJSON = Result.fromThrowable(
```
lucacavallaro, line 3: **nitpick:** at first glance, this function seems to do the opposite of what its name says. I somehow expected to get a `JSON string` as output instead of a JS object.

I suggest to rename to `parseJson`

```suggestion
export const parseJson = Result.fromThrowable(
```

apps/cli/src/adapters/node/repository.ts
```diff
@@ -18,8 +24,56 @@ const findRepositoryRoot = (
...
+    // For now it is not possible to use the fs.glob function (from node:fs/promises)
+    // because it is not possible to run it on Node 20.x
+    glob.glob(pattern, { cwd: repoRoot }),
+    (cause) =>
+      new Error(`Failed to resolve workspace glob: ${pattern}`, {
+        cause,
+      }),
+  ).map((subDirectories) =>
+    // Create the absolute path to the subdirectory
+    subDirectories.map((directory) => join(repoRoot, directory)),
```
lucacavallaro, line 38: **nitpick:** can you include the namespace here? I think `join` is a generic name hard to understand reading the code, `path.join` it is clearer.


```suggestion
    subDirectories.map((directory) => path.join(repoRoot, directory)),
```

apps/cli/src/adapters/node/repository.ts
```diff
@@ -18,8 +24,56 @@ const findRepositoryRoot = (
...
+  );
+
+const getWorkspaces = (repoRoot: string): ResultAsync<Workspace[], Error> =>
+  readFile(join(repoRoot, "pnpm-workspace.yaml"))
+    .andThen(parseYaml)
+    // Decode the pnpm-workspace.yaml file to a zod schema
+    .andThen((json) =>
```
lucacavallaro, line 45: **question:** `parseYaml` outputs a `JSON`?

apps/cli/src/domain/workspace.ts
```diff
@@ -0,0 +1,52 @@
...
+  const { repositoryReader } = dependencies;
+  const checkName = "Workspaces";
+
+  const workspacesResult = await repositoryReader.getWorkspaces(monorepoDir);
+
+  if (workspacesResult.isErr()) {
+    return ok({
+      checkName,
+      errorMessage:
+        "Something is wrong with the workspaces configuration. If you need help, please contact the DevEx team.",
```
lucacavallaro, line 29: **suggestion:** I think it can be useful to print a "debug object" with the actual error and state to help us DX team give support to others.

apps/cli/src/adapters/node/repository.ts
kin0992: It returns a JS object. Would you change the `json` parameter name? Maybe to `obj` or something that reflects the fact it is a JS object?

lucacavallaro: i find `obj` better than `json`

PR #739, opened 2025-07-16
Refactors/go-modules

lucacavallaro:


providers/azure/docs/functions/resource_name.md
```diff
@@ -38,102 +36,4 @@ resource_name(configuration map of string) string
...
-| table_private_endpoint                 |    table-pep     |
-| function_table_private_endpoint        |  func-table-pep  |
-| dfunction_table_private_endpoint       | dfunc-table-pep  |
-| table_storage                          |      table       |
-| virtual_machine                        |        vm        |
-| virtual_network                        |       vnet       |
-| dns_forwarding_ruleset                 |      dnsfrs      |
-| dns_private_resolver                   |      dnspr       |
-| dns_private_resolver_inbound_endpoint  |        in        |
-| dns_private_resolver_outbound_endpoint |       out        |
```
mamu0, line 139: I’d suggest keeping this section, since it’s the page that appears in the registry to document the resource:
https://registry.terraform.io/providers/pagopa-dx/azure/latest/docs/functions/resource_name
Without it, it’s unclear how the resource is meant to be used.

providers/azure/docs/resources/available_subnet_cidr.md
```diff
@@ -1,82 +0,0 @@
----
-page_title: "dx_available_subnet_cidr Resource - terraform-provider-azure"
```
mamu0, line 2: ditto
this is the page about the subnet cidr resource
https://registry.terraform.io/providers/pagopa-dx/azure/latest/docs/resources/available_subnet_cidr

providers/azure/docs/resources/dx_available_subnet_cidr.md
```diff
@@ -0,0 +1,5 @@
+---
+# generated by https://github.com/hashicorp/terraform-plugin-docs
```
mamu0, line 2: this file i think is not complete

mamu0:
> ... find a way to do this step automatically without the need to backup and copy paste files. I will think about it.

That makes sense! But keep in mind that adding a new resource isn’t something we do very often, and documentation updates are made directly in the files themselves.
So if it’s too complex to make the tool ignore existing files, we might consider handling it manually in those rare cases, it’s a small trade-off for keeping things simple.

providers/azure/docs/resources/dx_available_subnet_cidr.md
mamu0: This file already exists under a different name: [available_subnet_cidr.md](https://github.com/pagopa/dx/blob/main/providers/azure/docs/resources/available_subnet_cidr.md)

is the documentation for the subnet CIDR resource that's visible on the Terraform Registry

mamu0:
Is this PR still active, or would it make sense to move it to draft or close it?

PR #741, opened 2025-07-16
Add reusable workflow for Azure CDN deployment

mamu0:
This PR introduces a new reusable workflow to automate the build and deployment of static sites from our monorepo to Azure CDN.

Resolves: CES-1051

### DIFF
--- a/.github/workflows/release-typescript-cdn-deploy-v1.yaml
+++ b/.github/workflows/release-typescript-cdn-deploy-v1.yaml
@@ -106,8 +106,10 @@ jobs:
 
       - name: Find build output directory path
         id: find_package
+        env:
+           TURBO_PRINT_VERSION_DISABLED: 1
         run: |
-          packages_path=$(TURBO_PRINT_VERSION_DISABLED=1 npx turbo ls --output json | jq -sr --arg WORKSPACE "$WORKSPACE" '.[].packages.items[] | select(.name==$WORKSPACE).path')
+          packages_path=$(npx turbo ls --output json | jq -sr --arg WORKSPACE "$WORKSPACE" '.[].packages.items[] | select(.name==$WORKSPACE).path')
           echo "Found $OUTPUT_DIR directory at: ./out/$packages_path/$OUTPUT_DIR"
           echo "packages_path=./out/$packages_path/$OUTPUT_DIR" >> $GITHUB_OUTPUT
         env:


.github/workflows/release-typescript-cdn-deploy-v1.yaml
```diff
@@ -0,0 +1,160 @@
...
+        env:
+          PACKAGE_MANAGER: ${{ steps.node-setup.outputs.package-manager }}
+        working-directory: ./out
+
+      - name: Find dist directory path
+        id: find_dist
+        run: |
+          dist_path=$(find ./out -path '*/node_modules' -prune -o -type d -name dist -print | head -n 1)
+          echo "Found dist directory at: $dist_path"
+          echo "dist_path=$dist_path" >> $GITHUB_OUTPUT
```
lucacavallaro, line 111: **question:** why is this step required?

.github/workflows/release-typescript-cdn-deploy-v1.yaml
```diff
@@ -0,0 +1,160 @@
...
+
+      - name: Install dependencies (pnpm)
+        if: steps.node-setup.outputs.package-manager == 'pnpm'
+        run: pnpm install --frozen-lockfile
+        working-directory: ./out
+
+      - name: Build project
+        run: $PACKAGE_MANAGER run build
+        env:
+          PACKAGE_MANAGER: ${{ steps.node-setup.outputs.package-manager }}
```
lucacavallaro, line 103: **suggestion:** I think we can try use directly turbo

```suggestion
        run: npx turbo build
```

.github/workflows/release-typescript-cdn-deploy-v1.yaml
```diff
@@ -0,0 +1,160 @@
...
+          blob_container_name: "${{ env.BLOB_CONTAINER_NAME }}"
+          resource_group_name: ${{ env.RESOURCE_GROUP_NAME }}
+          profile_name: ${{ env.CDN_PROFILE_NAME }}
+          endpoint_name: ${{ env.CDN_ENDPOINT_NAME }}
+        env:
+          STORAGE_ACCOUNT_NAME: ${{ inputs.storage_account_name }}
+          RESOURCE_GROUP_NAME: ${{ inputs.resource_group_name }}
+          CDN_PROFILE_NAME: ${{ inputs.profile_name }}
+          CDN_ENDPOINT_NAME: ${{ inputs.endpoint_name }}
+          BLOB_CONTAINER_NAME: ${{ inputs.blob_container_name }}
```
lucacavallaro, line 160: **question:** can't we use the inputs directly?

.github/workflows/release-typescript-cdn-deploy-v1.yaml
mamu0: is needed because turbo prune preserves the original folder structure inside the `out/` directory. This makes the final `dist` folder's path unpredictable. This step dynamically finds the correct path to ensure the artifact upload always works, regardless of the workspace's location.

lucacavallaro: can't we assume that the dist folder will be locaed in `out/workspace/dist` ?

.github/workflows/release-typescript-cdn-deploy-v1.yaml
mamu0: I wouldn’t recommend it, because it was meant to address a potential code injection issue reported by Copilot in the PR (which is no longer present thanks to the recent change)

.github/workflows/release-typescript-cdn-deploy-v1.yaml
lucacavallaro: What if there are multiple `dist` folder? what if the output folder has a different name? (like `public`, or `static`)?

mamu0: Could you please help me figure out how to solve this issue?
From what I understand, there shouldn’t be multiple `dist` folders under `out`, since the build only targets the specific package you pass to it.
At least, that’s how I believe it works, but if the folders have different names, I’m honestly not sure.

lucacavallaro: the build targets the specific package but it affect also its dependencies (if a monorepo have `app2` that depends on `app1` the `build` with `app2` as target will build also `app1`) so we need to restrict this search on the specific workspace folder

to obtain the workspace path by its name you can use `turbo` from the `./out` folder

```sh
TURBO_PRINT_VERSION_DISABLED=1  turbo ls --format json
```

you will get an output like
```json
{
  "packageManager": "pnpm9",
  "packages": {
    "count": 34,
    "items": [
      {
        "name": "@pagopa/azure-tracing",
        "path": "packages/azure-tracing"
      },
      {
        "name": "@pagopa/dx-cli",
        "path": "apps/cli"
      },
      {
        "name": "@pagopa/eslint-config",
        "path": "packages/eslint-config"
      }
    ]
  }
}
```

that you can filter with `jq` to extract the path

```sh
TURBO_PRINT_VERSION_DISABLED=1 npx turbo ls --output json | jq -sr --arg WORKSPACE "$WORKSPACE" '.[].packages.items[] | select(.name==$WORKSPACE).path'
```

---

I suggest to also add an `input` for specify the dist folder name, with `dist` as default

mamu0: i've updated the PR with this changes @lucacavallaro

PR #750, opened 2025-07-18
Refactor `doctor` command code

kin0992:
Move the logic in the `domain` layer, then pass the proper environment to the command (which executes the function).


apps/cli/src/adapters/commander/commands/doctor.ts
```diff
@@ -29,32 +30,50 @@ export const makeDoctorCommand = (
...
+              () => new Error("Error checking pre-commit configuration"),
+            ),
+            ResultAsync.fromPromise(
+              checkTurboConfig(repoRoot)(dependencies, config),
+              () => new Error("Error checking Turbo configuration"),
+            ),
+            ResultAsync.fromPromise(
+              checkMonorepoScripts(repoRoot)(dependencies),
+              () => new Error("Error checking monorepo scripts"),
+            ),
```
kin0992, line 48: **note**: i didn't change the `check*` return type on purpose.

we could decide to make those functions to return a ResultAsync, then this block is going to be even more simple to read (and does not have the error overhead, which is going to be wrapped inside the ResultAsync)

apps/cli/src/adapters/commander/commands/doctor.ts
```diff
@@ -29,32 +30,50 @@ export const makeDoctorCommand = (
...
+                process.exit(1);
+              }
+              return okAsync(void 0);
+            });
+        })
+        .orElse(({ message }) => {
+          logger.error(message);
+          process.exit(1);
+        });
     });
```
lucacavallaro, line 78: **Suggestion(non-blocking):** This snippet mixes imperative and declarative logic, and contains a `map` operation with a side effect. I think it is hard to understand and difficult to extract if in the future we want support other UI layers.

In my opinion, `ResultAsync.combine.map(...)` should produce a `DoctorResult` object that later will be printed on console.

Something like:

```js
const checks = [check1, check2, check3...]
const result = combine(checks).map(tofinalShape) // no printing here

printResult(result)

const exitCode = result.hasError ? 1 : 0;
process.exit(exitCode)
```

PR #751, opened 2025-07-18
Allow queue readers to set visibility timeout

christian-calabrese:
Assign the new custom role "PagoPA Storage Queue Data Message Contributor" to storage queue "reader" role


lucacavallaro:
@christian-calabrese why is this change needed? I find confusing to have a `Data Message Contributor` role with a `reader` label

mamu0 (approved):

PR #756, opened 2025-07-23
Show packageManager in `info` command

kin0992:
Create the `info` command and show the `packageManager` used in the project

Closes CES-1150

### DIFF
--- a/apps/cli/eslint.config.js
+++ b/apps/cli/eslint.config.js
@@ -1,3 +1,11 @@
 import lintRules from "@pagopa/eslint-config";
 
-export default lintRules;
+export default [
+  ...lintRules,
+  {
+    rules: {
+      // Override this rules to show an error when using interfaces, enforcing the use of types
+      "@typescript-eslint/consistent-type-definitions": ["error", "type"],
+    },
+  },
+];
--- a/apps/cli/src/config.ts
+++ b/apps/cli/src/config.ts
@@ -1,11 +1,11 @@
-export interface Config {
+export type Config = {
   minVersions: {
     turbo: string;
   };
   repository: {
     root: string;
   };
-}
+};
 
 export const getConfig = (repositoryRoot: string): Config => ({
   minVersions: {
--- a/apps/cli/src/config.ts
+++ b/apps/cli/src/config.ts
@@ -2,10 +2,16 @@ export interface Config {
   minVersions: {
     turbo: string;
   };
+  repository: {
+    root: string;
+  };
 }
 
-export const getConfig = (): Config => ({
+export const getConfig = (repositoryRoot: string): Config => ({
   minVersions: {
     turbo: "2",
+  },
+  repository: {
+    root: repositoryRoot,
   },
 });
--- a/apps/cli/src/domain/info.ts
+++ b/apps/cli/src/domain/info.ts
@@ -3,16 +3,16 @@ import { join } from "node:path";
 
 import { Config } from "../config.js";
 import { Dependencies } from "./dependencies.js";
-import { PackageJson } from "./package-json.js";
+import { PackageManager } from "./package-json.js";
 
 export interface InfoResult {
-  packageManager: "npm" | "pnpm" | "yarn";
+  packageManager: PackageManager;
 }
 
 const detectFromLockFile = async (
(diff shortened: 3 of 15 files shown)


apps/cli/src/domain/repository.ts
```diff
@@ -10,6 +10,7 @@ import { ValidationCheckResult } from "./validation.js";
 export interface RepositoryReader {
   existsPreCommitConfig(repoRoot: string): ResultAsync<boolean, Error>;
   existsTurboConfig(repoRoot: string): ResultAsync<boolean, Error>;
+  fileExists(path: string): ResultAsync<boolean, Error>;
```
kin0992, line 13: **note**: we could refactor this interface, removing the `existsPreCommitConfig` and `existsTurboConfig` functions in favour of the `fileExists` function.

I would do in another PR

apps/cli/src/config.ts
```diff
@@ -2,10 +2,16 @@ export interface Config {
   minVersions: {
     turbo: string;
   };
+  repository: {
+    root: string;
+  };
```
kin0992, line 7: **note**: if we pass the repo root in the configuration, then we could also simplify some logic in other commands.

In case, I would do in a separate PR

apps/cli/src/index.ts
```diff
@@ -11,22 +16,53 @@ import { Dependencies } from "./domain/dependencies.js";
...
       lowestLevel: "warning",
       sinks: ["console"],
     },
   ],
-  sinks: { console: getConsoleSink() },
+  sinks: {
+    console: getConsoleSink(),
+    json: getConsoleSink({
+      formatter: getJsonLinesFormatter(),
+    }),
```
kin0992, line 30: **note**: this use a JSON formatter

apps/cli/src/domain/info.ts
```diff
@@ -0,0 +1,70 @@
...
+export const getInfo = async (
+  dependencies: Dependencies,
+  config: Config,
+): Promise<InfoResult> => ({
+  packageManager: await detectPackageManager(dependencies, config),
+});
+
+export const printInfo = (result: InfoResult): void => {
+  const logger = getLogger("json");
+  logger.info(result as unknown as Record<string, unknown>);
```
kin0992, line 69: **note**: if `InfoResult` were a type, it wouldn't be necessary to cast to a Record.
Unfortunately, we have to deal with interfaces.

Moreover, if we want just to print out the bare InfoResult JSON (e.g. ```{"packageManager: "pnpm"}```, we should write a TextFormatter.
The one used, which removes some noise on the console, produces a JSON, but the message is wrapped inside the `message` property.

kin0992: there is an eslint rules that prefer interfaces to types

apps/cli/src/domain/info.ts
```diff
@@ -0,0 +1,70 @@
...
+const detectFromPackageJson = ({ packageManager }: PackageJson) => {
+  if (packageManager?.startsWith("pnpm")) {
+    return "pnpm";
+  } else if (packageManager?.startsWith("yarn")) {
+    return "yarn";
+  } else if (packageManager?.startsWith("npm")) {
+    return "npm";
+  }
+  return undefined;
+};
```
lucacavallaro, line 58: **suggestion:** if you want to simplify that part I think you can make a `zod` parser

if we have 

```js
const packageManagerSchema = z.enum(["pnpm", "yarn", "npm"])
```

we can build another schema

```js
const packageManagerFromPackageJsonSchema = z.string().transform(pm => pm.split("@") ... ).pipe(packageManagerSchema)
```

apps/cli/eslint.config.js
```diff
@@ -1,3 +1,11 @@
 import lintRules from "@pagopa/eslint-config";
 
-export default lintRules;
+export default [
+  ...lintRules,
+  {
+    rules: {
+      // Override this rules to show an error when using interfaces, enforcing the use of types
+      "@typescript-eslint/consistent-type-definitions": ["error", "type"],
```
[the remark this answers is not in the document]
kin0992: for now, the CLI prefers types over interfaces

PR #765, opened 2025-07-25
Implementation of AWS Core Infra module

christian-calabrese:
Resolves CES-1140


infra/modules/aws_core_infra/_modules/networking/variables.tf
```diff
@@ -0,0 +1,42 @@
+variable "naming_config" {
+  type = object({
+    prefix          = string
+    environment     = string
+    region          = string
+    name            = string
+    domain          = optional(string)
```
mamu0, line 7: **suggestion**: Because this is the core infrastructure maybe we could remove the domain value

christian-calabrese: Done!

PR #768, opened 2025-07-28
Add new field retention_description field to EventHub module

christian-calabrese:
In version 4.37.0 of the azurerm provider, a new field `retention_description` has been added to the `azurerm_eventhub` resource. The module has been updated to support it.


.changeset/fifty-impalas-care.md
```diff
@@ -0,0 +1,5 @@
+---
+"azure_event_hub": major
+---
+
+Add newly introduced retention_description field. Please note that the azurerm provider must be upgraded to at least version 4.37.0. Run a terraform init -upgrade to satisfy the constraint.
```
kin0992, line 5: **nitpick**: I would update the changelog with some style, like explained in [this other PR](https://github.com/pagopa/dx/pull/764#discussion_r2234965023)

infra/modules/azure_event_hub/main.tf
```diff
@@ -62,6 +62,11 @@ resource "azurerm_eventhub" "events" {
   resource_group_name = var.resource_group_name
   partition_count     = each.value.partitions
   message_retention   = each.value.message_retention_days
+
+  retention_description {
```
kin0992, line 66: **question**: since these changes do not change the contract with the end user (it adds just a new block with some pre-defined values), is this still a `major` update?

is the `major` bump related to the upgrade of the provider min version?

lucacavallaro: agree: if there are no changes in the API it should be a minor

christian-calabrese: Actually, the changes are 2:
- the minimum version of the azurerm provider
- a unique situation where the new field `retention_description.retention_time_in_hours` has the same behaviour of an existing field `message_retention`.  They are mutually exclusive, so for any decision we get it will cause a change in the first plan. 

What do you think?

kin0992: ok, my point was from the IC's point of view: if we merge this PR and create a new version, what does the IC have to do? does the IC have to change something (like remove a field, add a new property), or is it enough to upgrade the module's version?

if the IC has to change something (so the contract changed), then I think it should be a `major`, otherwise it could be a `minor` or even a `patch`.

If something changes in the plan, but the IC does not have to change how it has configured the module, then I would still not bump with a `major`, but perhaps I would document that behavior in the changelog.

What are your thoughts about?

Krusty93: IC will still face the issue even though they will not update the module to this version, as the `azurerm` provider version constraint of the latest public revision is `< 5.0`. Therefore, I suggest to release a new patch with the provider upper version constraint set to `< 4.37.0`, so they will be fine until an explicit provider upgrade.

I agree with @kin0992 comment: if ICs can do a **no-brain** apply we can release this change as `patch` then

mamu0: I also think a major version bump isn’t necessary in this case, because even if the underlying behavior changes, nothing changes for the user. Plus, this module isn’t widely used, so it’s likely that the minimum provider version increase won’t have any impact

christian-calabrese:
This issue has been recognized and fixed in the next version of the azurerm provider 4.38.0.
Evidences of the fix:

> azurerm_eventhub - fix perpetual diff with message_retention (https://github.com/hashicorp/terraform-provider-azurerm/issues/30169)

PR #771, opened 2025-07-29
Create release-azure-appsvc-v1.yaml

lucacavallaro:
This workflow replaces [web_app_deploy.yaml](https://github.com/pagopa/dx/blob/20ae77c2e2307034f8d72f5f3bc970c458148c7c/.github/workflows/web_app_deploy.yaml) and the other legacy workflows for Azure Webapps deployments (App Service, Function App)

### Changes from `web_app_deploy`:
1. The workflow file was renamed to follow our workflows naming convention*
2. It works with `pnpm`
3. It is turbo-aware, and calls directly `npx turbo build` instead `yarn build` or `pnpm build`
4. Does not enforce a specific package structure and does not bundle the files, it zips the entire content of the workspace to keep the artifact simple and reproducible in local environment
5. The auto approve of staging deployments is now the default
6. If `canary-monitor.sh` is present, it will do a canary deployment with incremental rollout

### Run examples

* [Staging+Production with canary deployment](https://github.com/pagopa/dx-playground/actions/runs/16776260294)
* [Staging+Production **without** canary depoyment](https://github.com/pagopa/dx-playground/actions/runs/16776439441)
* [Fail if no custom warmup is configured](https://github.com/pagopa/dx-playground/actions/runs/16776596419)
* [Production only if staging slot is not found](https://github.com/pagopa/dx-playground/actions/runs/16776792250)

[Full documentation, with migration guide](https://github.com/pagopa/dx/blob/be9be8344322cec207c4faa1de9f9d2d634d754a/apps/website/docs/pipelines/release-azure-appsvc.md)

Closes CES-1157
Depends on #711


pnpm-workspace.yaml
```diff
@@ -25,3 +25,5 @@ onlyBuiltDependencies:
 packageImportMethod: clone-or-copy
 
 preferWorkspacePackages: true
+
```
gunzip, line 28: suggestion: can you add a comment about the reason of these values?

.github/workflows/release-azure-webapp-v1.yaml
```diff
@@ -0,0 +1,268 @@
...
+        description: Environment where the artifact will be deployed.
+        type: string
+        required: true
+      use_staging_slot:
+        description: Use a staging slot for the deployment.
+        type: boolean
+        required: false
+        default: true
+      disable_auto_staging_deploy:
+        description: Disable automatic deployments to the staging slot.
```
gunzip, line 26: question: do you mean "automatic swaps"? I guess the code is deployed to the staging slot anyway when using `use_staging_slot`.

.github/workflows/release-azure-webapp-v1.yaml
```diff
@@ -0,0 +1,268 @@
...
+        type: boolean
+        required: false
+        default: false
+      use_private_agent:
+        description: Use a private agent to deploy the built artifact.
+        type: boolean
+        required: false
+        default: true
+      use_labels:
+        description: Use labels to start the right environment's GitHub runner. If use_labels is true, also use_private_agent must be set to true
```
gunzip, line 36: I'd try to be more clear than "the right environment's GitHub runner" (in which sense "the right" ?)

.github/workflows/release-azure-webapp-v1.yaml
```diff
@@ -0,0 +1,268 @@
...
+          fi
+
+          if [ "$PACKAGE_MANAGER" = "pnpm" ]; then
+            pnpm --filter $WORKSPACE_NAME deploy --prod bundle
+            cd bundle && zip -r $BUNDLE_NAME.zip .
+            echo "artifact-path=$(realpath $BUNDLE_NAME.zip)" >> $GITHUB_OUTPUT
+            exit 0
+          fi
+
+          yarn config set nodeLinker node-modules
```
gunzip, line 112: suggestion: I'd add i`f [ "$PACKAGE_MANAGER" = "yarn" ];` or use a switch case in order to make this code more explicit, early returns are not easy to spot and reason about

.github/workflows/release-azure-webapp-v1.yaml
gunzip: `auto_approve_staging_deploy` is a better name

apps/website/docs/pipelines/release-azure-webapp.md
```diff
@@ -0,0 +1,73 @@
+---
+sidebar_position: 50
+---
+
+# Deploying a Node.js Application to Azure
```
gunzip, line 5: question: is this the same page that will document container app deploy?

lucacavallaro: no, the container app deploy right now has a different workflow (`release-azure-containerapp-v1.yaml`)

gunzip: then it's maybe advisable to have a more specific title ie. "Deploying a Node.js Application to Azure App Service or Function" ?

.github/workflows/release-azure-webapp-v1.yaml
lucacavallaro: @gunzip I don't know either and I don't like it, but I decided that changing that is out of scope. `use_labels` and `override_labels` are present in almost all dx workflows. We need to simplify that, also because the `runs-on` property on jobs is unreadable.

.github/workflows/release-azure-webapp-v1.yaml
lucacavallaro: @gunzip I mean automatic deploy, not automatic swap. This workflow by default have a job that approves automatically the deploy to staging slot. 

In the previous version of the workflow (`web_app_deploy`) it was an _opt-in_ feature, not I changed it to be opt-out.

apps/website/docs/pipelines/release-azure-webapp.md
```diff
@@ -0,0 +1,73 @@
...
+
+## How It Works
+
+**Build and Deployment Overview:**
+
+- It installs the dependencies using `pnpm` or `yarn` and builds the application
+  using `turbo build` command
+- The build output is packaged into a zip artifact that contains the compiled
+  application code and all necessary dependencies in a `node_modules` directory.
+- The artifact is deployed to the specific Azure Web App using a staging slot
```
Krusty93, line 19:
```suggestion
- The artifact is deployed to the specific Azure Web App/Function App using a staging slot
```

apps/website/docs/pipelines/release-azure-webapp.md
```diff
@@ -0,0 +1,73 @@
...
+    uses: pagopa/dx/.github/workflows/release-azure-webapp-v1.yaml@main
+    name: Deploy My App
+    with:
+      workspace_name: "my-app"
+      environment: "app-prod"
+      resource_group_name: "my-resource-group"
+      web_app_name: "my-app-func-01"
+      # use_staging_slot: true # Optional: default is true
+      # use_private_agent: true # Optional: default is true
+      # disable_auto_staging_deploy: true # Optional: default is false
```
Krusty93, line 72: Devs usually don't know about optional parameters, so I like this approach. However, we've not adding them to the docs website to avoid forgotten/out of date doc

.github/workflows/release-azure-webapp-v1.yaml
```diff
@@ -0,0 +1,268 @@
...
+  RESOURCE_GROUP_NAME: ${{ inputs.resource_group_name }}
+  WEB_APP_NAME: ${{ inputs.web_app_name }}
+
+jobs:
+  build:
+    name: Build
+    runs-on: ubuntu-latest
+
+    env:
+      WORKSPACE_NAME: ${{ inputs.workspace_name }}
```
Krusty93, line 61: `BUNDLE_NAME` has the same value

.github/workflows/release-azure-webapp-v1.yaml
```diff
@@ -0,0 +1,268 @@
...
+
+      - name: Build
+        run: npx turbo build --filter $WORKSPACE_NAME
+
+      - name: Get workspace path
+        id: get-workspace-path
+        run: |
+          WORKSPACE_PATH=$(npx turbo ls --output=json |
+            jq -sr --arg NAME "$WORKSPACE_NAME" '.[].packages.items[] | select(.name == $NAME).path')
+          echo "path=$WORKSPACE_PATH" >> $GITHUB_OUTPUT
```
Krusty93, line 87:
```suggestion
	  echo "::notice::Workspace path: $WORKSPACE_PATH"
          echo "path=$WORKSPACE_PATH" >> $GITHUB_OUTPUT
```
or `::debug::`

.github/workflows/release-azure-webapp-v1.yaml
```diff
@@ -0,0 +1,268 @@
...
+      - name: Make artifact
+        id: make-artifact
+        env:
+          YARN_NODE_LINKER: node-modules
+          YARN_NM_HOISTING_LIMITS: workspaces
+          PACKAGE_MANAGER: ${{ steps.node-setup.outputs.package-manager }}
+        working-directory: ${{ steps.get-workspace-path.outputs.path }}
+        run: |
+          if grep -rq --include='next.config.*' 'output: "standalone"' .; then
+            echo "::debug:: the workspace contains a Next.js standalone app"
```
Krusty93, line 98:
```suggestion
            echo "::debug::the workspace contains a Next.js standalone app"
```

.github/workflows/release-azure-webapp-v1.yaml
lucacavallaro: good catch, I wanted to remove `BUNDLE_NAME` to keep only `WORKSPACE_NAME`

apps/website/docs/pipelines/release-azure-webapp.md
lucacavallaro: as decided a few hours ago, I will rename all "web app" references to "appsvc"

apps/website/docs/pipelines/release-azure-webapp.md
Krusty93: @lucacavallaro we can however discuss again the issue

.github/workflows/release-azure-webapp-v1.yaml
```diff
@@ -0,0 +1,268 @@
...
+        required: true
+      web_app_name:
+        description: Web App name.
+        type: string
+        required: true
+      environment:
+        description: Environment where the artifact will be deployed.
+        type: string
+        required: true
+      use_staging_slot:
```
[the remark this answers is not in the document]
Krusty93: `app-ci` should have enough permissions. If not, we can add them as we are **reading** information

.github/workflows/release-azure-webapp-v1.yaml
lucacavallaro: @Krusty93 I decided to keep different variable names because they have different roles despite the same content

.github/workflows/release-azure-appsvc-v1.yaml
```diff
@@ -0,0 +1,252 @@
...
+    runs-on: ubuntu-latest
+
+    env:
+      WORKSPACE_NAME: ${{ inputs.workspace_name }}
+
+    steps:
+      - name: Check out repository
+        uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2
+
+      - name: Setup Node.js
```
mamu0, line 67: **question**: I noticed in your workflow that you're operating from the repository root and calculating the workspace path as needed, instead of using turbo prune with a working-directory: ./out.

Was there a specific reason for this approach? Just curious to understand the pros and cons of each method.

.github/workflows/release-azure-webapp-v1.yaml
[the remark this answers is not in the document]
lucacavallaro: Done ✅ I've added a "validate" step that checks for "WARMUP" settings and slots availability

.github/workflows/release-azure-appsvc-v1.yaml
```diff
@@ -0,0 +1,252 @@
...
+        type: string
+        required: false
+        default: ""
+
+concurrency:
+  group: ${{ github.workflow }}-cd
+  cancel-in-progress: true
+
+env:
+  BUNDLE_NAME: ${{ inputs.workspace_name }}
```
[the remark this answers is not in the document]
mamu0: Got it! In that case, could you use a variable for both?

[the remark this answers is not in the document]
lucacavallaro: I prefer using different variables because they have different scopes, the content is just a detail. If we replace them with only one variable, we will couple two different part of the pipeline, right now they share just the same value but they are indipendent (if we change the BUNDLE_NAME we don't have to change the WEB_APP_NAME)

apps/website/docs/pipelines/release-azure-webapp.md
gunzip: I still see the "old" title

.github/workflows/release-azure-appsvc-v1.yaml
```diff
@@ -0,0 +1,299 @@
...
+        run: |
+          WORKSPACE_PATH=$(npx turbo ls --output=json |
+            jq -sr --arg NAME "$WORKSPACE_NAME" '.[].packages.items[] | select(.name == $NAME).path')
+          echo "path=$WORKSPACE_PATH" >> $GITHUB_OUTPUT
+
+      - name: Make artifact
+        id: make-artifact
+        env:
+          YARN_NODE_LINKER: node-modules
+          YARN_NM_HOISTING_LIMITS: workspaces
```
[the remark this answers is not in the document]
lucacavallaro: Good catch, removed ✅ They were used by `yarn`, but `yarn config set` made them redundant.

apps/website/docs/pipelines/release-azure-appsvc.md
```diff
@@ -0,0 +1,267 @@
...
+   the following permissions to your repository:
+
+   | Permission   | Access Level   |
+   | ------------ | -------------- |
+   | Actions      | Read           |
+   | Contents     | Read           |
+   | Deployments  | Read and write |
+   | Environments | Read           |
+
+   The PAT should be associated with a user (e.g., `pagopa-bot`) that has
```
Krusty93, line 104: **thought**: Theoretically, this key creates a circular dependency between `bootstrap` (kv secret) and `resources` (kv). So, this secret should be moved to the subscription-wide key vault, generally referred as `common` in our context. This KV is created by the `core` module, as it contains also the PAT for the GitHub runner. Maybe we could think about adding also this new PAT at the subscription bootstrapping phase? The only issue I can think of is its scope setup, but it might be addressed at the creation of a new repository
@pagopa/engineering-team-devex

Krusty93: As you have a single bot user for each product, it will end up in the same way

.github/workflows/release-azure-appsvc-v1.yaml
```diff
@@ -0,0 +1,297 @@
...
+
+      - name: Azure Login
+        uses: pagopa/dx/.github/actions/azure-login@main
+        env:
+          ARM_USE_OIDC: true
+          ARM_CLIENT_ID: ${{ secrets.ARM_CLIENT_ID }}
+          ARM_TENANT_ID: ${{ secrets.ARM_TENANT_ID }}
+          ARM_SUBSCRIPTION_ID: ${{ secrets.ARM_SUBSCRIPTION_ID }}
+
+      - name: Deploy
```
Krusty93, line 204: **praise**: cool!

Krusty93 (approved):

apps/website/docs/pipelines/release-azure-appsvc.md
lucacavallaro: Updated with `module.core_values.common_key_vault.id`

PR #777, opened 2025-07-30
Remove GitHub repository configuration from Azure Bootstrap module

christian-calabrese:
Remove GitHub repository configuration from the Azure GitHub Environment Bootstrap module to improve separation of concerns

Resolves CES-1178


apps/website/docs/conventions/infra-folder-structure.md
```diff
@@ -28,12 +38,16 @@ subfolder. Each configuration can be used for multiple CSP regions.
 
 ```
 infra/
-├─ identity/
+├─ bootstrapper/
 │  ├─ prod/
 │  │  ├─ main.tf
```
gunzip, line 43: question: in case we need to init multiple csp (mutiple identities for the github runner that let it operate on both azure and aws) they will both go within the same `main.tf `?

gunzip:
> ask to make a migration script

lets try. we can distribute a typescript gist to run using npx?

apps/website/docs/conventions/infra-folder-structure.md
gunzip: 1) is this decision blocking for this pr? 
2) are the proposed changed compatible with a multi-csp scenario?

Krusty93: What's the issue of keeping the same environment for multiple CSPs? Secret names are not the same.
Also, I think having multiple Terraform providers in the same `main.tf` is not a problem at all?

christian-calabrese: No, having multiple providers is not a problem at all. Of course, who applies the bootstrap will need to login to both csps, but that's not an issue neither.

The only issue I see is that one of the two bootstrap modules needs to create the environments, while the other will only reference them.
I approached this issue while developing the [AWS bootstrap](https://github.com/pagopa/dx/blob/ee63b0fd0aab4af158dad567702c300bce19c4f5/infra/modules/aws_github_environment_bootstrap/github_environments_cd.tf#L21) with a similar approach to the `repository.configure` flag (that we disliked and discarded).

lucacavallaro: @gunzip I think we should have "one state" per "tf workspace". So if we have a "production" workspace I expect to have only one state file stored in my primary CSP.

gunzip: > one of the two bootstrap modules needs to create the environments, while the other will only reference them.

what if we put the github enviroment creation (which is agnostic to csp) into repository module and enviroment configuration (which is csp aware) in the <csp>-github-enviroment-bootstrap?

christian-calabrese: true, but where do you envision the environments creation in a multi-environment repository?

christian-calabrese: @pagopa/engineering-team-devex  Isn't [this create_environments flag](https://github.com/pagopa/dx/blob/ee63b0fd0aab4af158dad567702c300bce19c4f5/infra/modules/aws_github_environment_bootstrap/github_environments_cd.tf#L21) easier than creating a new module and instantiate it in each repo together with the bootstrap?
99% of the times it's gonna be true. Maybe its name may be changed to reflect the fact that environments are already managed somewhere else (the other bootstrap instance)

Krusty93: `configure_github_environments`?

christian-calabrese: Yep, or otherwise a variable of the bootstrap module like `is_main_csp`

christian-calabrese: I think cause we usually create resources in their env folder (es. infra/resources/dev) and store them in the environment's state. This would be a little bit of an exception that we avoided in other occasions

mamu0: **Please note that this would need to be tested to be certain**, but from what I know, when creating a new secret to add to an environment, only the environment **name** is required and If the environment doesn’t exist, it’s automatically created with that name to store the secrets.
So, we could consider to simply let the bootstrap process add the new secrets to the specified environment, and if it exists, they’ll be added; if not, the environment will be created automatically.

christian-calabrese: Done!

.changeset/orange-trams-act.md
```diff
@@ -0,0 +1,153 @@
...
+# infra/bootstrapper/<env>/<env>.tf
+module "bootstrapper" {
+  source = "../_modules/bootstrapper"
+
+  environment = {
+    prefix          = "dx"
+    env_short       = "d"
+    location        = "italynorth"
+    domain          = "test"
+    app_name        = "monorepo_starter_pack"
```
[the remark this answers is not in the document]
christian-calabrese: True, it's been removed from the variable, but not from the tests and examples

apps/website/docs/conventions/infra-folder-structure.md
```diff
@@ -28,12 +38,16 @@ subfolder. Each configuration can be used for multiple CSP regions.
 
 ```
 infra/
```
Krusty93, line 40: I think it is not clear the dependency graph within the modules.
Moreover, can repository and bootstrapper be set in the same Terraform state?

infra/modules/azure_github_environment_bootstrap/outputs.tf
```diff
@@ -8,10 +8,9 @@ output "resource_group" {
 }
 
 output "repository" {
-  description = "Details of the GitHub repository, including ID and name."
+  description = "Details of the GitHub repository."
```
Krusty93, line 11:
```suggestion
  description = "GitHub repository name."
```

apps/website/docs/conventions/infra-folder-structure.md
christian-calabrese: What do you find it's not clear?

infra/modules/azure_github_environment_bootstrap/outputs.tf
```diff
@@ -8,10 +8,9 @@ output "resource_group" {
...
   value = {
-    id   = try(module.github_repository["repo"].id, null)
-    name = local.repository_name
+    name = var.repository.name
```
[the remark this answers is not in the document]
christian-calabrese: I don't understand the change, do you mean you'd like to make it become a simple string instead of an object?

apps/website/docs/conventions/infra-folder-structure.md
Krusty93: what module depends to what other module, then what configuration should be applied first

.changeset/orange-trams-act.md
```diff
@@ -0,0 +1,183 @@
...
+If you were relying on this module to manage your GitHub repository configuration, you must now use the dedicated GitHub module for repository management.
+
+## Migration Guide
+
+To migrate from the previous version, you need to split the configuration between two modules:
+
+### Before (v2.x.x)
+
+```hcl
+# infra/bootstrapper/<env>/<env>.tf
```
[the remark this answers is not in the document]
Krusty93: I think so, I don't see any value in having the (i.e.) `prod.tf` in `prod` folder, etc.

Krusty93 (approved):

PR #778, opened 2025-07-31
Add @pagopa/pnpm-config with basic configuration

lucacavallaro:


packages/pnpm-config/README.md
```diff
@@ -0,0 +1,40 @@
+# @pagopa/pnpm-config
+
+Default configuration for PagoPA DX projects using [pnpm](https://pnpm.io).
+
+## Default configuration
+
+- Sets [workspacePackagePatterns](https://pnpm.io/settings#workspacepackagepatterns) to `["apps/*", "packages/*"]`.
+- Sets [linkWorkspacePackages](https://pnpm.io/cli/run#linkworkspacepackages) to `true`.
+- Sets [packageImportMethod](https://pnpm.io/settings#hoistpattern) to `clone-or-copy`
+- Adds `dx` [catalog](https://pnpm.io/catalogs) which includes recommended version of the packages supported by the PagoPA DX toolchain. Here you can find the full [list of packages included](#).
```
gunzip: issue: it looks like there's a missing link

packages/pnpm-config/README.md
```diff
@@ -0,0 +1,40 @@
...
+{
+  // .....
+  "devDependencies": {
+    "eslint": "catalog:dx",
+    "vitest": "catalog:dx",
+  },
+}
+```
+
+## Installation
```
gunzip, line 26: suggestion: I'd put this first

packages/pnpm-config/README.md
```diff
@@ -0,0 +1,40 @@
...
+
+```
+pnpm add --config @pagopa/pnpm-config
+```
+
+This will add `@pagopa/pnpm-config` to the [configDependencies](https://pnpm.io/config-dependencies) field in your `pnpm-workspace.yaml`.
+
+### Use the pnpmfile
+
+To use the `pnpmfile.cjs` provided by `@pagopa/pnpm-config`, you need to add the following line to your `pnpm-workspace.yaml`:
```
gunzip, line 36: suggestion: I'd clarify what this step enables (what if I miss this? is this mandatory?)

packages/pnpm-config/README.md
lucacavallaro: already suggested by Copilot https://github.com/pagopa/dx/pull/778#discussion_r2245143099 - but I needed to push it first to obtain the link

packages/pnpm-config/README.md
lucacavallaro: I discovered that this step can be omitted using the `pnpm-plugin-*` prefix for config packages, so I renamed the workspace to `pnpm-config-pagopa`

actions/pr-comment/package.json
```diff
@@ -33,12 +33,12 @@
     "@pagopa/eslint-config": "workspace:^",
     "@tsconfig/node22": "catalog:",
     "@types/node": "catalog:",
-    "@vitest/coverage-v8": "catalog:",
-    "eslint": "catalog:",
+    "@vitest/coverage-v8": "catalog:dx",
+    "eslint": "catalog:dx",
```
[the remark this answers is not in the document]
lucacavallaro: Nope, the `dx` catalog is included in our plugin, `pnpm-plugin-pagopa`

Once installed as a config dependency, the plugin registers a hook ([`updateConfig`](https://github.com/pagopa/dx/blob/6dadf593545cf7554abc54ffba743482082d44d1/packages/pnpm-plugin-pagopa/pnpmfile.cjs#L3)) that modifies the base configuration of `pnpm-workspace.yaml`. I wrote the hook in a way that allows devs to override the settings if they want (for example, by default it registers `apps/*` and `packages/*` as workspace folders, but the consumers can change that if they have different needs).

I expect that in the typical case, the pnpm-workspace.yaml will contain only the import of pnpm-plugin-config and, at most, an additional catalog with common dependencies specific to that particular product.

packages/pnpm-plugin-pagopa/package.json
```diff
@@ -0,0 +1,12 @@
+{
+  "name": "pnpm-plugin-pagopa",
```
kin0992, line 2: **question**: is there a constraint on the naming?

since this is going to be published, shouldn't we use the `@pagopa` scope?

lucacavallaro: [Yes, there is a constraint on the naming.](https://pnpm.io/config-dependencies#:~:text=If%20your%20config%20dependency%20is%20named%20following%20the%20pnpm%2Dplugin%2D*%20pattern%2C%20pnpm%20will%20automatically%20load%20the%20pnpmfile.cjs%20from%20its%20root.). If we add the `@pagopa` scope, our consumers have to explicitly set `pnpmfile: node_modules/.config-dependencies/@pagopa/pnpm-config/pnpmfile.js` in their `pnpm-workspace.yaml`

PR #781, opened 2025-08-01
Fix infra_plan workflow removing hidden-links

mamu0:
- Fixed problem about `hidden-links` showed inside PR comment with plan result.
- Patch for `azure_infra_core` module to resolve typos.

Resolves: CES-1190


.github/workflows/infra_plan.yaml
```diff
@@ -145,13 +145,13 @@ jobs:
         id: plan
         working-directory: ${{ steps.directory.outputs.dir }}
         run: |
-          terraform plan -lock-timeout=3000s -no-color -out=plan.out 2>&1 | grep -v "hidden-link:" | tee tf_plan_stdout.txt
+          terraform plan -lock-timeout=3000s -no-color -out=plan.out 2>&1 | tee tf_plan_stdout.txt
```
gunzip, line 148: thoughts: I guess we should maintain this filter here since action logs are public as well

.changeset/fluffy-parts-doubt.md
```diff
@@ -0,0 +1,5 @@
+---
+"azure_core_infra": patch
+---
+
+Resolved typos inside examples
```
gunzip, line 5:
```suggestion
Correct the missing filter that causes sensitive keys to be exposed in GitHub comments during Terraform plan execution
```

christian-calabrese (approved):

infra/modules/azure_core_infra/example/complete/locals.tf
```diff
@@ -9,10 +9,11 @@ locals {
   }
 
   tags = {
-    CreatedBy   = "Terraform"
-    Environment = "Dev"
-    Owner       = "DevEx"
-    Source      = "https://github.com/pagopa/dx/modules/azure_core_infra/examples/complete"
-    CostCenter  = "TS700 - ENGINEERING"
+    CostCenter     = "TS000 - Tecnologia e Servizi"
```
gunzip, line 12: issue: is it possible to remove unrelated changes from this PR? we can fix these with anoher one and make this dependant

PR #785, opened 2025-08-01
Add GitHub Workflow for Static Web App build and deploy

mamu0:
This PR introduces a new reusable GitHub Actions workflow to build and deploy static sites to **Azure Static Web Apps**. It standardizes our deployment process and adds support for automated Pull Request previews.

Resolves: CES-1180

### DIFF
--- a/.github/workflows/release-typescript-static-website-deploy-v1.yaml
+++ b/.github/workflows/release-typescript-static-website-deploy-v1.yaml
@@ -131,6 +131,7 @@ jobs:
     permissions:
       id-token: write
       contents: read
+      pull-requests: write
 
     steps:
       - name: Checkout
--- a/.github/workflows/release-typescript-static-website-deploy-v1.yaml
+++ b/.github/workflows/release-typescript-static-website-deploy-v1.yaml
@@ -1,0 +1,170 @@
+name: Deploy Static Site to Azure
+
+# This reusable workflow builds a static application from a monorepo workspace,
+# uploads the assets to an Azure Static Web App, and deploys them.
+# It is designed to be called by other workflows.
+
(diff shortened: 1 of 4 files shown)


.github/workflows/release-typescript-static-website-deploy-v1.yaml
lucacavallaro, line 1: **suggestion:** we should rename it `release-azure-staticapp-v1.yaml` to keep it aligned with other workflows (`release-azure-containerapp`, `release-azure-appsvc`)

- we don't need to specify `typescript` because, since it uses `turbo`, this workflow can be adapter for other stacks as well
- we don't need te `-deploy` suffix, because we already have the `release` prefix
- `staticapp` is the name used by azure for this kind of resources, and it is shorter than `static-website`

apps/website/docs/pipelines/build-deploy-static-web-app.md
```diff
@@ -0,0 +1,92 @@
...
+   - Runs on a configurable runner (public or self-hosted) based on input
+     parameters.
+   - Downloads the build artifact.
+   - Logs into Azure using OIDC credentials configured as GitHub secrets.
+   - Retrieves the **deployment token** (API key) from the target Azure Static
+     Web App. This token is used to authorize the deployment.
+   - Calls the official `Azure/static-web-apps-deploy@v1` action to handle the
+     final deployment, uploading the pre-built application assets. This action
+     also automatically handles the creation and management of **preview
+     environments** for Pull Requests.
```
lucacavallaro, line 42: **suggestion:** this description is too detailed and need to be updated for every change on the action, we should keep it more high level

PR #797, opened 2025-08-06
Add codebuild runner

christian-calabrese:


.changeset/rotten-pumas-open.md
```diff
@@ -0,0 +1,5 @@
+---
+"aws_github_environment_bootstrap": minor
```
mamu0, line 2:
```suggestion
"aws_github_environment_bootstrap": patch
```

Actually, the module is still at version 0.0.0 and is not being used. I think we can avoid incrementing the minor version for now

infra/modules/aws_github_environment_bootstrap/variables.tf
```diff
@@ -9,6 +9,7 @@ variable "environment" {
     env_short       = string
     region          = string
     domain          = string
+    app_name        = string
```
mamu0, line 12: **question**: is it mandatory?

mamu0 (approved):
Great job! Just a few comments

infra/modules/aws_github_environment_bootstrap/variables.tf
christian-calabrese: The codebuild module requires it to build the name of resources. I'm not sure whether to remove it from the codebuild module (with a major) or keep it

PR #801, opened 2025-08-06
Add PAT ssm parameter in aws-core-infra and debug tags

christian-calabrese:
Needed to connect AWS Codebuild with GitHub


mamu0 (approved):
LGTM

PR #804, opened 2025-08-08
Add SBOMs and generation scripts

mamu0:
This PR introduces:

- Scripts to `generate` and `validate` SBOMs using [syft](https://github.com/anchore/syft) and [grype](https://github.com/anchore/grype)
- Integration of `syft` and `grype` into the `devcontainer`
- A new `sboms/` folder containing all generated **SBOM** files
- Pre-commit hooks to automatically run SBOM generation and validation

Each SBOM is generated per module, provider, or workspace to allow finer granularity, avoid duplication, and simplify future validation or comparison.

The two scripts are optimized to skip execution if no relevant changes have been made to the corresponding `module`, `provider`, or `pnpm` package.

An alternative implementation using a single unified SBOM file is available in the [feat-create-sbom-file-alternative](https://github.com/pagopa/dx/tree/feat-create-sbom-file-alternative) branch, but it may miss some components and lacks the same level of control.


.pre-commit-config.yaml
```diff
@@ -16,6 +16,18 @@ repos:
...
+        entry: infra/scripts/generate-sboms.sh
+        language: script
+        pass_filenames: false
+        verbose: true
+      - id: validate_sboms
+        name: Validate all SBOMs
+        entry: infra/scripts/validate-sboms.sh
+        language: script
+        pass_filenames: false
+        verbose: true
```
lucacavallaro, line 30: **suggestion**: I would not add these scripts to `pre-commit`. We don't need (and don't want) to generate sboms at every commit/change, but only when requested by our org.

The right place for this script is the root `package.json`

Krusty93: @lucacavallaro what's the process you suggest to manage SBOM generation? Agree to avoid to put them in pre commit

I think it could attached within produced artifacts (updating the attestation on GitHub maybe?), but as deps dynamically change, some sort of continuous monitoring should be put in place. Also, before deploying something the provenance should be verified

.devcontainer/features/syft/README.md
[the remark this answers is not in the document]
lucacavallaro: are these files generated by devcontainer cli (`devcontainer features generate-docs`) ?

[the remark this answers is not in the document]
mamu0: My apologies, I hadn’t understood what you meant. I’ve updated the code to use the existing features, there wasn’t any specific reason for not using them before!

.pre-commit-config.yaml
Krusty93: To me is ok to go ahead with this PR, but I suggest to think a workflow to actual use and exploit these files as a routine process

sbom.sh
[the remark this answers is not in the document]
kin0992: I would also update the README with the requirements to run the scripts. If I just run it as it is, without doing anything, I think I will have an error about a missing package (syft)

kin0992 (changes requested):

sbom.sh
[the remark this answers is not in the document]
mamu0: Thanks for the suggestions! I updated `package.json` to include the script commands and created a documentation page on the website. Please take a look!

apps/website/docs/infrastructure/sboms.md
```diff
@@ -0,0 +1,113 @@
...
+     the `infra/modules/` directory.
+3. **Stores the SBOMs**: All generated JSON files are stored inside the `sboms`
+   directory for easy access and validation.
+
+By default, the script will not regenerate an SBOM file if it already exists.
+You can override this behavior using the `--force-generate` flag.
+
+### Usage
+
+The script can be run directly or through Turborepo task aliases.
```
Krusty93, line 63: What these aliases are?

mamu0: What’s the official name used to refer to the aliases in `package.json`? I mean the ones defined inside `"scripts": {}`

mamu0: I’ve updated this part. I’m just calling them `tasks` now and specifying that they’re in the `package.json`. Let me know if this makes it clearer.

sbom.sh
```diff
@@ -0,0 +1,201 @@
...
+# HELP FUNCTION
+# ==============================================================================
+
+show_help() {
+    echo "Usage: $(basename "$0") [OPTION]"
+    echo "Manages the generation and validation of SBOMs."
+    echo ""
+    echo "Options:"
+    echo "  -g, --generate        Run the SBOM generation process (default action)."
+    echo "  -f, --force-generate  Force regeneration of all existing SBOM files."
```
lucacavallaro, line 21: **question:** why are there two different `generate` command? what is the different? can't be merged into one?

apps/website/docs/infrastructure/sboms.md
```diff
@@ -0,0 +1,112 @@
...
+   - A single `sbom-npm-workspace.json` is created for all TypeScript and
+     Node.js dependencies.
+   - An individual SBOM file is generated for each Go provider located in the
+     `providers/` directory.
+   - An individual SBOM file is generated for each Terraform module located in
+     the `infra/modules/` directory.
+3. **Stores the SBOMs**: All generated JSON files are stored inside the `sboms`
+   directory for easy access and validation.
+
+By default, the script will not regenerate an SBOM file if it already exists.
```
lucacavallaro, line 58: **question**: won't the script regenerate the SBOM file even if the dependencies are updated?

apps/website/docs/infrastructure/sboms.md
```diff
@@ -0,0 +1,112 @@
...
+
+  ```sh
+  sh ./sbom.sh --validate
+  ```
+
+- **To force regenerate all SBOMs**, overwriting existing ones:
+
+  ```sh
+  sh ./sbom.sh --force-generate
+  ```
```
lucacavallaro, line 89: **nitpick**: It seems redundant

apps/website/docs/infrastructure/sboms.md
```diff
@@ -0,0 +1,112 @@
...
+SBOMs are crucial for modern software development because they:
+
+- **Enhance Security**: By providing a clear view of all dependencies, you can
+  quickly identify components with known vulnerabilities.
+- **Ensure Compliance**: Many regulations and industry standards now require a
+  detailed software inventory.
+- **Improve Transparency**: They offer a clear and auditable record of what's
+  inside your software supply chain.
+
+## SBOM Management Script
```
lucacavallaro, line 23: **nitpick**: redundant

sbom.sh
mamu0: The `--generate` command creates only **SBOMs** that don’t already exist. So, for example, if some **SBOMs** have already been generated and new resources are added, running it again will generate only the missing ones, speeding up the process. If you want to regenerate everything and replace even the existing **SBOMs**, you can use `--force-generate`.

apps/website/docs/infrastructure/sboms.md
mamu0: When it ran as a `pre-commit` hook, it only regenerated SBOMs that had changed (modules and providers). I could add similar logic, but now as a regular script I can’t assume it’s run on a `non-main` branch. Alternatively, we could always use `--force-generate` to refresh all SBOMs regardless.

apps/website/docs/infrastructure/sboms.md
mamu0: What exactly? It has three commands, which is why there are three points

sbom.sh
lucacavallaro: I think it is the safest option to keep only the "force" version, because if dependencies are not considered by the tool, we risk having outdated files.

apps/website/docs/infrastructure/sboms.md
lucacavallaro: I think we can keep only

```
An SBOM (Software Bill of Materials) is a formal, machine-readable inventory of
software components and dependencies contained in an application. It provides a
detailed list of all the libraries, frameworks, and modules that make up the
software, along with their versions and licenses.
```

sbom.sh
mamu0: You’re not wrong, but if we plan to automate the execution, we’ll probably need something that regenerates only the modified or missing ones instead of everything every time. So I’d say we could make the current `--force-generate` the default `--generate`, and give a different name to the command that only handles new ones, something like `--soft-generate` or similar

kin0992 (approved):

PR #805, opened 2025-08-11
Add reusable GitHub action for monorepo builds

mamu0:
This PR introduces a new reusable action, build-monorepo-workspace, to centralize the build and packaging logic for workspaces within our monorepo.

The following workflows have been refactored to use this new action, significantly reducing code duplication and simplifying their structure:

- `release-azure-appsvc-v1.yaml`
- `release-azure-staticapp-v1.yaml`
- `release-typescript-cdn-deploy-v1.yaml` -> `renamed:release-azure-cdn-v1.yaml`

This change makes our CI/CD pipelines easier to maintain, as the core build logic is now managed in a single place.

Resolves: CES-1195

### DIFF
--- a/.github/workflows/release-azure-appsvc-v1.yaml
+++ b/.github/workflows/release-azure-appsvc-v1.yaml
@@ -66,18 +66,11 @@ jobs:
         with:
           workspace_name: ${{ env.WORKSPACE_NAME }}
 
-      - name: Get workspace path
-        id: get-workspace-path
-        run: |
-          WORKSPACE_PATH=$(npx turbo ls --output=json |
-            jq -sr --arg NAME "$WORKSPACE_NAME" '.[].packages.items[] | select(.name == $NAME).path')
-          echo "path=$WORKSPACE_PATH" >> $GITHUB_OUTPUT
-
       - name: Make artifact
         id: make-artifact
         env:
           PACKAGE_MANAGER: ${{ steps.build-package.outputs.package-manager }}
-        working-directory: ${{ steps.get-workspace-path.outputs.path }}
+        working-directory: ${{ steps.build-package.outputs.workspace-path }}
         run: |
           if grep -rq --include='next.config.*' 'output: "standalone"' .; then
             echo "::debug::The workspace contains a Next.js standalone app"
--- a/.github/workflows/release-azure-staticapp-v1.yaml
+++ b/.github/workflows/release-azure-staticapp-v1.yaml
@@ -86,15 +86,12 @@ jobs:
       
       - name: Get Artifact path
         id: get-artifact-path
+        working-directory: ${{ steps.build-package.outputs.workspace-path }}
         run: |
-          WORKSPACE_PATH=$(npx turbo ls --output=json | jq -sr --arg NAME "$WORKSPACE_NAME" '.[].packages.items[] | select(.name == $NAME).path')
-          cd $WORKSPACE_PATH
-
           echo "::debug::Packaging for static deployment"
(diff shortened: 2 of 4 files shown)


.github/actions/build-monorepo-workspace/action.yaml
```diff
@@ -0,0 +1,102 @@
...
+  workspace_name:
+    description: 'The name of the workspace to build.'
+    required: true
+  packaging_method:
+    description: "The packaging method to use: 'app-service' for zipped standalone apps, or 'static' for static site output directories."
+    required: true
+  output_dir:
+    description: 'For static packaging, the name of the build output directory.'
+    required: false
+    default: 'dist'
```
lucacavallaro, line 15: In my opinion this action should just build the project and not package it. The packaging phase is a detail for the workflow.

So it should contain all the steps until `Build`

mamu0: Are you sure the artifact creation part always varies significantly between workflows? Because by adding a dedicated step here, the workflow configuration would be much simpler.

_What I mean is_: if the differences between workflows are minor, it might be more convenient to have the artifact creation step here. But if they are indeed always completely different, then I agree with you and each workflow can have its own step directly.

lucacavallaro: yes, the artifact creation steps are different:

- On static projects like nextjs, `build` is enough to produce the `dist` folder to deploy
- Only on App Services projects we need to produce ZIP bundles
- On other workflows we have to build OCI images

.github/workflows/release-azure-appsvc-v1.yaml
```diff
@@ -60,20 +60,11 @@ jobs:
...
-
-      - name: Install dependencies (yarn)
-        if: ${{ steps.node-setup.outputs.package-manager == 'yarn' }}
-        run: yarn install --immutable
-
-      - name: Build
-        run: npx turbo build --filter $WORKSPACE_NAME
+      - name: Build Package
+        id: build-package
+        uses: pagopa/dx/.github/actions/build-monorepo-workspace@main
```
lucacavallaro, line 65: **suggestion:** rename to `build-workspace`, all our abstractions are for monorepo, it is redundant! 😄

lucacavallaro (commented):
@mamu0 the change looks good, have you tested the updated workflow in the playground repo?

.github/workflows/release-azure-staticapp-v1.yaml
```diff
@@ -87,10 +87,14 @@ jobs:
       - name: Get Artifact path
         id: get-artifact-path
         run: |
+          WORKSPACE_PATH=$(npx turbo ls --output=json | jq -sr --arg NAME "$WORKSPACE_NAME" '.[].packages.items[] | select(.name == $NAME).path')
+          cd $WORKSPACE_PATH
```
lucacavallaro, line 91: **question:** what is the purpose of `cd $WORKSPACE_PATH` here?

mamu0: is to use properly `realpath`, in this way found the right path of the directory `OUTPUT_DIR`

lucacavallaro: I think that `Get Artifact path` can be useful in the reusable workflow (exported as `workspace-path`).

Then in `staticapp-v1` you can run a step with `working-dir: workspace-path` that exports just the `artifact-path`

mamu0: done 👍

lucacavallaro (approved):
LGTM, but please test each workflow in the playground before merging

PR #807, opened 2025-08-11
Replaced `tier` variable with `use_case` for Cosmos

mamu0:
Replaced the `tier` variable with a new `use_case` variable for tiering configuration. New values are:

- development
- default

Resolves: CES-1208

### DIFF
--- a/.changeset/cute-brooms-agree.md
+++ b/.changeset/cute-brooms-agree.md
@@ -1,0 +1,5 @@
+---
+"azure_cosmos_account": major
+---
+
+Replaced the `tier` variable with a new `use_case` variable for tiering configuration. The previous values (`s`, `l`) have been replaced by three new options: `development` (formerly `s`), and `default` (formerly `l`). This change simplifies and clarifies the selection of Cosmos Account tiers.
--- a/infra/modules/azure_cosmos_account/README.md
+++ b/infra/modules/azure_cosmos_account/README.md
@@ -14,12 +14,12 @@ This Terraform module provisions an Azure Cosmos DB Account with configurable se
 - **Serverless Mode**: Enables serverless mode for cost-efficient, on-demand scaling.
 - **Role Assignment**: Assigns SQL role permissions (Reader or Contributor) to specified principal IDs, enabling fine-grained access control.
 
-## Tiers and Configurations
+## Use cases and Configurations
 
-| Tier  | Description                                                                                        | Serverless Mode |
-|-------|----------------------------------------------------------------------------------------------------|-----------------|
-| `s`   | Recommended for development or testing environments where cost efficiency and flexibility are key. | Enabled         |
-| `l`   | Suitable for production environments requiring predictable performance and provisioned throughput. | Disabled        |
+| Use case      | Description                                                                                        | Serverless Mode |
+|---------------|----------------------------------------------------------------------------------------------------|-----------------|
+| `development` | Recommended for development or testing environments where cost efficiency and flexibility are key. | Enabled         |
+| `default`     | Suitable for production environments requiring predictable performance and provisioned throughput. | Disabled        |
 
 ## Usage Example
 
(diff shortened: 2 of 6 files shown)


infra/modules/azure_cosmos_account/variables.tf
```diff
@@ -28,16 +28,15 @@ variable "subnet_pep_id" {
   description = "The ID of the subnet designated for private endpoints."
 }
 
-variable "tier" {
```
lucacavallaro, line 31: **question:** it is not possible to keep this variable (and mark the change as "minor")? it would make the migration smoother

mamu0: In the specific case of Cosmos, it’s indeed one of the very few, if not the only, resources where the use cases remain the same as before. For the others, there are structural changes that would make the code more complex and harder to maintain if we tried to preserve backward compatibility.

That said, I believe making this a `minor` would be counterproductive. The variable has been replaced, which in itself is a breaking change: some infra, documentation, tests, and use cases all need to be updated accordingly. Marking it as a `major` makes it clearer and safer, since consumers will explicitly update their code against the new version. In fact, this is more consistent with semantic versioning, as it reflects a real improvement rather than a transparent patch.

For future changes, I’d see `minors` more as the right space for adjustments or additions/removals of use cases.

lucacavallaro: I don't agree. Tagging it `minor` right now, with backward compatibility, doesn't exclude a subsequent `major` version with legacy code removed.

It is an updating pattern used by every major tool:

1. Introduce a new feature as "experimental", or if stable, make it the "informal" standard and add a "deprecation notice" to the old variable
2. Give time for the consumer base to update, leave feedback, etc.
3. Iterate on the same `minor` branch
4. Finally, release a new `major` with deprecated code removed

Our user base is currently struggling to keep their modules updated, so we should think of ways to make this change smoother. The risk I see is that no one will update to the latest major version!

---

I would tag a new `major` only if the backward compatibility is too difficult to implement. It is the case?

Are there other strategies that we can apply to make this migration smoother?

infra/modules/azure_cosmos_account/README.md
```diff
@@ -14,12 +14,12 @@ This Terraform module provisions an Azure Cosmos DB Account with configurable se
...
-## Tiers and Configurations
+## Use cases and Configurations
 
-| Tier  | Description                                                                                        | Serverless Mode |
-|-------|----------------------------------------------------------------------------------------------------|-----------------|
-| `s`   | Recommended for development or testing environments where cost efficiency and flexibility are key. | Enabled         |
-| `l`   | Suitable for production environments requiring predictable performance and provisioned throughput. | Disabled        |
+| Use case      | Description                                                                                        | Serverless Mode |
+|---------------|----------------------------------------------------------------------------------------------------|-----------------|
+| `development` | Recommended for development or testing environments where cost efficiency and flexibility are key. | Enabled         |
```
Krusty93, line 21:
```suggestion
| `development` | Recommended for development or testing environments where cost efficiency and flexibility are key. Do not use in production. | Enabled         |
```

infra/modules/azure_cosmos_account/variables.tf
Krusty93: I believe we need to agree to a sort of migration plan first, which primarily involves the semver management. I can take the ownership of writing an RFC

mamu0: I see that Cosmos is currently at version `0.1.0` and not used much (only in two projects). To avoid extra steps that would slow things down, we could bump it as a minor to `0.2.0`, and apply the same approach to other modules in a similar situation

mamu0: Good! So we can use also a `patch`

.changeset/cute-brooms-agree.md
```diff
@@ -0,0 +1,44 @@
+---
+"azure_cosmos_account": patch
```
Krusty93, line 2:
```suggestion
"azure_cosmos_account": minor
```

Krusty93 (approved):

PR #808, opened 2025-08-11
Replace `tier` variable with `use_case` for Posgres

mamu0:

- default

Add new variables for replica:

- `create_replica`: Set to `true` by default, this determines whether or not to create a replica.
- `replica_location`: optional value used to specify a different location for the replica

Resolves: CES-1209


infra/modules/azure_postgres_server/monitor.tf
```diff
@@ -43,9 +43,4 @@ resource "azurerm_monitor_diagnostic_setting" "this" {
   enabled_log {
     category = "PostgreSQLLogs"
   }
-
```
Krusty93, line 46: **question**: why this change?

infra/modules/azure_postgres_server/locals.tf
```diff
@@ -10,27 +10,30 @@ locals {
...
       },
-      var.tier,
-      "GP_Standard_D2ds_v5" # Default
+      var.use_case,
+      "GP_Standard_D2ds_v5"
     )
   }
 
+  replica = {
+    create   = var.use_case != "development" && var.create_replica == true
```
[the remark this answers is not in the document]
mamu0: It wasn’t a mistake, I added it proactively for a possible `development` use case. For now it has no impact, since the condition will never be triggered. If it’s confusing or useless, I can remove it!

infra/modules/azure_postgres_server/monitor.tf
mamu0: I think it's `prettier`'s fault 🫤

infra/modules/azure_postgres_server/replica.tf
```diff
@@ -45,7 +45,7 @@ resource "azurerm_postgresql_flexible_server_virtual_endpoint" "endpoint" {
 #-----------------------------#
 
 resource "azurerm_postgresql_flexible_server_configuration" "pgbouncer_replica" {
-  count = var.tier == "l" && var.pgbouncer_enabled ? 1 : 0
+  count = local.replica.create == true && var.pgbouncer_enabled ? 1 : 0
```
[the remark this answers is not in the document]
mamu0: It is not automatic so when `pgbouncer_enabled` is `false`. If you meant whether there are cases in which it should be disabled, I don't know, but as far as I know, it is always enabled (and it is enabled by default).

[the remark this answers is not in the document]
Krusty93: Does it worth to remove the variable and keeping the feature enabled?

infra/modules/azure_postgres_server/variables.tf
```diff
@@ -113,6 +113,18 @@ variable "replica_zone" {
...
 
+variable "create_replica" {
+  type        = bool
+  description = "Indicates whether a replica PostgreSQL Flexible Server should be created. Defaults to true."
+  default     = true
+}
+
+variable "replica_location" {
+  type        = string
+  description = "The location where the replica PostgreSQL Flexible Server should be created. Defaults to the same as the primary server."
```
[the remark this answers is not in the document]
mamu0: Keep in mind now, the replication was always configured to be created in the same region (which didn’t make much sense). Changing the `default` now would be a change in logic, and since it’s a `major`, it makes sense. Is there a reason why Spain Central is preferred as the default instead of, for example, West Europe?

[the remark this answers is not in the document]
mamu0: On second thought, using the same region isn’t really a problem, because you improve high availability, while using a different region improves disaster recovery. So we should decide which of the two we want as the default

infra/modules/azure_postgres_server/variables.tf
```diff
@@ -113,6 +113,18 @@ variable "replica_zone" {
...
+variable "create_replica" {
+  type        = bool
+  description = "Indicates whether a replica PostgreSQL Flexible Server should be created. Defaults to true."
+  default     = true
+}
+
+variable "replica_location" {
+  type        = string
+  description = "The location where the replica PostgreSQL Flexible Server should be created. Defaults to another region to improve Disaster Recovery."
+  default     = "spaincentral"
```
Krusty93, line 125: **thought**: I think we can remove the default option while adding a validation rule to ensure the value is different from `environment.location`

infra/modules/azure_postgres_server/replica.tf
Krusty93, line 48:
```suggestion
  count = local.replica.create && var.pgbouncer_enabled ? 1 : 0
```

infra/modules/azure_postgres_server/locals.tf
```diff
@@ -10,27 +10,29 @@ locals {
...
+
   # Backup
   # Geo redundant backup is not available in Italy North
   # ZoneRedundant HA is not available in West Europe
-  geo_redundant_backup_enabled = (var.tier == "m" || var.tier == "l") && lower(var.environment.location) != "italynorth"
-  high_availability_enabled    = var.high_availability_override ? true : (var.tier == "m" || var.tier == "l") && lower(var.environment.location) != "westeurope"
-  auto_grow_enabled            = var.tier == "m" || var.tier == "l"
+  geo_redundant_backup_enabled = (var.use_case == "default") && lower(var.environment.location) != "italynorth"
+  high_availability_enabled    = var.high_availability_override ? true : (var.use_case == "default") && lower(var.environment.location) != "westeurope"
+  auto_grow_enabled            = var.use_case == "default"
```
Krusty93, line 33: ditto. the property can be set to the resource
```suggestion
```

infra/modules/azure_postgres_server/locals.tf
```diff
@@ -10,27 +10,29 @@ locals {
...
 
   # Monitoring
   metric_alerts         = var.custom_metric_alerts != null ? var.custom_metric_alerts : var.default_metric_alerts
-  replica_metric_alerts = var.tier == "l" ? local.metric_alerts : {}
+  replica_metric_alerts = var.use_case == "default" && local.replica.create ? local.metric_alerts : {}
```
Krusty93, line 37: ditto
```suggestion
  replica_metric_alerts = local.replica.create ? local.metric_alerts : {}
```

infra/modules/azure_postgres_server/variables.tf
mamu0: Just to clarify: at the beginning we agreed on changing the default from primary region to Spain Central for DR purposes. Now you’re suggesting to remove the default completely and make it mandatory with a validation rule, right?

Krusty93 (approved):

PR #823, opened 2025-09-01
Address VPN and NAT Gateway names created by Azure core module

Krusty93:
Close CES-1204

### DIFF
--- a/.changeset/vast-bushes-battle.md
+++ b/.changeset/vast-bushes-battle.md
@@ -1,0 +1,5 @@
+---
+"azure_core_infra": patch
+---
+
+Address an issue with NAT Gateway and VPN Gateway generated names
--- a/infra/modules/azure_core_infra/tests/common.tftest.hcl
+++ b/infra/modules/azure_core_infra/tests/common.tftest.hcl
@@ -15,14 +15,7 @@ run "core_is_correct_plan" {
   command = plan
 
   variables {
-    environment = {
-      prefix          = "dx"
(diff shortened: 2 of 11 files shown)


.changeset/vast-bushes-battle.md
```diff
@@ -0,0 +1,5 @@
+---
+"azure_core_infra": patch
+---
+
+Address an issue with NAT Gateway and VPN Gateway generated names
```
kin0992, line 5: **note**: what kind of issue?

I think it would be useful to add some additional information about the bug/issue, so if any users are going to search for that issue in the changelogs, it can find it more easily

infra/modules/azure_core_infra/tests/common.tftest.hcl
```diff
@@ -53,8 +46,18 @@ run "core_is_correct_plan" {
...
+  }
+
+  assert {
+    condition     = module.network.pep_snet.name == "dx-d-itn-pep-snet-${run.setup_tests.environment.instance_number}"
+    error_message = "Pep subnet name is not correct"
+  }
+
+  assert {
+    condition     = module.nat_gateway[0].nat_gateways[0].name == "dx-d-itn-ng-${run.setup_tests.environment.instance_number}"
+    error_message = "The NET Gateway name configuration must be correct"
```
kin0992, line 60: NET or NAT?
```suggestion
    error_message = "The NAT Gateway name configuration must be correct"
```

PR #824, opened 2025-09-01
Add editorconfig template

kin0992:
Closes CES-1232


packages/monorepo-generator/templates/monorepo/.editorconfig
```diff
@@ -0,0 +1,7 @@
+[*]
+charset = utf-8
+insert_final_newline = true
+end_of_line = lf
+indent_style = space
+indent_size = 2
+max_line_length = 80
```
Krusty93, line 7: I suggest to add:

```md
[*]
trim_trailing_whitespace = true
```

lucacavallaro: @Krusty93 this `editorconfig` file was copied by the [suggested one by prettier](https://prettier.io/docs/configuration#:~:text=Here%E2%80%99s%20a%20copy%2Bpaste%2Dready%20.editorconfig%20file%20if%20you%20use%20the%20default%20options%3A), that mimic the same rules of the tool.

we can add other rules, but they need to be scoped to a specific language and should not have conflicts with the configured formatter (`prettier` for js/ts/yaml, `go fmt` for go, `terraform fmt` for tf)

Krusty93: It's actually identical to the one we use for this repository but for this line. I don't think it'll cause any problem as trailing whitespaces are generally part of any language syntax.

P.S. From the Prettier resource you sent:
```md
# Caveat: Prettier won’t trim trailing whitespace inside template strings, but your editor might.
# trim_trailing_whitespace = true
```
This property might help us to have consistent behavior on a gap let by Prettier

PR #828, opened 2025-09-01
Add support for multiple action groups to Function App module

Krusty93:
Close CES-1221


.changeset/late-nails-cough.md
```diff
@@ -0,0 +1,5 @@
+---
+"azure_function_app": minor
```
gunzip, line 2: isn't this a breaking change?

PR #838, opened 2025-09-03
Create function to generate a monorepo scaffold

kin0992:
Closes CES-1236

### DIFF
--- a/.changeset/clean-penguins-begin.md
+++ b/.changeset/clean-penguins-begin.md
@@ -2,4 +2,4 @@
 "@pagopa/monorepo-generator": minor
 ---
 
-Create a function to generate a monorepo scaffold
+Create a `plop` generator to scaffold a new monorepo


packages/monorepo-generator/tsup.config.ts
```diff
@@ -0,0 +1,15 @@
...
+
+export default defineConfig({
+  bundle: true,
+  clean: true,
+  dts: true,
+  entry: ["src", "!src/index.ts"],
+  external: ["plop"],
+  format: ["esm", "cjs"],
+  platform: "node",
+  publicDir: "./templates",
```
kin0992, line 11: **note**: this copies files from the specified `publicDir` to the `output` (`dist` in our case)

packages/monorepo-generator/tsup.config.ts
lucacavallaro, line 1: **suggestion(non-blocking):** I think we should use js instead of ts for this config file. The outcome is the same, but using js we don't need the `exclude` field in `tsconfig.js`

packages/monorepo-generator/README.md
```diff
@@ -1,11 +1,18 @@
...
+> Work in progress — expect improvements to templates and prompts.
 
+## Recommended usage (from repository root)
+
+Run the package script from the monorepo root so workspace dependencies and filtered scripts are resolved:
+
+```sh
+# run the package script which builds and runs plop
+pnpm --filter @pagopa/monorepo-generator run generate
+```
```
lucacavallaro, line 18: **thought**: since it is a plop generator I was expecting to use it with `turbo gen` or `plop` directly and not via a custom `generate` script

pnpm-workspace.yaml
```diff
@@ -11,7 +11,7 @@ catalog:
   '@tsconfig/node22': 22.0.2
   '@types/node': ^22.16.2
   prettier: 3.6.2
-  tsup: ^8.5.0
+  tsup: 8.5.0
```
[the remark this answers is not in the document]
kin0992: I don't know, I think `pnpm` did something when I installed `tsup`.
Anyway, I will revert this change

packages/monorepo-generator/src/plopfile.ts
lucacavallaro, line 1: **suggestion**: `plop` and its `plopfile.ts` should be managed outside the generator package. In my opinion this package should export only the generator giving the consumer the ability to include in its own configuration. 

The typical use of `plop` consist to have only one centralized plopfile per repo with multiple registered generators.

In our scenario, `dx generate <name>` command will wrap this logic to give consumers another way to use generators without setting up `plop` in their repo.

PR #857, opened 2025-09-09
Add codemod command

lucacavallaro:
[Read this comment for the Pull Request Overview](https://github.com/pagopa/dx/pull/857#pullrequestreview-3200844721)

⚠️ **THIS IS NOT A STABLE RELEASE**

The command outputs needs to be integrated with our logging stack, for now it just uses `console.log` and the built-in abstractions of `commander.js`.

I think that we need to unify/refactor some code to handle errors in a common way - so I decided to push these changes first behind a feature flag without changing other parts of the codebase.

👃 **CODE SMELL**

I added `CliDependencies` to `Dependencies` object in the `domain` layer. This is wrong because the domain should be agnostic to the `ui` layer. We need to refactor the dependency management to be placed in the infra layer.

Closes CES-1262

### DIFF
--- a/apps/cli/src/domain/__tests__/data.ts
+++ b/apps/cli/src/domain/__tests__/data.ts
@@ -28,7 +28,7 @@ export const makeMockPackageJson = (
 };
 
 export const makeMockDependencies = () => ({
-  applyCodemod: vi.fn(),
+  applyCodemodById: vi.fn(),
   listCodemods: vi.fn(),
   packageJson: makeMockPackageJson(),
   packageJsonReader: mock<PackageJsonReader>(),


apps/cli/src/use-cases/apply-codemod.ts
```diff
@@ -0,0 +1,17 @@
+import { errAsync, okAsync, ResultAsync } from "neverthrow";
+
+import { CodemodRegistry } from "../domain/codemod.js";
+
+export type ApplyCodemod = (id: string) => ResultAsync<void, Error>;
+
+export const applyCodemod =
```
kin0992, line 7: **suggestion(non-blocking)**: maybe we could call `findAndApplyCodemod` or, if we want to keep the same name, we should pass the codemod itself, not its id

apps/cli/src/use-cases/apply-codemod.ts
```diff
@@ -0,0 +1,17 @@
...
+  (registry: CodemodRegistry): ApplyCodemod =>
+  (id: string) =>
```
kin0992, line 9: **nitpick**: param type is redundant
```suggestion
  (id) =>
```

kin0992 (changes requested):
the changelog is missing

apps/cli/src/use-cases/apply-codemod.ts
lucacavallaro: I would like to rename it to `applyCodemodById`. Does that sounds good to you?

apps/cli/src/adapters/codemods/registry.ts
```diff
@@ -0,0 +1,19 @@
...
+  add(codemod: Codemod) {
+    this.#m.set(codemod.id, codemod);
+  }
+  getAll() {
+    return okAsync(Array.from(this.#m.values()));
+  }
+  getById(id: string) {
+    return okAsync(this.#m.get(id));
+  }
+}
```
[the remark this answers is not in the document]
kin0992: ok, thanks for sharing that information. as I said, it was a curiosity, nothing that I would change or something else :) so, for me, we could keep the class and evaluate later if refactor the existing code just to have consistency

apps/cli/src/use-cases/__tests__/apply-codemod.test.ts
```diff
@@ -0,0 +1,87 @@
...
+      description: "test codemod",
+      id: "foo",
+    };
+
+    const registry: CodemodRegistry = {
+      getAll: vi.fn(),
+      getById: vi
+        .fn()
+        .mockReturnValue(okAsync<Codemod | undefined, Error>(codemod)),
+    };
```
kin0992, line 23: **nitpick**: I think here you could use vitest-mock-extended to mock a type (like `CodemodRegistry`), instruct the mock in a single test, and move this boilerplate, which is duplicated in other tests, in a separate file (like we already did somewhere else with the `makeMock...Fn`.

kin0992 (approved):
I left a comment, which is non-blocking! Nice job though!

PR #858, opened 2025-09-09
Fix APIM module with Private Endpoint support for StandardV2

mamu0:
Resolves the issue where **APIM** cannot be created correctly when `use_case = cost_optimized` is selected.
To do this, the _virtual network type_ become `External` and vnet configuration will be done only for this case as it is not supported by **StandardV2**, and the `private endpoint` is created.

Resolves: CES-1259


infra/modules/azure_api_management/locals.tf
```diff
@@ -13,21 +13,48 @@ locals {
...
+      private_endpoint     = true
+      zones                = null
+    }
+    high_load = {
+      sku                  = "Premium_2"
+      virtual_network_type = "Internal"
+      public_network       = true
+      autoscale            = true
+      alerts               = true
+      private_endpoint     = false
```
gunzip, line 49: question: why private endpoints are disabled for high load tier?

infra/modules/azure_api_management/locals.tf
```diff
@@ -13,21 +13,48 @@ locals {
...
+      public_network       = false
+      autoscale            = false
+      alerts               = true
+      private_endpoint     = true
+      zones                = null
+    }
+    high_load = {
+      sku                  = "Premium_2"
+      virtual_network_type = "Internal"
+      public_network       = true
```
gunzip, line 46: question: why public_network is true for high load?

mamu0: This is the configuration it has always had, I haven’t changed anything to avoid introducing breaking changes. The PR is only meant to fix an incompatibility issue with `cost_optimized` use case.

christian-calabrese (approved):

PR #863, opened 2025-09-10
Features/scaffold terraform repository configuration

kin0992:
Now, the `@pagopa/monorepo-generator` generates Terraform files to manage the GitHub repository during scaffolding. The implementation includes new plop actions to fetch the latest Terraform provider and module versions from GitHub, updates to the generator's documentation, and the addition of required dependencies

Closes CES-1266


packages/monorepo-generator/src/actions/terraform.ts
```diff
@@ -0,0 +1,62 @@
...
+        answers[answerKey] = `${major}.${minor}`;
+        return `Fetched latest version: ${answers[answerKey]}`;
+      },
+      ({ message }) => {
+        // eslint-disable-next-line no-console
+        console.warn(`Could not fetch latest version`);
+        // Plop handles the Promise rejection or exception thrown as a failure
+        throw new Error(message);
+      },
+    );
```
lucacavallaro, line 36: **suggestion:** if this block is not intended to have recoverable errors (such as a fallback value) it is better to avoid using `Result` or refactor it to separate the different parts of the code.

In general, `match` should be used as alternative to `unwrapOr` to return the same value in each branch (like a map for both cases).

Example 1 (using Promise without Result)

```js
try {
  const { major, minor } = await fetchSemverFn();
  answers[answerKey] = `${major}.${minor}`
} catch (e) {
  console.warn(`Could not fetch latest version`);
  throw new Error("Could not fetch latest version", { cause: e });
}
```

Example 2 (using both Result and Promise)

```js
const version = await fetchSemverFn()
    .andThen((semver) =>
      semver ? okAsync(semver) : errAsync(new Error("Invalid version found")),
    )
    .map(({ major, minor }) => `${major}.${minor}`);
if (version.isErr()) {
    console.warn(`Could not fetch latest version`);
  throw new Error("Could not fetch latest version", { cause: version.error });
}
answers[answerKey] = version.value;

```

kin0992: That's even better! I updated the code accordingly

gunzip (commented):
lgtm

PR #866, opened 2025-09-11
Add GitHub Action for filtered Terraform Plan

mamu0:
This PR introduces a new, reusable GitHub Action designed to securely run `terraform plan` and prevent sensitive data from being exposed in logs or outputs.
It integrates logic previously in the `infra_plan.yaml` workflow to remove unnecessary logs, shortening the output. 
The action includes options to lsit sensitive key to hide, disable state refresh and create a plan artifact. 
It outputs a concise summary (the final plan line) and the path to the plan file artifact.

> **NOTE**: In addition to the keys provided as input, there are some generic regex patterns designed to redact values that are likely sensitive. The advantage of this approach is that it doesn’t require maintaining or explicitly listing every key, and additional patterns can be added in the future. However, it comes with a significant downside: it may produce false positives, i.e., it can redact values that shouldn’t be hidden.
> 
> The patterns added in this PR have been tested extensively and, so far, have not produced any false positives.
> 
> Here a small table with pros and cons
> 
> | Approach | Pros | Cons |
> |----------|------|------|
> | **Key-based (input list)** | Clear control, readable output, fewer false positives | Needs maintenance, risk of missing new keys |
> | **Generic regex** | No maintenance, covers unknown sensitive fields | Risk of false positives, output may become less useful |
> | **Hybrid (keys + targeted regex)** | Balanced: safe defaults + flexibility | Slightly more complex implementation |

Resolves: CES-1258 CES-1189


actions/filter-terraform-plan/action.yaml
Krusty93, line 1: **thought**: do we need a separate action to filter the output or we can put it next  to the plan gh action?

mamu0: It’s needed because we want to replace the values during the planning phase, not after it’s done, this way, anything sensitive gets removed in real time.
The part that cleans up the excess also makes more sense when combined with it, since it’s still part of the filtering process.
Besides, those extra elements don’t need to be planned anyway and removing them helps provide a cleaner plan.

actions/filter-terraform-plan/run_plan.sh
Krusty93, line 1: @copilot check for issues in this file. In particular, checks the logic matches what described in PR's description and the modified README file. Check also for security flaws, and eventually propose code improvements

gunzip:
suggestion: add tests for the bash script in order to facilitate eventual refactorings

actions/filter-terraform-plan/action.yaml
```diff
@@ -0,0 +1,53 @@
...
+    required: false
+    default: "."
+  no-refresh:
+    description: "If true, the terraform plan will be executed with the -refresh=false and -lock=false flags."
+    required: false
+    default: "false"
+  sensitive-keys:
+    description: "A comma-separated list of sensitive keys to filter from the plan output."
+    required: true
+  upload-artifact:
```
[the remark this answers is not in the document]
Krusty93: I meant the `upload-artifact` variable. I assume it will always be set to true?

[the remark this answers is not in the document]
lucacavallaro: If not used, this input should be removed ([YAGNI](http://en.wikipedia.org/wiki/You_aren%27t_gonna_need_it)). We can add it back in the future if actual use cases arise.

lucacavallaro (commented):
LGTM but before merging can you link a `infra_plan.yaml` run to make sure there are no regressons?

lucacavallaro (approved):

PR #869, opened 2025-09-11
Enable and configure pnpm when scaffolding a new monorepo

kin0992:
This pull request introduces support for `pnpm` as the package manager in the monorepo generator, along with automatic configuration of the `pnpm-plugin-pagopa` plugin. The changes ensure that new monorepos are set up with `pnpm` and the relevant plugin out of the box.

Closes CES-1268

### DIFF
--- a/packages/monorepo-generator/src/actions/pnpm.ts
+++ b/packages/monorepo-generator/src/actions/pnpm.ts
@@ -1,24 +1,26 @@
-import { execSync } from "node:child_process";
-import { ActionType } from "plop";
+import type { ActionType } from "plop";
 
-export const enablePnpm: ActionType = async (answers) => {
-  try {
-    execSync("corepack use pnpm", {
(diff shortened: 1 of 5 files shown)


packages/monorepo-generator/src/actions/pnpm.ts
```diff
@@ -0,0 +1,26 @@
...
+    $({
+      cwd: `${answers.repoSrc}/${answers.repoName}`,
+    })`corepack use pnpm@latest`,
+    () => new Error("Error enabling pnpm"),
+  ).match(
+    () => "pnpm enabled",
+    ({ message }) => message,
+  );
+
+export const addPagoPaPnpmPlugin: ActionType = async (answers) =>
```
kin0992, line 17: **thought**: even though we decided to not use the `catalog:dx` coming from the `pnpm-plugin-pagopa`, I think we could still use the plugin for other configurations, such as the other options the plugin is going to configure.

PR #943, opened 2025-09-29
Set standard tier to dx website

christian-calabrese:
The former "free" sku size limited the number of preview environments to 3. This will increase them to 10.

### DIFF
--- a/infra/resources/_modules/dx_website/main.tf
+++ b/infra/resources/_modules/dx_website/main.tf
@@ -9,6 +9,7 @@ resource "azurerm_static_web_app" "this" {
   resource_group_name = var.resource_group_name
   location            = "westeurope"
   sku_tier            = "Standard"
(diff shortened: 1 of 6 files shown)


mamu0 (approved):
LGTM

infra/resources/_modules/dx_website/main.tf
```diff
@@ -9,6 +9,7 @@ resource "azurerm_static_web_app" "this" {
   resource_group_name = var.resource_group_name
   location            = "westeurope"
   sku_tier            = "Standard"
```
gunzip, line 11: having both `sku_tier` and `sku_size` is confusing

christian-calabrese: I removed the sku_tier. Changing it, had no repercussions on the plan

PR #953, opened 2025-10-01
Replace `tier` variable with `use_case` for Function App

mamu0:

- default
- high_load

Add new variables `size` for SKU name override (check the `README.md` for allowed sizes).

Add documentation for module in `README.md`.

Resolves: CES-1327


rodoviti:
@mamu0 referring to commit [fix: removed unused size](https://github.com/pagopa/dx/pull/953/commits/2fe3923628c556b7063e5122601eb6ec0faa2b07), why did you remove support for _App Service Plan sku_ `P1mv3`?

mamu0:
> @mamu0 referring to commit [fix: removed unused size](https://github.com/pagopa/dx/pull/953/commits/2fe3923628c556b7063e5122601eb6ec0faa2b07), why did you remove support for _App Service Plan sku_ `P1mv3`?

Actually, that SKU wasn't supported by our module. It was added in this PR, but after an analysis, we determined that since we've never used it, it would be better not to integrate it right now. 
Therefore, for anyone who wants to use the module, the best course of action is to choose the SKU that most closely meets their needs from the whitelisted ones
cc @Krusty93

PR #958, opened 2025-10-02
Output subscription and tenant id in core

christian-calabrese:
These two new outputs reduce the DX cli need to generate additional data sources

Resolves CES-1331

### DIFF
--- a/.changeset/warm-peaches-play.md
+++ b/.changeset/warm-peaches-play.md
@@ -1,0 +1,6 @@
+---
+"azure_core_values_exporter": patch
+"azure_core_infra": patch
+---
+
+Output the subscription and tenant IDs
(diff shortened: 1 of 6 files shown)


.changeset/warm-peaches-play.md
kin0992, line 1: **nitpick**: It might be clearer to split the changelogs into two separate files. That way, the `azure_core_infra` changelog could also include a short usage guide.

For example:
```
# How to use these values
<description>

If you already use the `azure_core_values_exporter`:
1. Remove from your configuration:
   ```hcl
   data "azurerm_subscription" "current" {}
   ```
2. Replace:
```hcl
subscription_id = data.azurerm_subscription.current.id
```
with:
```hcl
subscription_id = module.<exporter>.subscription_id
```
---
Sorry for the formatting here, but I hope my nitpick is clear enough

christian-calabrese: Thank you!

PR #961, opened 2025-10-02
Replace `tier` variable with `use_case` for Container App

mamu0:

- default
- high_load

Add new variables `size` for CPU/Memory values override (check the `README.md` for allowed values).

Resolves: CES-1306

### DIFF
--- a/infra/modules/azure_container_app/tests/containerapp.tftest.hcl
+++ b/infra/modules/azure_container_app/tests/containerapp.tftest.hcl
@@ -97,7 +97,7 @@ run "container_app_is_correct_plan" {
   }
 
   assert {
-    condition     = azurerm_container_app.this.template[0].max_replicas == 2
+    condition     = azurerm_container_app.this.template[0].max_replicas == 8
     error_message = "The container app maximum replicas is not correct"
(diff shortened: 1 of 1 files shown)


.changeset/ninety-steaks-join.md
```diff
@@ -0,0 +1,51 @@
...
+| m         | default          | Ideal for `production` environments |
+| l         | _none_           | Does not exist anymore              |
+
+This change simplifies and clarifies the selection of Container App.
+
+To migrate to this new major version:
+
+1. Update the module version to `~> 4.0` in your Terraform configuration.
+2. Update your `module` configuration to use the new `use_case` variable instead of `tier`.
+3. Optionally, configure the new `size` variable to use the desired CPU/Memory configuration within the Container App.
```
kin0992, line 27: **question**: what if I do not configure the `size`? are there some default values used?

.changeset/ninety-steaks-join.md
kin0992, line 1: **note**: if you think this could be helpful, I would also mention the changes on the autoscaler; now the `replicas` are optional

PR #974, opened 2025-10-06
Update Container App module adding support to Functions

krusty93:
Close CES-1339

Depends on #961


infra/modules/azure_function_app/examples/container_app/main.tf
```diff
@@ -0,0 +1,83 @@
...
+  name = provider::dx::resource_name(merge(local.naming_config, {
+    domain        = "",
+    name          = "pep",
+    resource_type = "subnet"
+  }))
+  virtual_network_name = local.virtual_network.name
+  resource_group_name  = local.virtual_network.resource_group_name
+}
+
+data "azurerm_log_analytics_workspace" "common" {
```
mamu0, line 13: **suggestion**: What’s its purpose? It doesn’t seem to be called anywhere. if it’s not needed, we can remove it

infra/modules/azure_container_app/examples/function_app/main.tf
```diff
@@ -0,0 +1,104 @@
...
+
+      liveness_probe = {
+        path = "/"
+      }
+    },
+  ]
+
+  container_app_environment_id = module.container_app_environment.id
+
+  user_assigned_identity_id = module.container_app_environment.user_assigned_identity.id
```
gunzip, line 101: question: can we put this setting into the container app module itself? in order to avoid users specify this and just provide the cae

Krusty93: I'm afraid we can't unless we use a `data` in the module, but it will not be able to create the environment and the app at the same time as there's an unexisting dependency in the plan phase

gunzip:
Postponed.

PR #967, opened 2025-10-03
First working mcp server

christian-calabrese:
This MCP server allows GitHub Copilot (or any other GenAI agent) to query the DX documentation website contents

PR #977, opened 2025-10-07
Mcp server infrastructure

christian-calabrese:
This PR implements the infrastructure needed to expose the PoC of the [DX MCP Server](https://github.com/pagopa/dx/pull/967)
The terraform changes have already been applied to verify they're working correctly. The MCP server is already reachable here: https://api.dev.dx.pagopa.it/mcp


gunzip (commented):
suggestion: add comments that clarify the purpose of resources

lucacavallaro:
@christian-calabrese if you want, you can merge these changes into the PR with the mcp server code

christian-calabrese:
Some notes:
- Please note that the bedrock knowledge base and the S3 Vector bucket are not created via terraform. This is because the latter is not yet in GA, hence no terraform resources exist yet.
- The Terraform Validation job fails due to an unknown trivy finding. This shouldn't cause any issue after merging, can be seen as a bug.
- The terraform plan on prod fails due to some missing permissions that are being fixed on a different stream (https://github.com/pagopa/eng-azure-governance/pull/199). This PR does not actually edit any file there, but the modules used in dev.

PR #1591, opened 2026-04-09
Fix documentation link for canary deployment in rollout script

gunzip:
Update the link in the rollout script to point to the correct section of the workflow documentation for canary deployments.

Close CES-1811


kin0992 (approved):
LGTM.

It's up to you to decide whether to add the changelog file or not since the updated action is a workspace.

PR #1617, opened 2026-04-10
Refactor Container App Environment module

Krusty93:
Refactor the module structure for improved organization, enhance test coverage with updated test cases, and provide a clearer example for usage. 

Resolves CES-1920

Depends on #1650


infra/modules/azure_container_app_environment/container_app_environment.tf
```diff
@@ -28,6 +31,8 @@ resource "azurerm_container_app_environment" "this" {
 }
 
 resource "azurerm_management_lock" "cae_lock" {
+  count = var.environment.env_short == "d" ? 0 : 1
```
gunzip: question: shouldn't this reflect the `use_case` instead ?

infra/modules/azure_container_app_environment/container_app_environment.tf
```diff
@@ -1,19 +1,22 @@
...
-  infrastructure_subnet_id       = var.subnet_id == null ? azurerm_subnet.this[0].id : var.subnet_id
-  internal_load_balancer_enabled = !var.public_network_access_enabled
+  identity {
+    type = "SystemAssigned"
+  }
+
+  infrastructure_subnet_id       = azurerm_subnet.this.id
+  internal_load_balancer_enabled = !var.networking.public_network_access_enabled
+  logs_destination               = "azure-monitor"
+  zone_redundancy_enabled        = var.environment.env_short != "d" ? true : false
```
gunzip: question: shouldn't this reflect the use_case instead ?

krusty93: implemented

infra/modules/azure_container_app_environment/container_app_environment.tf
krusty93: implemented

infra/modules/azure_container_app_environment/README.md
```diff
@@ -4,19 +4,47 @@
 
 This Terraform module deploys an Azure Container App Environment along with necessary networking components.
 
+## Diagram
+
+The following diagram illustrates the architecture and relationships between the main components of this module:
+
+![diagram](diagram.svg)
```
mamu0:
```suggestion
![diagram](https://raw.githubusercontent.com/pagopa/dx/refs/heads/main/infra/modules/azure_container_app_environment/diagram.svg)
```

[question](https://pullpo.io/cc?l=question): Did you use the new SKILL? It's strange that it didn't generate the [link to the raw file on GitHub](https://github.com/pagopa/dx/blob/main/plugins/terraform/skills/generate-terraform-module-diagram/SKILL.md#building-the-raw-github-url)

krusty93: yes I did! we could do another test with the next module

.nx/version-plans/version-plan-1775806969479.md
```diff
@@ -0,0 +1,101 @@
...
+| ------------------------ | --------------------------------------------------------------------------------------------- |
+| `user_assigned_identity` | **Removed.** The module now uses a system-assigned identity.                                  |
+| `principal_id`           | **New.** Principal ID of the system-assigned managed identity.                                |
+| `subnet`                 | **New.** Exposes `id`, `name`, and `resource_group_name` of the automatically created subnet. |
+
+### Migration Guide (v1 → v2)
+
+#### 1. Remove `subnet_id` and `subnet_cidr`
+
+The module now manages its own subnet. Remove both variables and ensure the target virtual network has enough address space for a `/23` block.
```
christian-calabrese: [question](https://pullpo.io/cc?l=question) 
Changing the infrastructure_subnet_id causes recreation of the container app environment. I know it might sound bad, but what do you think about leaving the subnet_id and subnet_cidr variable to avoid this?

krusty93: Can the subnet be moved within the state?
Consider also we only have 3 occurrences, of which:
- 1 not used
- 1 not in production yet
- 1 in dev

christian-calabrese: I don't think you can move it cause the provider function would generate a different cidr cause the existing one is occupied by the existing subnet

I see that it is used by io-services-cms and io-messages, but if they are not in production, that's ok!

krusty93: @mamu0 is it possible to add the import capability to the cidr generator resource?

mamu0: This needs to be verified.
I'm not exactly sure how an **import** would work, because the resource doesn't actually create anything in **Azure** that can be imported via an ID.
It simply returns a CIDR value to be passed as an input to the subnet

krusty93: Being a resource, does it store something in the terraform state, right? maybe it only needs to import a value there, like:
`terraform import <id> "10.0.0.0/24"`

mamu0: As I mentioned, this needs to be verified. An _import_ would have to be implemented, but the resource doesn't have a corresponding entity on **Azure**.
What I mean is that there’s no 'real-world' resource to import; the resource simply generates a CIDR string to be passed to the subnet resource.

Furthermore, what would be the benefit of importing this? It essentially calculates the best available CIDR based on existing ones, if it has already been used to create a subnet, what’s the point of importing it back?

krusty93: Because of the concerns highlighted by @christian-calabrese , see above.

It does not create any resource on the CSP, but it does on the state file. In case of migration it is relevant.

I've asked Copilot to implement the `import` method and it did this: https://github.com/pagopa/dx/pull/1650

krusty93: it worked!

```terraform
➜  terraform plan -target="azurerm_subnet.this"
Terraform will perform the following actions:

  # azurerm_subnet.this will be updated in-place
  ~ resource "azurerm_subnet" "this" {
      ~ address_prefixes                              = [
          - "10.51.9.0/29",
        ] -> (known after apply)
        id                                            = "/subscriptions/35e6e3b2-4388-470e-a1b9-ad3bc34326d1/resourceGroups/dx-d-itn-network-rg-01/providers/Microsoft.Network/virtualNetworks/dx-d-itn-common-vnet-01/subnets/test-subnet"
        name                                          = "test-subnet"
        # (8 unchanged attributes hidden)
    }

  # dx_available_subnet_cidr.test_subnet will be created
  + resource "dx_available_subnet_cidr" "test_subnet" {
      + cidr_block         = (known after apply)
      + id                 = (known after apply)
      + prefix_length      = 29
      + virtual_network_id = "/subscriptions/35e6e3b2-4388-470e-a1b9-ad3bc34326d1/resourceGroups/dx-d-itn-network-rg-01/providers/Microsoft.Network/virtualNetworks/dx-d-itn-common-vnet-01"
    }

Plan: 1 to add, 1 to change, 0 to destroy.


➜  terraform import dx_available_subnet_cidr.test_subnet "/subscriptions/35e6e3b2-4388-470e-a1b9-ad3bc34326d1/resourceGroups/dx-d-itn-network-rg-01/providers/Microsoft.Network/virtualNetworks/dx-d-itn-common-vnet-01/subnets/test-subnet"
dx_available_subnet_cidr.test_subnet: Import prepared!
  Prepared dx_available_subnet_cidr for import

Import successful!

➜  terraform plan -target="azurerm_subnet.this"
No changes. Your infrastructure matches the configuration.

➜  terraform state show dx_available_subnet_cidr.test_subnet
# dx_available_subnet_cidr.test_subnet:
resource "dx_available_subnet_cidr" "test_subnet" {
    cidr_block         = "10.51.9.0/29"
    id                 = "_subscriptions_35e6e3b2-4388-470e-a1b9-ad3bc34326d1_resourceGroups_dx-d-itn-network-rg-01_providers_Microsoft.Network_virtualNetworks_dx-d-itn-common-vnet-01_29_10.51.9.0_29"
    prefix_length      = 29
    virtual_network_id = "/subscriptions/35e6e3b2-4388-470e-a1b9-ad3bc34326d1/resourceGroups/dx-d-itn-network-rg-01/providers/Microsoft.Network/virtualNetworks/dx-d-itn-common-vnet-01"

➜  terraform state show azurerm_subnet.this                             
# azurerm_subnet.this:
resource "azurerm_subnet" "this" {
    address_prefixes                              = [
        "10.51.9.0/29",
    ]
    default_outbound_access_enabled               = true
    id                                            = "/subscriptions/35e6e3b2-4388-470e-a1b9-ad3bc34326d1/resourceGroups/dx-d-itn-network-rg-01/providers/Microsoft.Network/virtualNetworks/dx-d-itn-common-vnet-01/subnets/test-subnet"
    name                                          = "test-subnet"
    private_endpoint_network_policies             = "Disabled"
    private_link_service_network_policies_enabled = true
    resource_group_name                           = "dx-d-itn-network-rg-01"
    sharing_scope                                 = null
    virtual_network_name                          = "dx-d-itn-common-vnet-01"
}
```

@christian-calabrese we can then remove and import again both the subnet and the generator to the terraform state file

christian-calabrese: Amazing @Krusty93 !

infra/modules/azure_container_app_environment/locals.tf
```diff
@@ -1,13 +1,27 @@
...
-  has_existing_subnet = var.subnet_id != null
+  use_case_features = local.use_cases[var.use_case]
+
+  vnet_id       = provider::azurerm::normalise_resource_id("${data.azurerm_subscription.current.id}/resourceGroups/${var.networking.virtual_network.resource_group_name}/providers/Microsoft.Network/virtualNetworks/${var.networking.virtual_network.name}")
+  subnet_pep_id = provider::azurerm::normalise_resource_id("${local.vnet_id}/subnets/${local.pep_subnet_name}")
+  pep_subnet_name = provider::dx::resource_name(merge(var.environment, {
+    domain          = "",
+    app_name        = "pep",
+    resource_type   = "subnet"
+    instance_number = "01"
```
krusty93: issue: make it dynamic

infra/modules/azure_container_app_environment/tests/setup/main.tf
```diff
@@ -1,100 +1,53 @@
...
-      version = ">= 0.0.6, < 1.0.0"
-    }
-  }
-}
-
 locals {
-  naming_config = {
-    prefix          = var.environment.prefix,
-    environment     = var.environment.env_short,
+  existing_resources = {
```
gunzip, line 2: question: other modues uses `naming_config` which becames a sort of de facto standard. aren’t we at risk of worsening the dx by changing its name?

infra/modules/azure_container_app_environment/locals.tf
gunzip: may you link the related jira task?

infra/modules/azure_container_app_environment/README.md
```diff
@@ -4,19 +4,47 @@
...
+
+The following diagram illustrates the architecture and relationships between the main components of this module:
+
+![diagram](https://raw.githubusercontent.com/pagopa/dx/refs/heads/main/infra/modules/azure_container_app_environment/diagram.svg)
+
 ## Features
 
 - **Azure Container App Environment**: Deploys an Azure Container App Environment for hosting containerized applications.
-- **Flexible Ingress**: Choose between private (internal VNet only) or public ingress via the `public_network_access_enabled` variable (default: `false` — private mode).
+- **Flexible Ingress**: Choose between private (internal VNet only) or public ingress via `networking.public_network_access_enabled` (default: `false` — private mode).
```
gunzip: thoughts: if I read this description and I don't know anything about how it works, I find myself wondering: does "internal vnet" still mean public access via app gateway in this case?

infra/modules/azure_container_app_environment/tests/setup/main.tf
krusty93: `naming_config` is just an internal wrapper of the `environment`, deriving from the incompatible contract between the dx provider and dx modules. It can be removed from the entire dx codebase, to me

infra/modules/azure_container_app_environment/locals.tf
krusty93: what task? CES-1920?

infra/modules/azure_container_app_environment/tests/setup/main.tf
gunzip: do module users use it? if it's only used internally, it's fine to rename it

infra/modules/azure_container_app_environment/locals.tf
gunzip: nevermind my fault

infra/modules/azure_container_app_environment/tests/setup/main.tf
krusty93: Nope (and this occurrence is in tests files)

infra/modules/azure_container_app_environment/README.md
krusty93: Removed the details, it now looks easier to understand to me

gunzip (commented):
lgtm but it looks like tests dont pass

Krusty93:
> lgtm but it looks like tests dont pass

I'm waiting for the v0.10 of the provider to become available...

edit: it is

.nx/version-plans/version-plan-1775806969479.md
```diff
@@ -0,0 +1,101 @@
...
+
+# After
+module "cae" {
+  source = "..."
+
+  networking = {
+    virtual_network = {
+      name                = "my-vnet"
+      resource_group_name = "rg-network"
+    }
```
kin0992: **question(non-blocking)**: why not provide a different DX, where we can ask the resource ID and then, internally, we parse it to extract any information we need?

something like this (pseudocode):
```hcl
type = object({
    vnet_id                       = string
    private_dns_zone_id           = optional(string, null)
    public_network_access_enabled = optional(bool, false)
})
```
and then parse the IDs accorgingly
```hcl
  vnet_id                  = provider::azurerm::normalise_resource_id(var.networking.vnet_id)
  parsed_vnet              = provider::azurerm::parse_resource_id(local.vnet_id)
  vnet_name                = local.parsed_vnet.resource_name
  vnet_resource_group_name = local.parsed_vnet.resource_group_name
```

What do you think?

kin0992: sorry for my late reply, I just discovered that during the #1661 😄

krusty93: It might work to me. Do you think it'd be easier from ICs pov?

kin0992: Yes, I mean, it reduces the number of inputs and maybe the cognitive load too (they just need to point an ID of a resource, that's it).

Anyway, think of this as a proposal, since we are already going to bump a major

krusty93: Sorry, I read just now your comment about this topic in #1661.
We can try this pattern out, I'm going to update this PR

krusty93: Pushed! Although I kept `private_dns_zone_resource_group_name` as the zone id is not usually at hand (and known neither). But switching to vnet id involved some simplification

kin0992: LGTM

.nx/version-plans/version-plan-1775806969479.md
```diff
@@ -0,0 +1,101 @@
...
+}
+
+# After
+module "cae" {
+  source = "..."
+
+  networking = {
+    virtual_network = {
+      name                = "my-vnet"
+      resource_group_name = "rg-network"
```
kin0992: **issue**: you should update this as well

kin0992 (changes requested):

.nx/version-plans/version-plan-1775806969479.md
krusty93: done

kin0992 (approved):
👍

PR #1620, opened 2026-04-13
Give required permissions to GItHub workflows to access Common resource group

Krusty93:
Enhance the `addModule` function to accept additional data and implement role assignments for identities in the Core's common resource group when a new environment is initialized. This change improves role management and access control in the infrastructure setup.

Resolves CES-1862

### DIFF
--- a/apps/cli/src/adapters/plop/generators/environment/actions.ts
+++ b/apps/cli/src/adapters/plop/generators/environment/actions.ts
@@ -7,11 +7,7 @@ import { Environment } from "../../../../domain/environment.js";
 import { formatTerraformCode } from "../../../terraform/fmt.js";
 import { payloadSchema } from "./prompts.js";
 
-const addModule = (
-  env: Environment,
-  templatesPath: string,
-  extraData: Record<string, unknown> = {},
-) => {
+const addModule = (env: Environment, templatesPath: string, init = false) => {
   const cloudAccountsByCsp = Object.groupBy(
     env.cloudAccounts,
     (account) => account.csp,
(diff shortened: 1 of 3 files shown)


apps/cli/src/adapters/plop/generators/environment/actions.ts
```diff
@@ -50,7 +54,9 @@ export default function getActions(
 
     const { env, github, init } = payloadSchema.parse(payload);
 
-    const addEnvironmentModule = addModule(env, templatesPath);
+    const addEnvironmentModule = addModule(env, templatesPath, {
+      init: !!init,
```
kin0992: **question**: why `!!`?

apps/cli/src/adapters/plop/generators/environment/actions.ts
```diff
@@ -7,7 +7,11 @@ import { Environment } from "../../../../domain/environment.js";
 import { formatTerraformCode } from "../../../terraform/fmt.js";
 import { payloadSchema } from "./prompts.js";
 
-const addModule = (env: Environment, templatesPath: string) => {
+const addModule = (
+  env: Environment,
+  templatesPath: string,
+  extraData: Record<string, unknown> = {},
```
kin0992: **question**: wouldn't it be possible to improve this type?

apps/cli/src/adapters/plop/generators/environment/actions.ts
krusty93: Copilot added it, I thought it is correct to enforce boolean type as it's set as optional in zod schema

apps/cli/src/adapters/plop/generators/environment/actions.ts
krusty93: What's the best way to use that record as a bag of mixed properties? Or do you prefer to explicitly set `init: boolean`?

kin0992: If possible, I would manage a typed object, like the `cloudAccountsByCsp`.

So, my point here is: do you think it is possible to change the `addModule` function signature to create the record inside the function and have that record properly typed?

krusty93: Is it better now?

kin0992 (approved):

PR #1627, opened 2026-04-14
Improve terraform projects discovery

lucacavallaro:
- Skip creating Nx projects for Terraform files located under `tests`/`_tests` and `examples`/`example` directories.
- Add `projectType` to generated project configurations, classifying paths under `modules`/`_modules` as `library` and others as `application`.

### How to test

```bash
# list all terraform apps (modules that can be applied)
nx show projects --type app --projects tag:terraform 
```

```bash
# list all libraries
nx show projects --type lib --projects tag:terraform
```

### DIFF
--- a/.nx/version-plans/version-plan-1776329201422.md
+++ b/.nx/version-plans/version-plan-1776329201422.md
@@ -1,0 +1,5 @@
+---
+"@pagopa/nx-terraform-plugin": preminor
+---
+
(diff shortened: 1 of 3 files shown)


gunzip (commented):
question: shouldn't we document this behavior somewhere? is the plugin versioned? in case we should add a version plan.

mamu0 (approved):

PR #1629, opened 2026-04-15
Document Azure token validation

kin0992:
This pull request documents why the DX CLI still runs `az group list` during the Azure precondition check.

The extra command is intentional: `az account show` reads the cached CLI context, while `az group list` fails when the current Azure session token has expired. Keeping that behavior documented makes the precondition check easier to understand and avoids future regressions during cleanup work.

### DIFF
--- a/.nx/version-plans/version-plan-1776247234159.md
+++ b/.nx/version-plans/version-plan-1776247234159.md
@@ -1,7 +1,0 @@
----
-'@pagopa/dx-cli': patch
----
-
-Remove `az group list` execution since the output is not used.
-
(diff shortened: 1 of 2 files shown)


Krusty93:
**issue**: that command was introduced by me and @lucacavallaro to check if the current session token is still valid as it expires after 12 hours.

gunzip:
maybe we should add a comment in the code?

kin0992:
I'm going to change this PR by adding a comment, since it is not clear, right now, the reason why that command is there

christian-calabrese (approved):

Krusty93 (approved):

PR #1630, opened 2026-04-15
Update the tracing package to be compatible with Managed Identity

kin0992:
This pull request introduces support for Microsoft Entra ID (Managed Identity) authentication in the `@pagopa/azure-tracing` package, making it possible to securely connect to Application Insights using Azure credentials instead of only a connection string. The connection-string-only authentication remains the default for backward compatibility but is now deprecated and will be removed in the next major release. The documentation and environment handling have been updated accordingly, and comprehensive tests have been added to ensure correct behavior in both authentication modes.

### DIFF
--- a/packages/azure-tracing/README.md
+++ b/packages/azure-tracing/README.md
@@ -43,10 +43,16 @@ For more background on this workaround, see:
 
 In order to enable tracing, you also need to set the following environment variables:
 
-| **Name**                                  | **Required** | **Default** |
-| ----------------------------------------- | ------------ | ----------- |
-| **APPINSIGHTS_SAMPLING_PERCENTAGE**       | false        | 5           |
-| **APPLICATIONINSIGHTS_CONNECTION_STRING** | true         | -           |
+| **Name**                                      | **Required** | **Default** |
+| --------------------------------------------- | ------------ | ----------- |
+| **APPINSIGHTS_SAMPLING_PERCENTAGE**           | false        | 5           |
+| **APPLICATIONINSIGHTS_CONNECTION_STRING**     | true         | -           |
+| **APPLICATIONINSIGHTS_ENTRA_ID_AUTH_ENABLED** | false        | false       |
+
+> [!WARNING]
+> Connection-string-only authentication (the current default) is **deprecated** and will be removed in the next major version.
+> Set `APPLICATIONINSIGHTS_ENTRA_ID_AUTH_ENABLED=true` to adopt Microsoft Entra ID authentication now.
+> See [Microsoft Entra ID Authentication](#microsoft-entra-id-authentication) for details.
 
 #### Step 2: Register Azure Function Lifecycle Hooks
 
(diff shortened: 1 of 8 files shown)


packages/azure-tracing/src/azure/monitor/start-from-env.ts
```diff
@@ -1,10 +1,42 @@
...
+ * Deprecation warning printed when connection-string-only auth is used.
+ * Entra ID auth (DefaultAzureCredential) will become the default in the next major release.
+ */
+const DEPRECATION_WARNING =
+  "[@pagopa/azure-tracing] Connection-string-only authentication is deprecated " +
+  "and will be removed in the next major version. " +
+  "Set APPLICATIONINSIGHTS_ENTRA_ID_AUTH_ENABLED=true to use Microsoft Entra ID " +
+  "authentication (recommended). See README for details.";
 
 export const initFromEnv = () => {
```
kin0992, line 9: the connection string is used to route the metrics to the proper application insights instace, while the credential are the authentication layer.

when we will migrate to support the managed identity as the only authentication method, we will still need the connection string.

the main difference here is:
- if the application insights instance has the `local_authentication_disabled` set to `true`, then having just the connection string would not be enough to send the metrics (you will get a 401); the credentials are required
- if the application insights instance has the `local_authentication_disabled` set to `false`, then the connection string is enough

I hope this clarifies your thoughts.

gunzip:
> the main difference here is:
> 
> if the application insights instance has the local_authentication_disabled set to true, then having just the connection string would not be enough to send the metrics (you will get a 401); the credentials are required
> if the application insights instance has the local_authentication_disabled set to false, then the connection string is enough
> 

should we document this or adapt terraform modules?

packages/azure-tracing/README.md
```diff
@@ -43,10 +43,16 @@ For more background on this workaround, see:
...
 
-| **Name**                                  | **Required** | **Default** |
-| ----------------------------------------- | ------------ | ----------- |
-| **APPINSIGHTS_SAMPLING_PERCENTAGE**       | false        | 5           |
-| **APPLICATIONINSIGHTS_CONNECTION_STRING** | true         | -           |
+| **Name**                                      | **Required** | **Default** |
+| --------------------------------------------- | ------------ | ----------- |
+| **APPINSIGHTS_SAMPLING_PERCENTAGE**           | false        | 5           |
+| **APPLICATIONINSIGHTS_CONNECTION_STRING**     | true         | -           |
+| **APPLICATIONINSIGHTS_ENTRA_ID_AUTH_ENABLED** | false        | false       |
```
krusty93, line 50: question: is this variable name mandatory in some way? Because they usually use something like `MANAGED_IDENTITY` or `SYSTEM_ASSIGNED`

Krusty93:
question: What are the required roles for the managed identity?

kin0992:
> > the main difference here is:
> > if the application insights instance has the local_authentication_disabled set to true, then having just the connection string would not be enough to send the metrics (you will get a 401); the credentials are required
> > if the application insights instance has the local_authentication_disabled set to false, then the connection string is enough
> 
> should we document this or adapt terraform modules?

@gunzip well, there is no Terraform module that is going to create an application insights instance, but we can think about it.

The scope of this PR is just to prepare the `@pagopa/azure-tracing` package to use the Managed Identity.
If we want to provide the best DX (and IMHO we should), then we should add the activities to the backlog (there are many and they involve the function-app/app-service modules, APIM as well, and roles assignment too).

kin0992:
> question: What are the required roles for the managed identity?

@Krusty93 the required role is "Monitoring Metrics Publisher". I documented it in the README

Krusty93:
> > question: What are the required roles for the managed identity?
> 
> @Krusty93 the required role is "Monitoring Metrics Publisher". I documented it in the README

Sorry I missed it. Nice!

packages/azure-tracing/README.md
kin0992: no, the name has no requirements in some way. I defined it that way to make it clear to the developers just by reading the name, but we could find a convention.

kin0992: note that, IMHO, the `APPLICATIONINSIGHTS_ENTRA_ID_AUTH_ENABLED` env variable is just temporary.
Once we decide to support just the Managed Identity as authentication method, we are going to create a new major and we will remove the `APPLICATIONINSIGHTS_ENTRA_ID_AUTH_ENABLED` environment variable.

For now, that variable is used for backward compatibility.

Krusty93 (approved):

PR #1638, opened 2026-04-16
Add targets to terraform projects

lucacavallaro:
Closes CES-1454

### DIFF
--- a/.nx/version-plans/version-plan-1776356493909.md
+++ b/.nx/version-plans/version-plan-1776356493909.md
@@ -1,5 +1,9 @@
 ---
-'@pagopa/nx-terraform-plugin': preminor
+"@pagopa/nx-terraform-plugin": preminor
 ---
 
-Add targets
+Terraform projects now expose a default set of targets for common Terraform CLI commands.
+
+Every Terraform project will include `init`, `format`, `validate`, `test`, `output`, and `console`. Terraform application projects will also include `plan` and `apply`.
+
+All target names are configurable, via a dedicated `<targetName>TargetName` property on plugin configuration.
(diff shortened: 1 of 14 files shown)


.nx/version-plans/version-plan-1776356493909.md
```diff
@@ -0,0 +1,5 @@
+---
+'@pagopa/nx-terraform-plugin': preminor
+---
+
+Add targets
```
kin0992: **question**: once the release is ready (so we are bumping the package to a `minor` instead of `preminor` how would it be the changelog?

If the changelog is going to be generated starting from these `pre*` version plans, then I'd improve the changelog lines to add more details.
If we need to create a brand new file, then we can proceed with this minimal changelog

packages/nx-terraform-plugin/src/options.ts
```diff
@@ -0,0 +1,62 @@
...
+    .partial()
+    .safeParse(options ?? {});
+  if (!parseResult.success) {
+    const validationErrors = parseResult.error.issues
+      .map((issue) => {
+        const path = issue.path.length > 0 ? issue.path.join(".") : "options";
+        return `${path}: ${issue.message}`;
+      })
+      .join("; ");
+    throw new Error(`Invalid Terraform plugin options: ${validationErrors}`);
```
kin0992, line 47: **thought**: maybe `z.prettifyError(result.error)` could be useful?

lucacavallaro: tried it, but I don't like the formatting (it is not meant to be wrapped by another error message)

.nx/version-plans/version-plan-1776356493909.md
lucacavallaro: Done!

kin0992 (approved):

PR #1647, opened 2026-04-20
Add managed Redis resource naming

kin0992:
## Summary
- add `managed_redis` with the `amr` abbreviation in the Azure DX provider naming map
- add regression coverage to keep `managed_redis` distinct from `redis_cache`

## Why
Azure Managed Redis needs a dedicated resource type in the provider so downstream modules can generate the expected naming without relying on the legacy `redis_cache` type.

Resolves CES-1958

### DIFF
--- a/.nx/version-plans/version-plan-1776693686799.md
+++ b/.nx/version-plans/version-plan-1776693686799.md
@@ -2,4 +2,8 @@
 provider-azure: minor
 ---
 
-Add managed_redis resource naming support
+Breaking change: remove the legacy `redis_cache` resource type from the Azure `resource_name` function.
+
+Consumers must migrate any `resource_type = "redis_cache"` usage to the supported `managed_redis` value.
+
+This remains a minor bump because the provider is still in the `0.x` series, where minor releases can carry breaking changes.
--- a/.nx/version-plans/version-plan-1776693686799.md
+++ b/.nx/version-plans/version-plan-1776693686799.md
@@ -2,4 +2,4 @@
 provider-azure: minor
 ---
 
-Add managed_redis resource naming support
+Add `managed_redis` resource naming support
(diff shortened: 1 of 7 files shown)


.nx/version-plans/version-plan-1776693686799.md
```diff
@@ -0,0 +1,5 @@
+---
+provider-azure: minor
+---
+
+Add managed_redis resource naming support
```
kin0992: **thought**: maybe we could drop support for `redis_cache` and bump a `major` instead

gunzip: deal

krusty93: we're still on version `0`, we can break things :)

gunzip: question: avoiding a major update will break `io-fims` and `io-messages` IaC. is this acceptable for us? we can ask them to pin the previous release of the provider anyway.

kin0992: in this case, IMHO I'd bump a major

krusty93: @gunzip as the provider is still at v0 and they're using resource primitives, we may ask them to replace the name assignment with an hardcoded string. What I would like is to remove the need to specify suffixes for each resource type (e.g. `apim-snet`, `app-snet`, `cosno-pep`, `blob-pep`). I'd like to have a more dynamic solution, e.g. `resource_name(var.environment, mandatory-resource-type: "apim", optional-subresource: "subnet")`.

At the same time, we're still at v0, but the provider is so widespread within the codebases that makes difficult to do any change. However, we can think of a migration path while keeping backward compatibility for a while.

For all of these reasons, I suggest to avoid any break in this PR and cracking on with the work to do, and thinking about this issue elsewhere

gunzip: ook!

Krusty93 (approved):

PR #1648, opened 2026-04-20
Implement terraform module that merges multiple RBAC roles into a custom one

christian-calabrese:
This PR adds the new azure_merge_roles Terraform module to create a single Azure custom role from multiple source roles at subscription or management group scope.

It also adds validation, documentation, examples, Azure E2E coverage, and the workflow used to publish the probe image.

Resolves CES-1884


infra/modules/azure_merge_roles/README.md
```diff
@@ -0,0 +1,134 @@
+# DX - Azure Merge Roles
+
+![Terraform Module Downloads](https://img.shields.io/terraform/module/dm/pagopa-dx/azure-merge-roles/azurerm?label=downloads&cacheSeconds=5000&link=https%3A%2F%2Fregistry.terraform.io%2Fmodules%2Fpagopa-dx%2Fazure-merge-roles%2Fazurerm%2Flatest&logo=terraform)
+
+This module creates an Azure custom role definition by merging two or more compatible Azure roles into a single assignable role.
+
+It is designed for the RBAC reduction strategy: define a smaller set of reusable custom roles, then assign those roles to Entra ID groups instead of assigning many built-in roles directly to users, managed identities, or service principals.
+
+## What This Module Does
```
krusty93, line 9: [question](https://pullpo.io/cc?l=question) 
Does it worth to add a diagram here?

infra/modules/azure_merge_roles/README.md
```diff
@@ -0,0 +1,134 @@
...
+- Requires an explicit `scope` and defaults `assignable_scopes` to `[scope]`.
+- Supports role definitions created at management group scope as well as subscription scope.
+
+## Merge Strategy
+
+Azure RBAC computes effective permissions per role assignment as:
+
+`Actions - NotActions`
+
+and `NotActions` is not a deny rule across separate role assignments.
```
krusty93:
```suggestion
`Actions - NotActions` and `NotActions` is not a deny rule across separate role assignments.
```

infra/modules/azure_merge_roles/README.md
```diff
@@ -0,0 +1,134 @@
...
+    "Reader",
+    "Monitoring Reader",
+  ]
+
+  description = "Custom role for observability read access"
+}
+
+resource "azurerm_role_assignment" "observability_reader" {
+  scope                = data.azurerm_subscription.current.id
+  role_definition_name = module.observability_reader.custom_role_name
```
krusty93: [nitpick](https://pullpo.io/cc?l=nitpick) 
`role_definition_id` is preferable rather than `role_definition_name` in order to do not break things when names are changed

infra/modules/azure_merge_roles/variables.tf
```diff
@@ -0,0 +1,82 @@
...
+    error_message = "source_roles must contain at least two Azure role names. If you only need one role, assign it directly instead of creating an equivalent custom role."
+  }
+
+  validation {
+    condition     = length(var.source_roles) == length(distinct(var.source_roles))
+    error_message = "source_roles must not contain duplicates."
+  }
+}
+
+variable "assignable_scopes" {
```
krusty93: thought: From an UX perspective, does it worth to have this variable or we can assume that the `assignable_scopes` is always equals to `scope`?

infra/modules/azure_merge_roles/package.json
```diff
@@ -0,0 +1,13 @@
...
+  "name": "azure_merge_roles",
+  "version": "0.0.0",
+  "private": true,
+  "provider": "azurerm",
+  "description": "Easily merge multiple RBAC built-in roles into a single custom role to save on the number of the required role assignments",
+  "scripts": {
+    "tf-init": "terraform init",
+    "test:unit": "terraform test -filter='tests/unit.tftest.hcl'",
+    "test:contract": "terraform test -filter='tests/contract.tftest.hcl'",
+    "test:e2e": "cd tests && GOWORK=off go test -v -timeout 1h ./..."
```
krusty93, line 11: question: What is `GOWORK`?

infra/modules/azure_merge_roles/variables.tf
```diff
@@ -0,0 +1,82 @@
...
+  validation {
+    condition     = var.assignable_scopes == null ? true : length(var.assignable_scopes) == length(distinct(var.assignable_scopes))
+    error_message = "assignable_scopes must not contain duplicates."
+  }
+}
+
+variable "description" {
+  description = "Optional custom description for the merged role definition. Defaults to a generated description based on source_roles."
+  type        = string
+  default     = null
```
krusty93: thought: The risk here is to have custom roles without any explanation about usage nor reasons. I wonder if instead of `description` (as it can be autogenerated), we can ask for a `reason`. Then, merge the `reason` with the `description` for the `azurerm_role_definition`'s `description` field.
It's just a proposal, maybe you can think about a better approach to prevent the issue

infra/modules/azure_merge_roles/examples/blob_rbac_validation/locals.tf
```diff
@@ -0,0 +1,45 @@
...
+  environment = {
+    prefix          = "dx"
+    env_short       = "d"
+    location        = "italynorth"
+    domain          = "e2e"
+    app_name        = "mrg"
+    instance_number = format("%02d", random_integer.instance_number.result)
+  }
+
+  naming_config = {
```
krusty93: suggestion: I suggest to remove `naming_config` everywhere is possible as it's not needed anymore

infra/modules/azure_merge_roles/examples/blob_rbac_validation/fixtures.tf
```diff
@@ -0,0 +1,213 @@
...
+
+resource "azurerm_user_assigned_identity" "limited_probe" {
+  name                = "dx-e2e-blob-limited-mi-${local.environment.instance_number}"
+  location            = local.environment.location
+  resource_group_name = azurerm_resource_group.e2e_blob_rbac.name
+
+  tags = local.tags
+}
+
+resource "azurerm_user_assigned_identity" "full_probe" {
```
krusty93, line 48: question: Any reason to use the user-assigned rather than the system-assigned?

christian-calabrese: No, there's no difference in practice 👍

infra/modules/azure_merge_roles/README.md
christian-calabrese: Done!

infra/modules/azure_merge_roles/variables.tf
christian-calabrese: Done, removed the assignable_scopes that we don't usually use

infra/modules/azure_merge_roles/variables.tf
christian-calabrese: Done, good idea!

infra/modules/azure_merge_roles/package.json
christian-calabrese: GOWORK controls Go workspace mode. Here GOWORK=off is needed to force the E2E tests to use their local go.mod instead of inheriting the repository root go.work, which does not include this test module.

infra/modules/azure_merge_roles/README.md
christian-calabrese: I do not think a full infrastructure diagram adds much value here, because the module creates a single custom role and the tricky part is the merge semantics rather than the resource topology

Krusty93 (approved):
LGTM

PR #1660, opened 2026-04-22
Fix Container App rollout revisions

gunzip:
## Summary
- track the revision that is currently serving traffic before copying and promoting a new Azure Container App revision
- pass the current and target revision names to the incremental rollout action for container apps
- update rollout traffic using explicit revision weights instead of relying on `latest`

## Why
In multiple revision mode, Azure Container Apps can keep older revisions active even when they are not receiving traffic. When that happens, the reusable workflow can deploy from one active revision while ingress traffic still points to another, and the rollout action can leave production traffic on the wrong revision.

### DIFF
--- a/.github/workflows/release-azure-containerapp-v1.yaml
+++ b/.github/workflows/release-azure-containerapp-v1.yaml
@@ -138,16 +138,46 @@ jobs:
 
           echo "revision_mode=$revision_mode" >> $GITHUB_OUTPUT
 
-      - name: Get Container App Current Revision
+      - name: Get Container App Production Revision
         id: get_current_revision
         run: |
           set -euo
 
-          current_revision=$(az containerapp revision list \
+          container_app=$(az containerapp show \
             --name "$CONTAINER_APP_NAME" \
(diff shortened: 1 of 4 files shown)


.github/workflows/release-azure-containerapp-v1.yaml
```diff
@@ -138,16 +138,46 @@ jobs:
...
         run: |
           set -euo
 
-          current_revision=$(az containerapp revision list \
+          container_app=$(az containerapp show \
             --name "$CONTAINER_APP_NAME" \
-            --query 'reverse(sort_by([].{Revision:name,Active:properties.active,Created:properties.createdTime}[?Active!=`false`], &Created))| [0].Revision' -o tsv)
+            --output json)
+
+          current_revision=$(echo "$container_app" | jq -r '
```
krusty93, line 150: **question**: What does this code block do?

gunzip: this code sorts the existing instances based on the percentage of traffic they serve and selects the first one, falling back to latest if all are at 0%

PR #1661, opened 2026-04-23
Update Role Assignment module to add Azure Managed Redis role

kin0992:
Add Azure Managed Redis (AMR) support via a new `managed_redis` input variable, and raise the minimum `azurerm` provider version from `3.114` to `4.60`.

The `managed_redis` input accepts a list of objects with `id` (AMR resource ID), `role` (`reader`, `writer`, or `owner`), and `description`. Role mapping:

- `reader` → _Azure Managed Redis Reader_ (control-plane read-only)
- `writer` → data-plane `default` access policy (Redis commands)
- `owner`  → _Azure Managed Redis Contributor_ (control-plane) + data-plane `default` access policy

Resource IDs are validated using `provider::azurerm::parse_resource_id`. Writer and owner assignments targeting the same AMR id are deduplicated (both share the `default` data-plane access policy).

Legacy `redis` input (Azure Cache for Redis) is unchanged and remains backward compatible.

Closes CES-1960


infra/modules/azure_role_assignments/variables.tf
```diff
@@ -59,6 +59,33 @@ EOT
...
 }
 
+variable "managed_redis" {
+  description = <<EOT
+List of data-plane access policy assignments for Azure Managed Redis (AMR) instances.
+
+REQUIRED FIELDS:
+- name: Name of the Azure Managed Redis instance
+- resource_group_name: Resource group containing the instance
+- role: Permission level - MUST be one of: "writer", "owner". Both roles are
```
kin0992: **thought**: there are no such roles as Contributor, Reader, and so on. Maybe we can remove the `role` parameter since different inputs provide the very same outcome.

Krusty93:
issue: control plane RBAC roles are missing. I think they can be mapped as follow:
- reader: `Azure Managed Redis Reader`
- owner: `Azure Managed Redis Contributor`

These roles can be useful to both Entra ID groups and managed identities (GitHub workflows).
Meanwhile `writer` assigns the data plane policy

This pattern is just a proposal

infra/modules/azure_role_assignments/modules/managed_redis/variables.tf
```diff
@@ -0,0 +1,19 @@
...
+  description = <<EOT
+List of Azure Managed Redis (AMR) instance resource IDs to grant data-plane access to.
+
+Each entry is a full Azure resource ID, e.g.:
+  /subscriptions/{subId}/resourceGroups/{rgName}/providers/Microsoft.Cache/redisEnterprise/{name}
+
+AMR assigns the built-in "default" access policy (full data-plane access).
+There are no role choices; access is binary (granted or not).
+EOT
+  type        = list(string)
```
mamu0: [suggestion(non-blocking)](https://pullpo.io/cc?l=suggestion&d=non-blocking) 
You could follow the same pattern as the other resources by composing the **ID** within the module. Typically, the variable with the module name (in this case `managed_redis`) contains the main information (name, rg...), which, combined with the `subscription_id` variable, in the module's `locals` creates the list of IDs you need

```suggestion
  type = list(object({
    name                = string
    resource_group_name = string
  }))
```

> This snippet is just for reference. It requires the subscription_id variable to be defined and the locals to be updated to work

mamu0:
[suggestion(blocking)](https://pullpo.io/cc?l=suggestion&d=blocking) 
@kin0992 It's probable that the Terraform validation workflow [failure](https://github.com/pagopa/dx/actions/runs/24843632809/job/72724430799?pr=1661) is caused by an outdated local version of **terraform-docs**. Try to update it and re-run the pre-commit

infra/modules/azure_role_assignments/modules/managed_redis/providers.tf
```diff
@@ -0,0 +1,8 @@
+terraform {
+  required_providers {
+    azurerm = {
+      source  = "hashicorp/azurerm"
+      version = "~> 4.23"
```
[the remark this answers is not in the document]
kin0992: Actually, the `azurerm_managed_redis_access_policy_assignment` has been introduced in version 4.60 (https://github.com/hashicorp/terraform-provider-azurerm/releases/tag/v4.60.0), so we cannot use the azurerm provider v3.

infra/modules/azure_role_assignments/modules/managed_redis/variables.tf
kin0992: yes, I see that pattern, but I decided to try a new way to (perhaps) improve the DX.

I think the resource ID is always available to a developer, so using the `parse` function of the azure provider, we can extract both `name` and `resource_group_name` of that resource.
And since this is new functionality is tied to Azure Managed Redis, the input is validated to make sure the id provided is related to AMR.

Doing this, we let developers to pass just the resource ID that is required to grant permissions.

I know I haven't followed the already existing pattern, but I wanted to explore some new ways to try to improve the DX. Take a look at it, and I am open to feedback. As far as I understand this, there are no drawbacks of this solution; we are just hiding some "complexity" to parse the incoming ID to extract the required fields. But maybe I am missing something.

You can see an [example in the dx-playground](https://github.com/pagopa/dx-playground/pull/375/changes#diff-328a0fe048fff5baf1f3f87e5c9110fe94e74d9cbdf7f54757035a0fb84d03f4R106-R109).

.nx/version-plans/version-plan-1776897311794.md
```diff
@@ -0,0 +1,26 @@
+---
+azure_role_assignments: minor
```
kin0992: **issue**: since this set a new lower limit to the azurerm provider, this should be a major

infra/modules/azure_role_assignments/modules/managed_redis/variables.tf
```diff
@@ -0,0 +1,25 @@
...
+  description = "The ID of the principal to which assign roles. It can be a managed identity."
+  type        = string
+}
+
+variable "managed_redis" {
+  description = <<EOT
+List of Azure Managed Redis (AMR) role assignments.
+
+Each entry contains:
+- id:   Full Azure resource ID of the AMR instance
```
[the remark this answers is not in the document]
kin0992: @copilot, validation is performed upfront, not in the inner module

infra/modules/azure_role_assignments/main.tf
```diff
@@ -30,6 +30,14 @@ module "redis" {
   redis           = var.redis
 }
 
+module "managed_redis" {
+  source = "./modules/managed_redis"
+  count  = length(var.managed_redis) > 0 ? 1 : 0
```
christian-calabrese: [nitpick](https://pullpo.io/cc?l=nitpick) 
It might not be a problem, but this adds another index to the resource address that will always be `[0]`
```suggestion
```

NB: in case of acceptance of this comments, there are some test assertions to be adjusted by removing the `[0]` index

infra/modules/azure_role_assignments/modules/managed_redis/providers.tf
```diff
@@ -0,0 +1,8 @@
+terraform {
+  required_providers {
+    azurerm = {
+      source  = "hashicorp/azurerm"
+      version = "~> 4.60"
```
krusty93, line 5: nitpick: Would it be possible to set this value for all the sub modules?

infra/modules/azure_role_assignments/providers.tf
```diff
@@ -2,7 +2,7 @@ terraform {
   required_providers {
     azurerm = {
       source  = "hashicorp/azurerm"
-      version = ">= 3.114, < 5.0"
+      version = ">= 4.60, < 5.0"
```
krusty93: nitpick: Can we set now `~> 4.60`?

Krusty93 (approved):
A couple of nitpicks, but it looks ok

infra/modules/azure_role_assignments/providers.tf
kin0992: So, in general I agree with that, but I want to discuss it:
changing the provider min version would make as bump a major (and it's ok for me).

Another possible way could be to remove the lower requirement of the terraform version in the root module (`infra/modules/azure_role_assignments/providers.tf` file) and just set the min version in every inner module.
In the submodules (like `managed-redis`, `key-vault` etc) we could add a `count` based on the cardinality of the input: e.g. with this input
```hcl
module "roles" {
  source  = "pagopa-dx/azure-role-assignments/azurerm"
  version = "~> 1.2"
  ...
  apim = [
    {
      name                = module.apim.name
      resource_group_name = module.apim.resource_group_name
      description         = "..."
      role                = "reader"
  }]
  key_vault = []
}
```
means `apim` has cardinality 1 and `key_vault` has cardinality 0, so adding the `count` condition in every submodule has the consequences to init just the `apim` submodule to be executed, but not the `key_vault`.

By adding that, our submodules (`apim`, `cosmos`, ...) depend only on the version specified in their `providers.tf` file; in other words, we are saying we could support both azurerm v3 and v4. But if we add that constraint in the `providers.tf` file in the module's root, we are just saying we are dropping support for v3 provider version.

I hope I explained it clearly. What do you think @Krusty93?
cc @pagopa/engineering-team-devex 

Honestly, I do not have a strong opinion on this.

gunzip: sounds nice!

krusty93: it works, but I'd go straightly to v4 support. None of our customers are using v3 anymore and they can use current and vnext module's version at the same time. In short, it's not a big deal to me: no work is required until you use the new Redis service (1 minute spent on `terraform init -upgrade`).

However, you already bumped the min version (I agree with it), my comment was about syntax only

PR #1663, opened 2026-04-23
Add tflint and docs target to terraform nx plugin

mamu0:
Added `tflint` and `terraform-docs` targets to the Nx Terraform plugin. Both targets are auto-injected, `tflint` both for `application` and `library` project types if the respective config file are present in the monorepo root (`.tflint.hcl`), `terraform-docs` is active everytime only for `library` project type. 

### How to Try the New Targets

To try the new **targets**, you'll need to add the plugin to your `nx.json` file:
```json
  "plugins": [
    {...},
    {
      "plugin": "@pagopa/nx-terraform-plugin"
    }
  ]
```
First, run the following commands in your local terminal:

```bash
export NX_DAEMON=false
export NX_CACHE_PROJECT_GRAPH=false
```

Then, to verify if **tflint** was added correctly, run:

`pnpm nx show projects --withTarget tflint --json`

For **terraform docs**, run:

`pnpm nx show projects --withTarget terraform-docs --json`

These commands will output a JSON list of projects that include the target. To deep-dive into a specific project (e.g., _azure_cdn_), use:

`pnpm nx show project azure_cdn --json`

Under the targets section, you should see the following configurations:
<details>
<summary>"targets":</summary>

```json
"tflint": {
  "cache": true,
  "inputs": ["default", "examples", "tests", "{workspaceRoot}/.tflint.hcl"],
  "executor": "nx:run-commands",
  "options": {
    "args": [
      "--disable-rule=terraform_required_version",
      "--disable-rule=terraform_required_providers",
      "--config",
      "../../../.tflint.hcl"
    ],
    "cwd": "infra/modules/azure_cdn",
    "command": "tflint"
  },
  "configurations": {},
  "parallelism": true
},
"terraform-docs": {
  "cache": true,
  "inputs": ["default", "{projectRoot}/README.md"],
  "outputs": ["{projectRoot}/README.md"],
  "executor": "nx:run-commands",
  "options": {
    "args": [
      "--output-file",
      "README.md",
      "--output-mode",
      "inject",
      "--hide",
      "providers",
      "--lockfile=false",
      "."
    ],
    "cwd": "infra/modules/azure_cdn",
    "command": "terraform-docs markdown table"
  },
  "configurations": {},
  "parallelism": true
},
```

</details>

Finally, to test the actual command execution, run:

- `nx run azure_cdn:tflint`
- `nx run azure_cdn:terraform-docs`

Resolves: CES-1947 CES-1948


packages/nx-terraform-plugin/src/project.ts
```diff
@@ -37,15 +37,48 @@ const getProjectType = (root: string): ProjectType => {
     : "application";
 };
 
+const getRootConfigPath = (root: string, configFileName: string) =>
+  path.relative(root, configFileName) || configFileName;
```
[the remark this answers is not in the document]
mamu0: @copilot I checked this locally by changing the process `cwd` and by comparing the current implementation with the proposed helper. In our case both resolve the same `--config path` for the **roots** handled by this plugin, so I do not think this is a correctness issue. For that reason I would keep the current implementation.

packages/nx-terraform-plugin/src/__tests__/project.test.ts
```diff
@@ -25,6 +27,54 @@ const expectedNamedInputs = {
...
+  options: {
+    args: [
+      "--disable-rule=terraform_required_version",
+      "--disable-rule=terraform_required_providers",
+      "--config",
+      path.relative(root, ".tflint.hcl") || ".tflint.hcl",
+    ],
+    cwd: "{projectRoot}",
+  },
+});
```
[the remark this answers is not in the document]
mamu0: answer [here](https://github.com/pagopa/dx/pull/1663#discussion_r3129885383)

packages/nx-terraform-plugin/src/index.ts
```diff
@@ -11,12 +12,20 @@ import { getTerraformProjectFiles } from "./project-file.ts";
...
   // We create a terraform project for each directory containing .tf files
   "**/*.tf",
   async (configFiles, options, context) => {
     const opts = parseOptions(options);
+    const hasRootTflintConfig = existsSync(
+      path.join(context.workspaceRoot, rootTflintConfigFileName),
+    );
+    const hasRootTerraformDocsConfig = existsSync(
+      path.join(context.workspaceRoot, rootTerraformDocsConfigFileName),
+    );
```
lucacavallaro: **issue:** `createNodesV2` is called for `*.tf` file found in the workspace, calling `existSync` for every file found is too expensive and redundant (for example, in the `dx` repo will be called 969 times, to obtain the same output!). 

This code will be potentially called on every `nx` command.

---

Have you tried to implement this lookup by using another `createNodesV2` block? Theoretically, you should create new `Projects` block with the `lint` and `docs` target that `Nx` will merge with the current ones.

packages/nx-terraform-plugin/src/project.ts
```diff
@@ -37,15 +37,48 @@ const getProjectType = (root: string): ProjectType => {
...
+    lint: [
+      "--disable-rule=terraform_required_version",
+      "--disable-rule=terraform_required_providers",
+      "--config",
+      rootTflintConfigPath,
+    ],
   };
 
-  const cwd = "{projectRoot}";
+  const projectCwd = "{projectRoot}";
```
lucacavallaro: **question:** why you renamed this variable? it was named after the corresponding nx option. It is an idiomatic pattern on js.

packages/nx-terraform-plugin/src/project.ts
```diff
@@ -37,15 +37,48 @@ const getProjectType = (root: string): ProjectType => {
...
+  const terraformInputs = ["default", "examples", "tests"];
+  const terraformDocsInputs = [
+    "default",
+    "{projectRoot}/.terraform.lock.hcl",
+    "{projectRoot}/README.md",
+    "{workspaceRoot}/.terraform-docs.yml",
+  ];
```
lucacavallaro: **suggestion:** this variable is redundant, as it is used only by one target

packages/nx-terraform-plugin/src/project.ts
```diff
@@ -37,15 +37,48 @@ const getProjectType = (root: string): ProjectType => {
...
+      "--disable-rule=terraform_required_version",
+      "--disable-rule=terraform_required_providers",
+      "--config",
+      rootTflintConfigPath,
+    ],
   };
 
-  const cwd = "{projectRoot}";
+  const projectCwd = "{projectRoot}";
+  const terraformInputs = ["default", "examples", "tests"];
```
lucacavallaro: **suggestion:** rename to inputs to follow the nx option name

```suggestion
  const inputs = ["default", "examples", "tests"];
```

packages/nx-terraform-plugin/src/project.ts
```diff
@@ -37,15 +37,48 @@ const getProjectType = (root: string): ProjectType => {
...
+  hasRootTerraformDocsConfig: boolean,
 ): Record<string, TargetConfiguration> => {
+  const rootTflintConfigPath = getRootConfigPath(root, ".tflint.hcl");
   const defaultArgs = {
     fmt: ["-list=true", "-recursive=true"],
+    lint: [
+      "--disable-rule=terraform_required_version",
+      "--disable-rule=terraform_required_providers",
+      "--config",
+      rootTflintConfigPath,
```
lucacavallaro: **suggestion:** `defaultArgs.lint` is used only by one configuration, so it is not a "default". It should be moved to the target definition.

packages/nx-terraform-plugin/src/project.ts
```diff
@@ -37,15 +37,48 @@ const getProjectType = (root: string): ProjectType => {
...
+};
+
 const getTargets = (
   opts: TerraformPluginOptions,
+  root: string,
   projectType: ProjectType,
+  hasRootTflintConfig: boolean,
+  hasRootTerraformDocsConfig: boolean,
 ): Record<string, TargetConfiguration> => {
+  const rootTflintConfigPath = getRootConfigPath(root, ".tflint.hcl");
```
lucacavallaro, line 49: **question:** is it not equivalent to `{workspaceRoot}/.tflint.hcl`?

lucacavallaro (changes requested):

.terraform-docs.yml
krusty93, line 1: [question](https://pullpo.io/cc?l=question) 
Is this file really needed? When and why anyone should customize these options?

mamu0: I discussed this with @lucacavallaro, and we found that using a file as a trigger was the most straightforward way to signal that `terraform-docs` should be used in the monorepo.

As mentioned in the PR description, this file uses default configurations, so it doesn't change the tool's behavior (in **dx** monorepo); it acts as a flag for the plugin to add the target.
Similar to `.tflint.hcl`, we could include it in the DX CLI as a template file for new monorepos.

packages/nx-terraform-plugin/src/project.ts
mamu0: i restored it here [0ea6182](https://github.com/pagopa/dx/pull/1663/commits/0ea618278a6444d854414ec0d0542db5ea14142d)

packages/nx-terraform-plugin/src/project.ts
mamu0: thanks, resolved here [0ea6182](https://github.com/pagopa/dx/pull/1663/commits/0ea618278a6444d854414ec0d0542db5ea14142d)

packages/nx-terraform-plugin/src/project.ts
mamu0: done [0ea6182](https://github.com/pagopa/dx/pull/1663/commits/0ea618278a6444d854414ec0d0542db5ea14142d)

packages/nx-terraform-plugin/src/project.ts
mamu0: done [0ea6182](https://github.com/pagopa/dx/pull/1663/commits/0ea618278a6444d854414ec0d0542db5ea14142d)

packages/nx-terraform-plugin/src/project.ts
mamu0: It is not equivalent here because this value is used in the tflint CLI args, not only in Nx inputs. The command runs with cwd set to `{projectRoot}`, so the config path has to be resolved from the project directory. I also verified locally that the relative path from `projectRoot` works correctly in this setup, while using the workspace token in command args is not something I would rely on here.

packages/nx-terraform-plugin/src/index.ts
mamu0: I checked this again against the Nx `createNodesV2` API. In the current implementation these two `existsSync` calls run once per `createNodesV2` invocation, before `createNodesFromFiles` iterates over `configFiles`, not once per matched `.tf` file. So here the cost is constant per graph build, not proportional to the number of Terraform files.
For that reason I would keep this in a single block unless we want to split it for structural reasons rather than performance.

lucacavallaro: good, better than I thought! 

we can keep in this way, but I suggest two changes:

- (the major one) use async I/O instead of `existSync` (which blocks the event loop), you can try with `fs.access` but maybe there are better alternatives

- minor/nitpick one: I would put the config file name inline in path.join, to make it easier to read

.terraform-docs.yml
lucacavallaro: isn't there a way to make it shorter?

packages/nx-terraform-plugin/src/index.ts
mamu0: Thanks for the suggestions!
Applied here [3d1072a](https://github.com/pagopa/dx/pull/1663/commits/3d1072ae03436442c20efe35359bd56fcb6095be). I replaced the synchronous existence checks with async access-based checks and inlined the config file names in path.join as suggested.

.terraform-docs.yml
mamu0: Check the official [documentation](https://terraform-docs.io/user-guide/configuration/)

I ran some tests to make sure this file wouldn't introduce any changes; as it stands, it’s using all default values

lucacavallaro (approved):

.terraform-docs.yml
lucacavallaro: Uhm so if it using all the default values, it means that can be omitted?

If so, since we can't check if the actual binary is installed, we can always add docs without the config

packages/nx-terraform-plugin/src/index.ts
```diff
@@ -3,6 +3,7 @@ import {
   createNodesFromFiles,
   CreateNodesV2,
 } from "@nx/devkit";
+import { access } from "node:fs/promises";
```
lucacavallaro: nitpick: use namespace for node modules 

`import fs from "node:fs/promises"`, as we did with path

packages/nx-terraform-plugin/src/project.ts
```diff
@@ -100,12 +119,56 @@ const getTargets = (
...
+            "--config",
+            rootTflintConfigPath,
+          ],
+          cwd,
+        },
+      },
+    ]);
+  }
+
+  if (hasRootTerraformDocsConfig) {
```
lucacavallaro: issue: docs should be added only when the projectType is library - it should be removed from apps

lucacavallaro (changes requested):

packages/nx-terraform-plugin/src/project.ts
mamu0: Good catch! Resolved [e88a427](https://github.com/pagopa/dx/pull/1663/commits/e88a42765935a660dc56c10b8bc21f466f63bdde)

packages/nx-terraform-plugin/src/index.ts
mamu0: done [e88a427](https://github.com/pagopa/dx/pull/1663/commits/e88a42765935a660dc56c10b8bc21f466f63bdde)

.terraform-docs.yml
mamu0: I could remove the config file and remove the check, in this way If `terraform-docs` isn't installed, the target would still exist but it would fail. 
To avoid the overhead of implementing and maintaining binary checks.

lucacavallaro: @mamu0 it is ok, since the plugin does not check for the existence of `terraform` and `tflint` as well, so they can fail at runtime (it is the same behavior followed for official Nx plugins)

packages/nx-terraform-plugin/src/__tests__/options.test.ts
```diff
@@ -7,8 +7,10 @@ describe("parseOptions", () => {
     expect(parseOptions(undefined)).toEqual({
       applyTargetName: "tf-apply",
       consoleTargetName: "tf-console",
+      docsTargetName: "tf-docs",
       formatTargetName: "tf-fmt",
       initTargetName: "tf-init",
+      lintTargetName: "tf-lint",
```
lucacavallaro: **suggestion(non-blocking):** use `terraform-docs` and `tflint` as default target names

_rationale:_ `tf-*` should be reserved for terraform CLI commands, target names should be explicit about the fact that they are using different tools

we are about to rename other js scripts following this convention, the currents `lint` and `test` will be renamed to `eslint` and `vitest`

lucacavallaro (approved):

.terraform-docs.yml
mamu0: i've updated the target in this commit [e09a523](https://github.com/pagopa/dx/pull/1663/commits/e09a523df61a5b083a933ef06ee4aad0e545e810)

packages/nx-terraform-plugin/src/__tests__/options.test.ts
mamu0: done in this commit [f1a6562](https://github.com/pagopa/dx/pull/1663/commits/f1a6562c80ae9d8738f63e80ca8f50bcdaafb052)

PR #1664, opened 2026-04-23
Fix interpretation of environment variables in infra workflows

### LINKED ISSUE #1654: [Bug]: handle env_vars correctly in infra_plan/apply (avoid malformed GITHUB_ENV values)
### What are you trying to do?

Pass TF_VAR_image_tag through env_vars from caller workflows to DX reusable Terraform workflows (infra_plan/infra_apply) so Terraform uses the SHA-based image tag.

### What's the expected result?

TF_VAR_image_tag is written to GITHUB_ENV exactly as provided (e.g. sha-f21eae5), and terraform plan/apply resolves image changes correctly.

### What happened instead?

The reusable workflow treated env_vars as a Bash array and escaped values with printf "%q", producing malformed values like sha-f21eae5\[@\], which then appeared in Terraform output.

christian-calabrese:
This change fixes how `env_vars` are propagated in the reusable Terraform workflows. The input is now handled as a multiline list of `env=value` entries and written to `GITHUB_ENV` without shell escaping, avoiding malformed `TF_VAR_*` values during plan and apply.

Resolves CES-1965
Fixes #1654


christian-calabrese:
Some tests carried out in the dx-playground repo:
- Valid input with complex values ([here](https://github.com/pagopa/dx-playground/actions/runs/24829107269/job/72672928145#step:11:385))
- Invalid values with graceful error ([here](https://github.com/pagopa/dx-playground/actions/runs/24828155065/job/72669596410))

.github/workflows/infra_plan.yaml
```diff
@@ -303,7 +320,8 @@ jobs:
         id: set-plan-output
         if: always()
         env:
-          WORKFLOW_URL: "${{ github.server_url }}/${{ github.repository }}/actions/runs/${{ github.run_id }}"
+          WORKFLOW_URL: "${{ github.server_url }}/${{ github.repository
```
krusty93, line 323: question: does it still work?

christian-calabrese: Yep it [still works](https://github.com/pagopa/dx-playground/actions/runs/24829107269/job/72672928145#step:13:143)!

Krusty93 (approved):

PR #1676, opened 2026-04-28
Substitute multi role assignments with merged roles in bootstrapper

christian-calabrese:
Resolves CES-1966
Depends on #1684


krusty93 (approved):

krusty93:
As already discussed via Slack, before merging I'd like to discuss if this major should also remove variables like `apim_id`

PR #1680, opened 2026-04-29
Bootstrapper workflow uses GitHub app token

christian-calabrese:
Built-in GitHub PAT cannot write environment secrets. For this reason, a new GitHub PAT must be generated using the GitHub App.

Resolves CES-1983


.nx/version-plans/version-plan-1777458544889.md
```diff
@@ -0,0 +1,6 @@
+---
+"@pagopa/dx-cli": minor
```
mamu0: [suggestion(non-blocking)](https://pullpo.io/cc?l=suggestion&d=non-blocking) 
Although this is an additional feature, it's necessary for the current setup to work. We should consider it a patch to fix the bootstrap workflow.

```suggestion
"@pagopa/dx-cli": patch
```

apps/cli/src/adapters/azure/cloud-account-service.ts
```diff
@@ -245,6 +245,7 @@ export class AzureCloudAccountService implements CloudAccountService {
     const parameters = {
       location: cloudAccount.defaultLocation,
       tags: {
+        CreatedBy: "DX CLI",
```
mamu0, line 248: [question(non-blocking)](https://pullpo.io/cc?l=question&d=non-blocking) 
Was this added for a specific reason?

mamu0:
The workflow and overall logic look good to me, I only see one error in the `format:check` step of the [code review](https://github.com/pagopa/dx/actions/runs/25308572236/job/74189947140?pr=1680#step:9:219). 
@christian-calabrese  Could you check if it's related to the changes in this PR?

apps/cli/src/adapters/azure/cloud-account-service.ts
christian-calabrese: Usually, it is set to Terraform or ARM (if the resource has been created from the portal). With @Krusty93 we thought about this solution since the actual creator is the DX CLI

mamu0 (approved):

PR #1682, opened 2026-04-29
Document dx-cli side effects

lucacavallaro:
Enhance `dx init` and `dx add environment` output results, to add contextual documentation and document all the relevant side effects.

#### `dx init` 
Added an interactive confirmation step to optionally skip GitHub repo creation

image

#### `dx add environment`
Added an environment initialization “side effects” summary

image

----

Closes CES-1819


apps/cli/src/adapters/commander/commands/init.ts
```diff
@@ -47,8 +60,36 @@ const withSpinner = <T>(
...
+      `${step++}. Create the GitHub repository by applying the Terraform config scaffolded at ${chalk.cyan(`${payload.repoName}/infra/repository`)}:`,
+    );
+    console.log(
+      `       ${chalk.cyan(`cd ${payload.repoName}/infra/repository && terraform init && terraform apply`)}`,
+    );
+    console.log(
+      `${step++}. Push the scaffolded code to the newly created repository:`,
+    );
+    console.log(
+      `       ${chalk.cyan(`cd ${payload.repoName} && git init && git remote add origin <url> && git push`)}`,
```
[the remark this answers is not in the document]
lucacavallaro: `process.chdir` only affect the node.js process, not the user current folder

apps/cli/src/adapters/plop/generators/environment/prompts.ts
```diff
@@ -293,4 +295,54 @@ export const getCloudAccountToInitialize = (
...
+  const missingBackend = initStatus.issues.some(
+    (issue) => issue.type === "MISSING_REMOTE_BACKEND",
+  );
+
+  const sections: string[] = [];
+
+  for (const account of accountsToInit) {
+    sections.push(
+      [
+        chalk.bold.cyan(`  Azure subscription "${account.displayName}":`),
```
gunzip, line 317: question: is it a deliberate choice to pair the output with azure?

apps/cli/src/adapters/plop/generators/environment/prompts.ts
```diff
@@ -293,4 +295,54 @@ export const getCloudAccountToInitialize = (
...
+        `    • Common Key Vault storing the GitHub runner app credentials`,
+      ].join("\n"),
+    );
+  }
+
+  if (missingBackend) {
+    sections.push(
+      [
+        chalk.bold.cyan(`  Terraform remote backend:`),
+        `    • Azure resource group and Storage Account for the Terraform state`,
```
gunzip, line 330: ditto

gunzip:
nitpick

> Don't include images or links in the description, as they add noise to the Git history since the PR description is used as summary of the merge commit. Use a comment to provide more context instead.

ref. https://github.com/pagopa/dx/security/dependabot

lucacavallaro:
> nitpick
> 
> > Don't include images or links in the description, as they add noise to the Git history since the PR description is used as summary of the merge commit. Use a comment to provide more context instead.
> 
> ref. https://github.com/pagopa/dx/security/dependabot

This is the case with auto merge, but our current setup doesn't work in this way, the squashed commit is only prefilled with the pr body but github lets you edit it before merging.

I always strip unnecessary markdown and images before merging!

apps/cli/src/adapters/plop/generators/environment/prompts.ts
lucacavallaro: yes, they are azure details and we don't know yet the equivalent resources for aws setups

gunzip (approved):

PR #1685, opened 2026-04-30
Update __swap_versions to set traffic to 100% instead of reverting

lucacavallaro:
Fix an issue which prevented the incremental rollout of containers to Container App instances with the deployment mode set to `Multiple`

### DIFF
--- a/.github/workflows/release-azure-containerapp-v1.yaml
+++ b/.github/workflows/release-azure-containerapp-v1.yaml
@@ -26,7 +26,8 @@ on:
         default: PagoPA
         description: Authors names to use as image label
       build_args:
-        description: List of build arguments to use for Dockerfile build, given in env=value format.
+        description: List of build arguments to use for Dockerfile build, given in
+          env=value format.
         type: string
         required: false
       build_platforms:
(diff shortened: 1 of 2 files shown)


gunzip (approved):

.github/workflows/release-azure-containerapp-v1.yaml
```diff
@@ -243,15 +245,17 @@ jobs:
           echo "ca_id=$container_app_id" >> $GITHUB_OUTPUT
 
       - name: Check for New Revision Errors and Increment Traffic
-        uses: pagopa/dx/actions/incremental-rollout@main
-        if: ${{ steps.get_revision_mode.outputs.revision_mode == 'Multiple' && steps.create_new_revision.outputs.skip_remaining_steps == 'false' }}
+        uses: pagopa/dx/actions/incremental-rollout@fixes/incremental-rollout
```
christian-calabrese: [issue(blocking)](https://pullpo.io/cc?l=issue&d=blocking) 
Please remember to re-set the ref to `main` before merging

krusty93: done

lucacavallaro: tip: instead of reverting with a new commit, you could have dropped that commit (via `rebase -> drop`)! 

it was a separate commit on purpose!

PR #1688, opened 2026-05-04
Ignore CHANGELOG.md files in Prettier formatting

kin0992:
Ignores auto-generated CHANGELOG.md files in format and format:check scripts to prevent unnecessary formatting changes.

## Changes
- Add CHANGELOG.md to root .prettierignore
- Add CHANGELOG.md to all project-level .prettierignore files

## Testing
- ✅ format:check passes for all 16 projects
- ✅ CHANGELOG.md files are no longer checked
- ✅ No regression to existing format functionality


packages/eslint-config/.prettierignore
```diff
@@ -1 +1 @@
-dist
\ No newline at end of file
+distCHANGELOG.md
```
kin0992: **issue**: wrong formatting

apps/cli/templates/monorepo/.prettierignore
```diff
@@ -2,4 +2,4 @@
 **/*.hbs
 
 # Ignore built files
-dist
\ No newline at end of file
+distCHANGELOG.md
```
kin0992: **issue**: wrong formatting

.nx/version-plans/ignore-changelog.md
lucacavallaro, line 1: **question:** Can't we exclude `CHANGELOG.md` just from the root `.gitignore`? Having to manually exclude in each project is easy to forgot and will 100% creates problem in future.

.nx/version-plans/ignore-changelog.md
```diff
@@ -0,0 +1,20 @@
+---
+'@pagopa/opex-dashboard': patch
```
lucacavallaro: **issue (or maybe a thought):** Changing the `.gitignore` is not worth for a version plan. We don't have to always include a version plan for every change, but only for that changes we want to record o release. 

I suggest to remove this version plan file.

.nx/version-plans/ignore-changelog.md
kin0992: ~~sounds reasonable to me! On it~~
sorry, i replied to the wrong comment. anyway, i'll try and if feasible, i'll do it

.nx/version-plans/ignore-changelog.md
kin0992: sounds reasonable to me! On it

.nx/version-plans/ignore-changelog.md
kin0992: We investigated this thoroughly. Unfortunately, **Prettier does not inherit `.prettierignore` from parent directories** when a project has its own — it only reads the one in the current working directory. Since our `format:check` scripts run `prettier --check .` from each project directory, the root `.prettierignore` is simply not seen.

We also tried:
- **`--ignore-path`** pointing to the root file — works, but replaces the defaults so you need `--ignore-path .gitignore --ignore-path .prettierignore --ignore-path ../../.prettierignore` in every script (same boilerplate problem, just moved to `package.json`).
- **Nx `targetDefaults`** to centralize the command — Nx can't override the executor for targets inferred from `package.json` scripts (project-level config takes highest priority).
- **Prettier config (`.prettierrc`)** — doesn't support ignore path configuration at all.

So the per-project `.prettierignore` entry is the simplest and most reliable approach. The `cli/templates/monorepo` template already includes `CHANGELOG.md`, so new projects scaffolded from it will have it out of the box.

lucacavallaro: ok

mamu0 (approved):
LGTM

PR #1708, opened 2026-05-07
Implement additional permission actions for merged roles

christian-calabrese:
Introduce support for optional `additional_actions` and `additional_data_actions` to extend merged custom roles with extra control-plane and data-plane permissions, ensuring proper validation and deduplication of permissions.


infra/modules/azure_merge_roles/README.md
```diff
@@ -70,6 +78,10 @@ The `scope` input can also point to a management group, for example:
 scope = "/providers/Microsoft.Management/managementGroups/dx-platform"
 ```
 
+Use `additional_actions` only for extra control-plane grants that are not already included in `source_roles`. The module deduplicates them against the merged source actions and applies the same permissive overlap policy used for `not_actions`.
```
gunzip, line 91: question: why this split? reading the documentation, it's not clear exactly what 'additional' refers to. These are concepts that those using the module might not be familiar with, right? I would try to clarify this as much as possible for the casual reader.

gunzip: maybe this complexity of use could be hidden behind some 'use case'?

infra/modules/azure_merge_roles/locals.tf
```diff
@@ -43,51 +57,83 @@ locals {
...
-              length(regexall("^${replace(replace(lower(excluded_action), ".", "[.]"), "*", ".*")}$", lower(allowed_action))) > 0 ||
-              (
-                length(regexall("\\*", allowed_action)) > 0 &&
-                length(regexall("\\*", excluded_action)) > 0 &&
+      if !(
+        anytrue([
+          for permission in local.normalized_source_permissions : (
+            !contains([for denied_action in permission.not_actions : lower(denied_action)], lower(excluded_action)) &&
+            anytrue([
+              for allowed_action in permission.actions : (
```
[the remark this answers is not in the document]
gunzip: agree, there's too much complexity added. copilot suggest to refactor using locals and something like this:

```hcl
locals {
  # 1. Map every unique action to its regex pattern once
  action_regex_map = {
    for a in distinct(concat(
      local.merged_permissions.actions,
      local.merged_permissions.data_actions,
      flatten(local.normalized_source_permissions[*].not_actions),
      flatten(local.normalized_source_permissions[*].not_data_actions)
    )) : a => "^${replace(replace(lower(a), ".", "[.]"), "*", ".*")}$"
  }

  # 2. Simplified filtering logic using can(regex())
  merged_permissions_refined = {
    actions      = local.merged_permissions.actions
    data_actions = local.merged_permissions.data_actions

    not_actions = sort([
      for excl in distinct(flatten(local.normalized_source_permissions[*].not_actions)) : excl
      if !anytrue([
        for allow in local.merged_permissions.actions : 
          can(regex(local.action_regex_map[allow], lower(excl))) || 
          can(regex(local.action_regex_map[excl], lower(allow)))
      ])
    ])

    not_data_actions = sort([
      for excl in distinct(flatten(local.normalized_source_permissions[*].not_data_actions)) : excl
      if !anytrue([
        for allow in local.merged_permissions.data_actions : 
          can(regex(local.action_regex_map[allow], lower(excl))) || 
          can(regex(local.action_regex_map[excl], lower(allow)))
      ])
    ])
  }
}
```

gunzip (changes requested):

infra/modules/azure_merge_roles/locals.tf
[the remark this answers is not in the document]
christian-calabrese: I just split the locals to have a better view

infra/modules/azure_merge_roles/README.md
christian-calabrese: I just split the locals to have a better view

infra/modules/azure_merge_roles/README.md
```diff
@@ -13,9 +13,21 @@ It is designed for the RBAC reduction strategy: define a smaller set of reusable
...
 - Supports roles that include `not_actions` and `not_data_actions`.
 - Deduplicates merged permissions so the generated role definition remains stable.
+- Supports caller-provided `additional_actions` when the merged role must grant extra control-plane permissions beyond the merged source roles.
+- Supports caller-provided `additional_data_actions` when the merged role must grant extra data-plane permissions beyond the merged source roles.
 - Requires an explicit `scope` and always uses that same scope as the role's only assignable scope.
 - Supports role definitions created at management group scope as well as subscription scope.
 
+## Choosing Additional Permissions
+
+Most callers should start with `source_roles` only and leave both `additional_*` inputs empty.
```
gunzip, line 23: **issue**: this description is still too low-level. how can a user, without inspecting the module's source code, understand whether they need to enable additional permissions? I would suggest revisiting it, aiming to clarify _at a high level_ what kind of permissions are required and for which role or function within the team.

for example: "enable these permissions only if... some event or context user already has"

christian-calabrese: It's a pretty technical module that will hardly be used by ICs. I'd maybe make that explicit instead.

gunzip (approved):

PR #1732, opened 2026-05-11
Enhance GitHub App token handling for terrawiz

gunzip:
Improve the handling of GitHub App tokens by generating a short-lived installation access token for terrawiz, eliminating the need for a separate GITHUB_TOKEN. Add tests to ensure proper environment resolution for terrawiz imports.

### DIFF
--- a/apps/dx-metrics/scripts/lib/__tests__/import-context.test.ts
+++ b/apps/dx-metrics/scripts/lib/__tests__/import-context.test.ts
@@ -74,6 +74,20 @@ describe("buildGitHubAppOctokitAuthOptions", () => {
       privateKey: "-----BEGIN PRIVATE KEY-----\nprivate-key",
     });
   });
+
+  it("preserves a client ID string when building Octokit auth options", () => {
+    expect(
+      buildGitHubAppOctokitAuthOptions({
+        appId: "Iv1.client-id",
+        installationId: 456,
+        privateKey: "-----BEGIN PRIVATE KEY-----\nprivate-key",
+      }),
+    ).toEqual({
+      appId: "Iv1.client-id",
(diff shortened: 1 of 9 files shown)


apps/dx-metrics/scripts/lib/__tests__/import-context.test.ts
```diff
@@ -95,3 +97,51 @@ describe("createOctokitClient", () => {
...
+
+  it("creates an installation token resolver when GitHub App auth is used", async () => {
+    const createInstallationTokenResolver = vi.fn(
+      (githubApp: GitHubAppAuthSettings) => async () =>
+        `ghs_${githubApp.installationId}`,
+    );
+
+    const resolveTerrawizGitHubToken = createTerrawizGitHubTokenResolver(
+      {
+        appId: 123,
```
krusty93, line 144: thought: Perhaps you can use `client_id` instead?

gunzip: do you have some ref?

christian-calabrese: [Here](https://docs.github.com/en/apps/creating-github-apps/authenticating-with-a-github-app/generating-a-json-web-token-jwt-for-a-github-app) look at the issuer field description

gunzip: It now supports both. dx-metrics now resolves GitHub auth in this order: `GITHUB_CLIENT_ID` → `GITHUB_APP_ID` → `GITHUB_TOKEN`.

christian-calabrese (approved):

PR #1736, opened 2026-05-12
Add `publish` target to `nx-terraform-plugin`

lucacavallaro:
[Read the spec file to understand the designed solution](https://github.com/pagopa/dx/blob/d7fc64a6c8ad71214118f430f560d5f0f20b1ee4/docs/superpowers/specs/2026-05-07-nx-terraform-publish-design.md)

**How to test**
1. Pick a module on `infra/modules`
2. Inside the module, write a `module.json` file following the guideline (copy `version` from the `package.json` file). **Important:** set your github username as `github.owner`, in this way the module will be published inside your github account without touching `pagopa-dx` org.
3. Run `nx reset`, to force Nx re-index the changed project
4. Make a change, and version it with `nx release plan -p $MODULE-NAME` where `$MODULE-NAME` can be found calling `nx show projects -p tag:terraform:public`
5. Run `nx release -p $MODULE_NAME`. Nx will bump `module.json` file, write a `CHANGELOG` entry and will export the module in a dedicated repo owned by your personal account.

[Demo](https://github.com/lucacavallaro/terraform-aws-aws-azure-vpn)

Closes CES-1949


packages/nx-terraform-plugin/src/__tests__/index.test.ts
```diff
@@ -0,0 +1,233 @@
...
+import { parseOptions } from "../options.ts";
+
+describe("createNodesV2", () => {
+  it("discovers both terraform and module manifests", () => {
+    expect(createNodesV2[0]).toBe("**/{*.tf,module.json}");
+  });
+});
+
+describe("getDiscoveryStateWithValidation", () => {
+  const createdDirs: string[] = [];
```
gunzip: idea: another option, if you want to avoid a shared variable, is to use `extend` + `onCleanup` to provide test specific context, see https://vitest.dev/guide/test-context.html#setup-and-cleanup-with-oncleanup

gunzip: thoughts: another option would have been to use `memfs` https://main.vitest.dev/guide/mocking/file-system

packages/nx-terraform-plugin/src/adapters/github/octokit.ts
```diff
@@ -0,0 +1,36 @@
+// Provides the concrete Octokit-backed GitHub repository adapter for publish flows.
+
+import { Octokit } from "octokit";
+
+const getGitHubToken = () => process.env.GH_TOKEN ?? process.env.GITHUB_TOKEN;
```
gunzip, line 5: question: does this works even using GitHub Apps instead of PATs?

packages/nx-terraform-plugin/src/executors/publish/__tests__/publish.test.ts
gunzip, line 1: the name of this module diverges from the usual suffix `test.ts`

lucacavallaro: how? it has the `test.ts` suffix

docs/superpowers/specs/2026-05-07-nx-terraform-publish-design.md
```diff
@@ -0,0 +1,270 @@
...
+Implement a plugin-owned publish executor/service invoked by `nx-release-publish` command wiring.
+
+Responsibilities:
+
+1. Consume validated publish metadata from executor inputs/options:
+   - inferred target execution: metadata is injected from discovery,
+   - direct executor invocation: caller must provide metadata explicitly.
+2. Build repo coordinates (`org`, `repo`).
+3. Ensure target repo exists (create if missing).
+4. Perform subtree-based synchronization preserving current history semantics.
```
gunzip: question: what about cloning the whole repo, copying files and commit? `git subtree` may become slow on large monorepos. do we really need to preserve history on the target repo?

this may also limit parallelism in case nx spans multiple `git subtree`tasks (which locks the whole monorepo) in parallel.

docs/superpowers/specs/2026-05-07-nx-terraform-publish-design.md
```diff
@@ -0,0 +1,270 @@
...
+Eligibility errors prevent target inference; execution errors fail the target.
+Manifest inference failures and merged publish option failures are also emitted
+as structured logger warnings so CI and local runs can inspect the raw issue
+list programmatically.
+
+## Testing Strategy
+
+1. **Options/Schema tests**
+   - Validate publish config parsing (`mode`, `publish.github.owner`).
+   - Validate JSON Schema generation from the Zod manifest schema.
```
gunzip: remember that code-review must fail in case the JSON of the schema is not updated

docs/superpowers/specs/2026-05-07-nx-terraform-publish-design.md
lucacavallaro: This is a direct translation of the current workflow 

I don't remember why we decided to preserve the entire history, we can change that but we have to find a new way to extract the right folder

lucacavallaro: @christian-calabrese can we update the flow to keep only the latest commit or an "artificial one"? what do you think?

packages/nx-terraform-plugin/src/executors/publish/__tests__/publish.test.ts
gunzip: it was 'spec.ts' some commits ago

docs/superpowers/specs/2026-05-07-nx-terraform-publish-design.md
christian-calabrese: Yes, the idea here was to keep a coherent history, but it's not really needed. The registry just requires a tag per version.

lucacavallaro: I applied this change. The `nx-release-publish` target now just appends a single commit, with the message `Release $VERSION` for each published version.

The target repo is cloned in a temporary directory, and the updated module code is simply copied to that repo. In this way we don't need to use `git subtree`, which is slow and not distributed by default in every `git` distribution. (For example, it is not included in `git` on macOS)

packages/nx-terraform-plugin/src/adapters/github/octokit.ts
gunzip: @lucacavallaro

packages/nx-terraform-plugin/src/__tests__/index.test.ts
lucacavallaro: Fixed in `index.test.ts`: the temporary workspace cleanup is now registered per test with `onTestFinished`, so the shared `createdDirs` state is gone.

packages/nx-terraform-plugin/src/adapters/github/octokit.ts
lucacavallaro: Checked this against the GitHub API behavior and tightened the adapter accordingly. Organization repositories can be created with a GitHub App installation token when the app has the required admin permission, but user-owned repositories still require user-scoped credentials. The adapter now fails with an explicit message in that user-owner/app-token case instead of surfacing the raw API error.

packages/nx-terraform-plugin/src/adapters/github/octokit.ts
```diff
@@ -0,0 +1,73 @@
+// Provides the concrete Octokit-backed GitHub repository adapter for publish flows.
```
gunzip, line 1: question: the release mechanism for Terraform modules links the task (publishing on GitHub) to the task manager (NX). Is this the approach we’d like to encourage for all reusable tasks we want to distribute? For example, right now the plan and apply Terraform pipelines use some “custom” logic that will need to be moved into a reusable task. Will this be encapsulated and distributed with the NX plugin? I'm asking since this creates a lock-in on NX for all reusable tasks.

lucacavallaro: Do you have an alternative proposal? 

We can extract the task in their own library or move them to DX CLI, my only preference is to implement them in TypeScript instead of bash, to make them easier to test and maintain

gunzip: What about reducing the plugin's surface by placing the tasks in standalone packages? It's not necessary to publish these task packages (for now); we can simply share the NX plugins that use them.

lucacavallaro: Yes, agreed. I will split the task in a separate PR.

gunzip:
lgtm but has conflicts

packages/nx-terraform-plugin/src/executors/publish/schema.json
```diff
@@ -0,0 +1,34 @@
+{
+  "$schema": "https://json-schema.org/schema",
+  "version": 2,
+  "title": "Publish executor",
+  "description": "",
+  "type": "object",
+  "properties": {
+    "projectRoot": {
```
krusty93, line 8:
```suggestion
    "projectPath": {
```

packages/nx-terraform-plugin/src/executors/publish/schema.json
```diff
@@ -0,0 +1,34 @@
...
+    },
+    "version": {
+      "description": "Module version from module.json",
+      "type": "string"
+    },
+    "githubOwner": {
+      "description": "Optional GitHub owner override from module.json.github.owner",
+      "type": "string"
+    },
+    "workspaceRoot": {
```
krusty93, line 28:
```suggestion
    "workspacePath": {
```

packages/nx-terraform-plugin/module.schema.json
```diff
@@ -0,0 +1,37 @@
+{
+  "$schema": "https://json-schema.org/draft/2020-12/schema",
```
krusty93, line 2: question: `draft`?

docs/superpowers/plans/2026-05-07-nx-terraform-publish.md
krusty93, line 1: thought: Not sure about committing these files as superpowers is not officially adopted and we don't need them anymore

packages/nx-terraform-plugin/src/executors/publish/schema.json
lucacavallaro: `projectRoot` is Nx jargon, I would leave as is because it is unambiguous for Nx users

packages/nx-terraform-plugin/src/executors/publish/schema.json
lucacavallaro: `workspaceRoot` is Nx jargon, I would leave as is because it is unambiguous for Nx users

docs/superpowers/plans/2026-05-07-nx-terraform-publish.md
lucacavallaro: Yes, I left here for review, but I will remove them before merging

packages/nx-terraform-plugin/module.schema.json
lucacavallaro: yup, is the JSON schema version name (`Draft 2020-12` is the latest revision)

image

gunzip (commented):
lgtm but still has merge conflicts

gunzip (approved):

PR #1741, opened 2026-05-13
Include DevEx plugins installation in monorepo scaffolder

christian-calabrese:
Integrate DX Copilot plugin settings into the monorepo scaffold to ensure recommended marketplace and plugin configurations are included. This addition aims to streamline the setup process for developers using GitHub Copilot with PagoPA DX plugins.

Resolves CES-1866


apps/cli/src/adapters/plop/generators/monorepo/__tests__/copilot-settings-template.test.ts
lucacavallaro, line 1: suggestion(non blocking): remove the test, we trust `plop` and not need to test its api

apps/cli/templates/monorepo/.github/copilot/settings.json
lucacavallaro, line 1: nitpick: the use of the `hsb` extension is suggested only when handlebars is used, in this case the file can be named just `settings.json`

christian-calabrese: Done!

apps/cli/src/adapters/plop/generators/monorepo/__tests__/copilot-settings-template.test.ts
christian-calabrese: Done!

.nx/version-plans/version-plan-1778667777940.md
```diff
@@ -0,0 +1,5 @@
+---
+'@pagopa/dx-cli': major
```
kin0992: **question**: why a major?

what would be the migration path then?

christian-calabrese: Whoopsie! Changed to patch, it was a `major` typo

lucacavallaro (approved):

PR #1742, opened 2026-05-13
Update outputs for Container App

krusty93:
Add a new output for the URL of the latest revision deployed Container App and adjust the existing output for the URL. Update assertions to validate the new output.

Close CES-2017

Depends on #1744


krusty93:
Will be merged via #1744

PR #1746, opened 2026-05-13
Add diagram to Container App module

krusty93:
Close CES-2051

Depends on #1747


.nx/version-plans/version-plan-1778685649326.md
```diff
@@ -0,0 +1,5 @@
+---
+azure_container_app: major
```
mamu0, line 2: [question](https://pullpo.io/cc?l=question) 
Are we sure this is a `major` release? I see the versions were bumped, but they stay within the same range. We should check if there are any instances using a lower version; if not, this might actually just be a `patch`.

To clarify, `azurerm` previously had `version = ">= 4.16.0, < 5.0"` and now it's `version = "~> 4.70"`, which falls within the previous interval, so It's possible that recent applies have already updated it to version 4.70 or higher.

krusty93: This PR is stacked with other PRs that brings several breaking changes

PR #1748, opened 2026-05-13
Add custom Trivy rule to prevent Key Vault data source usage

krusty93:
Implement a custom Trivy rule that denies the use of Key Vault secrets via data sources in Terraform configurations. Update the workflow to include this new rule and ensure proper execution of Trivy checks.

Depends on #1780
Close CES-2004

### DIFF
--- a/.github/workflows/static_analysis.yaml
+++ b/.github/workflows/static_analysis.yaml
@@ -126,9 +126,11 @@ jobs:
             --config-check .dx-security/.trivy/checks/terraform \
             --check-namespaces user \
             --severity MEDIUM,HIGH,CRITICAL \
-            --exit-code 0 > /tmp/trivy-output.log 2>&1 || true
+            --format json \
+            --skip-version-check \
+            --exit-code 0 > /tmp/trivy-output.json || true
 
-          cat /tmp/trivy-output.log
+          cat /tmp/trivy-output.json
 
       - name: Run pre-commit on specific folders
         if: ${{ inputs.folder != '' }}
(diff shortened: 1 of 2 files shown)


.trivy/checks/terraform/forbid_azurerm_key_vault_secret_data_source.rego
```diff
@@ -0,0 +1,46 @@
...
+# Scans the raw Terraform AST for fskey field references matching "data.azurerm_key_vault_secret.*"
+# fskey is Trivy's internal field that tracks the full resource path in the configuration tree.
+# This rule catches any direct usage of the data source (e.g., data.azurerm_key_vault_secret.example)
+deny contains res if {
+  some _, obj in walk(input)
+  is_object(obj)
+  fskey := object.get(obj, "fskey", "")
+  startswith(fskey, "data.azurerm_key_vault_secret.")
+
+  msg := "Use of data.azurerm_key_vault_secret is forbidden. Prefer runtime secret resolution or write-only secret patterns."
```
gunzip:
```suggestion
  msg := "Use of data.azurerm_key_vault_secret is forbidden. Use runtime secret resolution or write-only secret patterns."
```

.github/workflows/static_analysis.yaml
```diff
@@ -85,6 +90,16 @@ jobs:
           fetch-depth: 0
           ref: ${{ github.sha }}
 
+      - name: Checkout DX Trivy policies
+        uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
+        with:
+          repository: pagopa/dx
+          ref: ${{ github.repository == 'pagopa/dx' && github.sha || inputs.dx_policies_ref }}
```
krusty93: thought: Does it worth? or we should stick to main?

lucacavallaro: in my opinion we should stick to `main`, we can change it later if needed

.github/workflows/static_analysis.yaml
```diff
@@ -85,6 +90,16 @@ jobs:
...
+          path: .dx-security
```
lucacavallaro, line 92: nitpick: `.pagopa-dx` instead `.dx-security`

lucacavallaro (approved):
LGTM, but I would remove `dx_policies_ref` for now

.github/workflows/static_analysis.yaml
krusty93: I'm going to remove it before merging

PR #1749, opened 2026-05-13
Deploy dx-metrics web image

gunzip:
## Why
The DX Metrics Container App was building the last stage from a multi-stage Dockerfile. Because the final stage is `import-runner`, the deployed revision started `dist/import.mjs` instead of the Next.js server and failed on GitHub authentication.

## What changed
- add an optional `build_target` input to the reusable Container App release workflow
- make the DX Metrics deploy workflow build the `runner` stage explicitly
- trigger the DX Metrics deploy workflow when the relevant workflow files change


gunzip:
closing in favor of #1750

PR #1751, opened 2026-05-14
Static web apps release to production can start when tags are created

christian-calabrese:
Revise the deployment workflow to trigger on tag releases instead of pushes to the main branch. Add a deprecation notice for the legacy push trigger to guide users towards the new tag-based approach.

Resolves CES-2000


.github/workflows/release-azure-staticapp-v1.yaml
```diff
@@ -145,9 +148,9 @@ jobs:
           STATIC_WEB_APP_NAME: ${{ inputs.static_web_app_name }}
           RESOURCE_GROUP_NAME: ${{ inputs.resource_group_name }}
 
-      - name: Deploy Preview
+      - name: Deploy Static Website
```
gunzip, line 167: question: it looks like we changed this _preview_ trigger to add 'release' (tagging). Is this the expected behvior?

.github/workflows/release-azure-staticapp-v1.yaml
```diff
@@ -79,6 +77,11 @@ jobs:
       bundle_name: ${{ env.BUNDLE_NAME }}
 
     steps:
+      - name: Warn Legacy Push Trigger Deprecation
+        if: github.event_name == 'push' && github.ref == 'refs/heads/main'
+        run: |
+          echo "::warning::The push-to-main trigger for this reusable workflow is deprecated and will be removed in the next release. Migrate callers to a tag-based release trigger."
+
```
lucacavallaro: **thought:** In my opinion, we should avoid mandating a specific way to call the workflow. Keeping it independent would make it much more versatile for the users. Why are we currently requiring this specific approach?

gunzip: agree

christian-calabrese: It is necessary to know when the website can be released to the production slot and when to the preview slot

lucacavallaro: Can't the condition be something like: PR event = preview, other = production?

christian-calabrese: @lucacavallaro @gunzip 
Fixed, now the caller has the possibility to choose how to trigger the deployment to production. Any pull-request-related events, instead, will touch only a preview deployment environment.

This behavior has been tested here:
- Trigger on push in main ([run](https://github.com/pagopa/dx-playground/actions/runs/25924185590))
- Trigger on tags created ([run](https://github.com/pagopa/dx-playground/actions/runs/25923591861))

.github/workflows/release-azure-staticapp-v1.yaml
christian-calabrese: It's the same action that does both things. `Deploy Preview` only describes one of the cases it carries out

.github/workflows/release-azure-staticapp-v1.yaml
```diff
@@ -109,8 +120,9 @@ jobs:
     name: Deploy Static Website
     runs-on: ubuntu-latest
     needs: [build]
-    if: ${{ !github.event.act }}
-    environment: ${{ inputs.environment }}-${{ ( github.event_name == 'push' && github.ref == 'refs/heads/main' ) && 'cd' || 'ci' }}
+    # Allow preview cleanup on closed PRs even when the build job was skipped.
+    if: ${{ always() && !github.event.act && ((github.event_name == 'pull_request' && github.event.action == 'closed') || needs.build.result == 'success') }}
+    environment: ${{ inputs.environment }}-${{ github.event_name == 'pull_request' && 'ci' || 'cd' }}
```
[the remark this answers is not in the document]
christian-calabrese: That's intended

.github/workflows/_release-bash-docusaurus-docs-website-azure.yaml
```diff
@@ -2,20 +2,26 @@ name: Deploy DX Website to Azure
...
 on:
   push:
-    branches:
-      - main
-    paths:
-      - "apps/website/**" # Trigger only when this app changes
+    # GitHub Actions does not evaluate path filters for tag pushes.
+    # Scope production deploys with the project-specific release tag prefix.
+    tags:
+      - "docs@*"
```
mamu0, line 8: [question](https://pullpo.io/cc?l=question) 
Have you verified that this tag trigger works as expected?

christian-calabrese: I successfully [tested it in dx-playground](https://github.com/pagopa/dx-playground/actions/runs/25923591861). I know that this trigger is not so reliable, but I think that the GitHub app makes it work better than PATs

lucacavallaro (approved):

PR #1754, opened 2026-05-14
Fix Drift Detection notification

mamu0:
This PR fixes an issue where drift detection was failing while constructing the Slack message body. It now removes HTML metadata since it's unnecessary and already hidden, and uses toJSON to ensure a safe format.

Resolves: CES-2046

### DIFF
--- a/.github/workflows/infra_drift_detection.yml
+++ b/.github/workflows/infra_drift_detection.yml
@@ -199,8 +199,12 @@ jobs:
               run_id: context.runId
             });
             const data = run.data;
-            // Converting commit message to single line
-            const commitMessageSingleLine = data.head_commit.message.replace(/\n/g, ' ');
+            // Strip hidden HTML comments from release metadata before posting to Slack.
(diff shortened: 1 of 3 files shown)


.github/workflows/infra_drift_detection.yml
```diff
@@ -199,8 +199,12 @@ jobs:
               run_id: context.runId
             });
             const data = run.data;
-            // Converting commit message to single line
-            const commitMessageSingleLine = data.head_commit.message.replace(/\n/g, ' ');
+            // Strip hidden HTML comments from release metadata before posting to Slack.
```
gunzip, line 202: question: in which cases do we have html metadata within the commit message?

mamu0: In the 'Version Packages' PRs, the NX tag list is hidden via HTML in the description. This is the root cause of the recent workflow failures.

gunzip (approved):

PR #1744, opened 2026-05-13
Refactor Container App module variables

krusty93:
This pull request introduces several improvements and breaking changes to the Azure Container App Terraform module. The main updates include a refactor of identity and diagnostics management, a change to the way public/internal access is configured, and support for a new "development" use case. Documentation, examples, and tests have been updated to reflect these changes.

Close CES-1974, CES-2001, CES-2017

From PRs stack:
Close CES-2050, CES-2051, CES-2049

Depends on #1745

PR #1760, opened 2026-05-18
Add support for managing Container App secrets without exposing them

krusty93:
Introduce the `use_in_container` option to allow secrets to be defined at the app level without being exposed as environment variables. This change helps to delete secrets from containers before removing them from the underlying cae.

Close CES-2059


.nx/version-plans/version-plan-1779101795382.md
```diff
@@ -0,0 +1,5 @@
+---
+azure_container_app: patch
+---
+
+Add `use_in_container` to keep selected Container App secrets defined without exposing them as container environment variables.
```
gunzip: suggestion: it might not be entirely clear what `use_in_container` means, why, and in which cases the value shouldn't be used in the container. 

maybe a clearer name that highlights the use case could be something like `retain_for_deletion`, `schedule_for_deletion`, `pending_deletion`, `decommissioning`, or `detach_for_deletion`?

infra/modules/azure_container_app/variables.tf
```diff
@@ -137,9 +137,13 @@ variable "secrets" {
   type = list(object({
     name                = string
     key_vault_secret_id = string
+    use_in_container    = optional(bool, true)
```
lucacavallaro: question(blocking): what if I have multiple containers in my container app?

example: my app + a sidecar container

krusty93: it's working as before - secrets are injected in any container

.nx/version-plans/version-plan-1779101795382.md
krusty93: When it is useful is specified in the variable description: https://github.com/pagopa/dx/pull/1760/changes#diff-5442de74993562abb27fe8d4adcacec3d6120771194b709d5501f6af717532f0R145

gunzip: the only useful information in that description is "useful for secrets deletion." I would use a name that immediately refers to that (the user doesn't know what 'use in the container' means)

infra/modules/azure_container_app/variables.tf
lucacavallaro: uhm, I don't like this solution, because in this way we are supporting this boolean only to remove secrets and not to enhance the secret management within the container app

to me the identified solution should address both issues

krusty93: What issue do you want to solve?

lucacavallaro: issue 1: give users the ability to delete secrets
issue 2: allow users to assign secrets to container, without binding all of them automatically

krusty93: 1. it must be done in 3 step (app, infra without secrets binding, infra without secret at all) until we tie infrastructure with configuration
2. requires a change in the interface, it can be done in https://github.com/pagopa/dx/pull/1744

lucacavallaro: ok, if you promise me that it will addressed on #1744 for me it is ok to merge now this minor change

lucacavallaro (approved):

infra/modules/azure_container_app/variables.tf
gunzip: I would still rename the variable. When I read the name, I had several questions:
1. what does 'used (or not used) in the container' mean?
2. where else could a secret be used (if not in the container)?
3. why shouldn't I use a secret that I have set?

If we use a variable name that clarifies the use case (that is, 'when' the user should use it)
would not give raise to further questions.

lucacavallaro: 3. the only purpose right now is to be able to delete it

do you have any suggestion for the alternative name?

krusty93: `scheduled_for_deletion` is ok for me

gunzip: deal!

infra/modules/azure_container_app/variables.tf
```diff
@@ -135,11 +135,15 @@ variable "authentication" {
 
 variable "secrets" {
   type = list(object({
-    name                = string
-    key_vault_secret_id = string
+    name                   = string
+    key_vault_secret_id    = string
+    scheduled_for_deletion = optional(bool, true)
```
gunzip: shouldn't be `false` by default?

infra/modules/azure_container_app/variables.tf
krusty93: done!

infra/modules/azure_container_app/variables.tf
krusty93: Yes I've updated code and description as well

gunzip: tests need fix

PR #1765, opened 2026-05-19
Use dedicated private endpoint type for Azure Managed Redis

kin0992:
## Summary

Updates the `azure_managed_redis` module to use the new `managed_redis_private_endpoint` resource type for naming the Redis private endpoint, producing `amr-pep` instead of the generic `pep` suffix.

This makes the private endpoint name self-descriptive and consistent with how other resources already work (e.g. `cosmos_private_endpoint` → `cosno-pep`).

Before: `dx-d-itn-modules-cache-pep-01`
After:  `dx-d-itn-modules-cache-amr-pep-01`

## Changes

- `infra/modules/azure_managed_redis/locals.tf`: `private_endpoint_name` now uses `resource_type = "managed_redis_private_endpoint"`
- `infra/modules/azure_managed_redis/tests/unit.tftest.hcl`: new assertion in `managed_redis_private_endpoint` run block verifies the `amr-pep` name

## Notes

> ⚠️ The unit tests for this PR will remain red until the companion provider PR is merged and the new provider version is published. After that, the module's lock file should be refreshed via `terraform init -upgrade`.

Depends on #1763

### DIFF
--- a/.nx/version-plans/version-plan-1779187310962.md
+++ b/.nx/version-plans/version-plan-1779187310962.md
@@ -1,0 +1,5 @@
+---
+azure_managed_redis: patch
+---
+
+Use dedicated private endpoint type for managed Redis private endpoint name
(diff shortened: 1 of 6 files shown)


krusty93 (changes requested):
issue: requires a version constraint update against dx provider

kin0992:
> issue: requires a version constraint update against dx provider

yes, working on it! thanks for the feedback

krusty93 (changes requested):
examples constraints must be updated as well

krusty93 (approved):

PR #1769, opened 2026-05-19
Add agent-ready CLI foundations

kin0992:
Add global flags and structured exit codes to support autonomous invocations by coding agents.

This PR introduces:
- Global flags: --non-interactive/-y to disable prompts, --output <text|json> for machine-readable output
- Structured exit codes (0-6) with error classification (E_USAGE, E_MISSING_INPUT, E_PRECONDITION, E_REMOTE, E_VALIDATION)
- JSON error envelopes when --output=json is active
- New modules: exit-codes.ts (constants and CliError class) and output.ts (envelope builders and output adapters)

No user-facing behavior changes yet (flags have no effect until commands opt in). All 251 CLI tests pass.

Resolves CES-1946 (part 1/5)

### DIFF
--- a/apps/cli/src/adapters/commander/output.ts
+++ b/apps/cli/src/adapters/commander/output.ts
@@ -1,0 +1,219 @@
+/**
+ * Output adapter for agent-friendly CLI invocations.
+ *
+ * The DX CLI supports two output modes:
+ *
+ * - `text` (default): human-friendly output via chalk/ora, unchanged from
+ *   today's behaviour.
+ * - `json`: a single JSON envelope is written to stdout at the end of the
+ *   command; optional NDJSON progress events are written to stderr while
+ *   the command runs.
+ *
+ * Coding agents use `json` mode to consume command outputs deterministically
+ * without parsing prose.
+ */
+import { Command } from "commander";
+
+import {
+  ErrorCode,
+  ErrorCodeValue,
+  ExitCode,
+  exitCodeForErrorCode,
+  ExitCodeValue,
+  isCliError,
+} from "./exit-codes.js";
+import { GlobalOptions } from "./index.js";
+
+export type ErrorEnvelope = {
+  command: string;
+  error: {
+    cause?: string;
+    code: ErrorCodeValue;
+    details?: unknown;
+    message: string;
+  };
+  ok: false;
+};
+
+export type OutputMode = "json" | "text";
+
+export type ProgressEvent = {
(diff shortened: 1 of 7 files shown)


apps/cli/src/adapters/commander/output.ts
```diff
@@ -0,0 +1,219 @@
...
+  if (opts.nonInteractive === true) {
+    return true;
+  }
+  if (opts.output === "json" && !process.stdout.isTTY) {
+    return true;
+  }
+  return false;
+};
+
+const flattenCauseChain = (value: unknown): string | undefined => {
```
gunzip, line 89: question: what's the purpose of this method?

apps/cli/src/adapters/commander/output.ts
```diff
@@ -0,0 +1,219 @@
...
+  const messages: string[] = [];
+  const seen = new Set<unknown>();
+  let current: unknown = value;
+  while (current !== undefined && current !== null && !seen.has(current)) {
+    seen.add(current);
+    if (current instanceof Error) {
+      if (current.message.length > 0) {
+        messages.push(current.message);
+      }
+      current = (current as { cause?: unknown }).cause;
```
gunzip, line 99: suggestion: can we avoid this cast?

apps/cli/src/adapters/commander/output.ts
```diff
@@ -0,0 +1,219 @@
...
+export const isNonInteractive = (command: Command): boolean => {
+  const opts = command.optsWithGlobals<GlobalOptions>();
+  if (opts.nonInteractive === true) {
+    return true;
+  }
+  if (opts.output === "json" && !process.stdout.isTTY) {
+    return true;
+  }
+  return false;
+};
```
lucacavallaro, line 87: suggestion: check also the `CI` environment variable, it is the de-facto standard that all tool (pnpm, nx ...) uses to check if they are running in an automation script

in order to be "non interactive" the `CI` variable should be a truthy value

apps/cli/src/adapters/commander/output.ts
kin0992: yes, and it looks like the agent made a mistage even though the instructions are clear.

there is no need to update the instructions then.

kin0992:
Superseded by new architecture: OutputLogger + Zod env pattern. New stacked PRs incoming from feats/cli-output-logger-interface.

PR #1770, opened 2026-05-19
Increase minimum release age in pnpm workspace

gunzip:
Update the minimum release age from 1440 to 4320 minutes to allow for a longer delay before releases.


lucacavallaro (approved):
we should reintroduce the config plugin to enforce this setting to our users as well

PR #1773, opened 2026-05-20
Add CommandPresenter interface and CliEnv schema

kin0992:
Introduces the foundational building blocks for agent-friendly CLI output.

Zod v4 schema that validates `process.env` once at the entrypoint and passes a typed `CliEnv` value down. Keeps env-var reads out of use-cases and adapters and makes them fully testable.

Adds `--output <text|json>` to the root `dx` program (default: `text`).

No behaviour change — port and schema only, wiring follows in subsequent PRs.

Closes CES-1946

### DIFF
--- a/apps/cli/src/adapters/commander/__tests__/env.test.ts
+++ b/apps/cli/src/adapters/commander/__tests__/env.test.ts
@@ -1,0 +1,45 @@
+/**
+ * Tests for the CLI environment schema.
+ * Validates that process.env is parsed correctly before being passed to the CLI.
+ */
+import { describe, expect, it } from "vitest";
+
+import { cliEnvSchema } from "../env.js";
+
+describe("cliEnvSchema", () => {
+  describe("CI field", () => {
+    it("is undefined when CI is not set", () => {
+      const result = cliEnvSchema.safeParse({});
+      expect(result.success).toBe(true);
+      expect(result.success && result.data.CI).toBeUndefined();
+    });
+
+    it("captures any non-empty CI value", () => {
+      const result = cliEnvSchema.safeParse({ CI: "true" });
+      expect(result.success).toBe(true);
+      expect(result.success && result.data.CI).toBe("true");
+    });
+
+    it("captures CI=false as a string (presence is the signal, not the value)", () => {
+      const result = cliEnvSchema.safeParse({ CI: "false" });
+      expect(result.success).toBe(true);
+      expect(result.success && result.data.CI).toBe("false");
+    });
+
+    it("captures CI=1 for numeric-style env vars", () => {
+      const result = cliEnvSchema.safeParse({ CI: "1" });
+      expect(result.success).toBe(true);
+      expect(result.success && result.data.CI).toBe("1");
+    });
+  });
+
+  it("passes through an object with unrelated env keys without failing", () => {
+    const result = cliEnvSchema.safeParse({
+      CI: "true",
+      HOME: "/home/user",
+      NODE_ENV: "test",
+    });
+    expect(result.success).toBe(true);
+    expect(result.success && result.data.CI).toBe("true");
+  });
+});
--- a/apps/cli/src/adapters/commander/__tests__/env.test.ts
+++ b/apps/cli/src/adapters/commander/__tests__/env.test.ts
@@ -1,44 +1,49 @@
 /**
  * Tests for the CLI environment schema.
- * Validates that process.env is parsed correctly before being passed to the CLI.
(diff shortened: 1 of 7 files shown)


apps/cli/src/adapters/commander/index.ts
```diff
@@ -16,6 +16,7 @@ import { formatErrorDetailed, toErrorMessage } from "./error-reporting.js";
 export type CliDependencies = CodemodCommandDependencies;
 
 export type GlobalOptions = {
+  output?: "json" | "text";
```
kin0992: **question**: shouldn't be text the default?

apps/cli/src/adapters/commander/env.ts
```diff
@@ -0,0 +1,20 @@
...
+import { z } from "zod";
+
+export const cliEnvSchema = z
+  .object({
+    // Standard CI marker. Presence (any string value) signals an automated
+    // pipeline where interactive prompts must not block. Follows the same
+    // convention used by `is-interactive` and `ora`.
+    CI: z.string().optional(),
+  })
+  .passthrough();
```
kin0992: **note**: `passthrough` is deprecated

apps/cli/src/adapters/commander/__tests__/env.test.ts
```diff
@@ -0,0 +1,45 @@
...
+ * Validates that process.env is parsed correctly before being passed to the CLI.
+ */
+import { describe, expect, it } from "vitest";
+
+import { cliEnvSchema } from "../env.js";
+
+describe("cliEnvSchema", () => {
+  describe("CI field", () => {
+    it("is undefined when CI is not set", () => {
+      const result = cliEnvSchema.safeParse({});
```
kin0992: **note**: pass `process.env` but stub it first with https://vitest.dev/api/vi.html#vi-stubenv

apps/cli/src/adapters/commander/env.ts
```diff
@@ -0,0 +1,20 @@
...
+import { z } from "zod";
+
+export const cliEnvSchema = z
+  .object({
+    // Standard CI marker. Presence (any string value) signals an automated
+    // pipeline where interactive prompts must not block. Follows the same
+    // convention used by `is-interactive` and `ora`.
+    CI: z.string().optional(),
+  })
+  .loose();
```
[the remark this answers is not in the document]
kin0992: `.strip` has been deprecated

apps/cli/src/domain/output-logger.ts
```diff
@@ -0,0 +1,35 @@
...
+   * a structured `{"ok":true,...}` envelope on stdout.
+   */
+  reportResult<T>(data: T): void;
+
+  /**
+   * Runs `task` while emitting start/end lifecycle events for the named step.
+   * In text mode this shows an interactive progress indicator; in JSON mode it
+   * emits NDJSON step events on stderr.
+   */
+  runStep<T>(name: string, task: () => Promise<T>): Promise<T>;
```
gunzip: naming: i wouldn't expect a logger interface to have a method called "run." The other methods are appropriately named "report." Maybe we could consider calling it something like "track" or "trackProgress"?

kin0992: yes, it totally agree. i am chatting about it too.

kin0992: @gunzip what do you think to rename the interface to `CommandPresenter` and the `runStep` function to `trackStep`?

gunzip: better!

gunzip (approved):

apps/cli/src/adapters/commander/env.ts
```diff
@@ -0,0 +1,20 @@
...
+ * adapters and makes them fully testable.
+ */
+import { z } from "zod";
+
+export const cliEnvSchema = z
+  .object({
+    // Standard CI marker. Presence (any string value) signals an automated
+    // pipeline where interactive prompts must not block. Follows the same
+    // convention used by `is-interactive` and `ora`.
+    CI: z.string().optional(),
```
lucacavallaro, line 16: suggestion: the content of `CI` it is always a string (because environment variables are strings), but to block the interactive prompt its content should be truthy

`CI=0` or `CI=false` should not disable the interactive mode

gunzip: we can use truthy values for this https://zod.dev/api?id=stringbool#stringbool

kin0992: I am going to address these comments in PR #1776

PR #1774, opened 2026-05-20
Add integration tests for DX Cli Plop generator

gunzip:
Introduce testing utilities and integration tests for the Plop generator, enhancing the testing framework for environment and monorepo generators. This update allows for better verification of file generation logic and artifact management during tests.

Closes CES-2045


apps/cli/README.md
```diff
@@ -65,6 +65,16 @@ npm run build
 node ./apps/cli/bin/index.js --help
 ```
 
+## 🧪 Testing
+
+To inspect the files generated by the Plop generator integration tests, set `DX_TEST_KEEP_ARTIFACTS=1` before running the test command.
```
gunzip: @pagopa/engineering-team-devex this mechanism allows for a dry run by leveraging vitest mocks, without requiring any changes to the CLI codebase

mamu0:
[question](https://pullpo.io/cc?l=question) 
@gunzip It's not entirely clear to me how to use this feature during development. Could you provide an example of how we would use it and what the expected output should be?

gunzip:
@mamu0 README.md contains needed instructions

ref. https://github.com/pagopa/dx/pull/1774/changes#diff-814f330f4d023072c7412643f20fea1745bfd132e3471ece07acc009a96c0c77R70

image

image

apps/cli/README.md
krusty93: Simplifying to `DRY_RUN=1` or `KEEP_ARTIFACTS=1`?

gunzip: done!

apps/cli/src/adapters/plop/generators/__tests__/temp-dir.ts
```diff
@@ -0,0 +1,19 @@
+/**
+ * Temp directory helpers for Plop generator integration tests.
+ */
+import * as fs from "node:fs/promises";
```
lucacavallaro: **nitpick:** `node:*` packages have a default export, so this import can be changed to `import fs from ..` (the effect is almost the same, but it depends on the package)

apps/cli/src/adapters/plop/generators/monorepo/__tests__/generation.test.ts
```diff
@@ -0,0 +1,173 @@
...
+  it("appends terraform lock file exclusions to .gitignore", async () => {
+    const content = await fs.readFile(
+      path.join(tmpDir, payload.repoName, ".gitignore"),
+      "utf-8",
+    );
+    expect(content).toContain("**/modules/**/.terraform.lock.hcl");
+    expect(content).toContain("**/_modules/**/.terraform.lock.hcl");
+  });
+
+  it("generates static config files (nx.json, pnpm-workspace.yaml)", async () => {
```
lucacavallaro: **thought:** it seems that we are testing `plop` instead of our own logic

in my opinion we should test only our own logics, given a specific _generator state_

apps/cli/README.md
```diff
@@ -65,6 +65,16 @@ npm run build
 node ./apps/cli/bin/index.js --help
 ```
 
+## 🧪 Testing
```
lucacavallaro: **thought:** this feature incentivizes manual testing over automatic one, which is a bad practice (not scalable, not reproducible)

developers should be guided to write new test cases

apps/cli/src/adapters/plop/generators/__tests__/temp-dir.ts
```diff
@@ -0,0 +1,19 @@
...
+
+export const shouldKeepTestArtifacts = (env: NodeJS.ProcessEnv): boolean =>
+  env.KEEP_ARTIFACTS === "1" || env.KEEP_ARTIFACTS === "true";
+
+export const cleanupTempDir = async (
```
kin0992: **question**: would it be possible to use a mock fs instead? like `memfs`

this would simplify a lot (we do not need to handle the temporary directory.
the ratio is: are we sure we need to manually investigate what has been generated by plop? or is it enough to add some assertions that help us to verify the generated code?

apps/cli/src/adapters/plop/generators/monorepo/__tests__/generation.test.ts
gunzip: i reviewed the tests by removing the specific expectations related to plop

apps/cli/README.md
gunzip: @lucacavallaro @kin0992 the rationale behind this approach  - which lets devs keep artifacts - isn’t to test correct functionality, but to allow devs to observe the side effects in file generation when introducing a new feature or performing a refactor. if files are necessarily generated only after other side effects (e.g., calls to azure or github), those developing have to proceed without being able to observe these effects.

apps/cli/src/adapters/plop/generators/__tests__/temp-dir.ts
gunzip: see https://github.com/pagopa/dx/pull/1774#discussion_r3318379882

apps/cli/README.md
lucacavallaro: If these side effects are predictable enough for a developer to "observe" and validate, why can't that validation be encoded into the test suite?

kin0992: @gunzip I see the point. I also agree with @lucacavallaro, since to me it is really the same thing to peek at a file or have a test that asserts the very same stuff I would personally look at. 
And if I have to choose a path to follow, I would go with doing it once (write the test once) and forget it. I think we will spend more time once, when the thes has to be written, but gain confidence for further test suite executions.

gunzip: I'm not taking sides here. Just want to point out that, according to this principle, we would be encouraging developers **not to try out changes locally** (for example, through complex `docker-compose.yaml` setups), but instead to write tests directly without the chance to manually observe the side effects. is this really the way forward?

lucacavallaro: > [!WARNING]
> Maybe I didn't understand the difference between test and observe. I'm assuming they are the same. But please correct me If I am wrong.

We are still encouraging developers to test changes locally, but rather than relying on "manual testing," we suggest encoding those same checks into a script.

Consider this scenario: *A developer introduces Azure-specific logic into a template file and wants to ensure it is implemented correctly.*

**Your Proposal**

1. The developer runs the test with `KEEP_ARTIFACT=true`.
2. The developer then manually opens their editor to inspect the content of the generated file.

> The test is **not reproducible** and **not recorded**. Consequently, reviewers are forced to repeat the entire manual process themselves.

**Our Proposal**

1. The developer adds a new test case to our suite. This test confirms the file’s existence, opens it, and makes assertions based on the content and the generation state.
2. The developer runs the test.

> The test is now a **permanent part of our suite**. It’s automated, verifiable, and saves everyone time.

gunzip: The ability to keep artifacts is valuable during development, as it allows you to see the side effects firsthand. Tests are added after development is complete, to verify correctness. I understand that, in a TDD approach, the opposite should happen, but that would represent a paradigm shift compared to what has historically occurred. Are we sure it’s appropriate to deprive developers of the opportunity to observe the results of what they’ve implemented?

kin0992: I think we are going out of scope here. My point was just related to this activity, which started from the `--dry-run` flag and ended up in a proposal with integration tests, as we discussed and agreed.

The only developers who could suffer this are the DevEx Team developers. Since you talked about TDD, then if we think that is really valuable, we can easily write (or use an existing) skill to do that and improve our Agent's instructions. I don't see it as a high mountain to climb, honestly.
To me, the effort to run a command (the integration tests), go to find a file and check if file X has some properties is greater than writing (or asking the agent to write) a test making sure that file X has some properties. Actually, I think the first scenario has even more effort, since I think that those kinds of checks should be tracked in a test.

But I think we should see what @pagopa/engineering-team-devex thinks about it, since we are all involved in this development process.

**TL: DR**: I would go with mocking the File System (do not persist files) and improving tests to code the check a developer would manually do (by reading the generated file). If, for any reason, most of the team thinks the concrete files are still useful, then I would keep them **BUT** I would like to have those manual checks coded in tests too.

lucacavallaro: I find TDD the best approach since it allow you to make sure to cover all the relevant parts by the unit tests. And you should consider that Vitest also supports snapshot testing that can be used in a similar way to keep artifact (use `toMatchSnapshot()` and regenerate the snapshot until it is ok).

mamu0: Both sides make a fair point here.
Developers do need a way to observe generated output before they can write meaningful assertions, but encoding that in a documented manual step is not great either.

I think @lucacavallaro 's hint at `toMatchSnapshot()` could be the right middle ground: first run generates the snapshot (developer sees the output), subsequent runs are fully automated. Worth exploring?

apps/cli/src/adapters/plop/generators/__tests__/temp-dir.ts
gunzip: The flag has been removed. Regarding the use of a virtual file system, I explored several solutions with `memfs` and `platformatic/vfs`: unfortunately, each of these ended up introducing more complexity than they actually solved.

apps/cli/README.md
gunzip: Done https://github.com/pagopa/dx/pull/1774/changes/4b1bb121410792cfe9adbddbcbb97a23b77fe3f4

kin0992 (approved):

PR #1776, opened 2026-05-20
Add TextCommandPresenter and JsonCommandPresenter adapters

kin0992:
Implements the two concrete adapters for the `CommandPresenter` domain port introduced in #1773, plus the `createCommandPresenter` factory and `isNonInteractive` helper.

### `TextCommandPresenter` (`text-command-presenter.ts`)

Human-readable adapter for interactive terminal sessions:
- `trackStep` — wraps `oraPromise` for spinner feedback
- `reportResult` — writes to stdout via `console.log`
- `reportError` — writes to stderr with chalk red colouring

### `JsonCommandPresenter` (`json-command-presenter.ts`)

Structured adapter for automated pipelines and coding agents:
- `trackStep` — emits NDJSON `{type,status,name}` events to stderr
- `reportResult` — emits `{ok:true,data}` envelope to stdout
- `reportError` — emits `{ok:false,error}` envelope to stdout

No behaviour change — adapters are not yet wired into any command.

Depends on #1773

Resolves part of CES-2112


apps/cli/src/adapters/commander/presenters/__tests__/json.test.ts
```diff
@@ -0,0 +1,130 @@
...
+
+const captureStdout = () => {
+  const written: string[] = [];
+  const spy = vi
+    .spyOn(process.stdout, "write")
+    .mockImplementation((data: unknown) => {
+      written.push(String(data));
+      return true;
+    });
+  return { restore: () => spy.mockRestore(), written };
```
gunzip: **nitpick**: mock cleanup is not guaranteed in tests. Both `json.test.ts` and `text.test.ts` used manual `.restore()` calls at the end of each test. If any assertion threw before the restore line, the spy would leak and cause "already spied on" errors in every subsequent test, masking the original failure.

You may consider replace manual `spy.restore()` calls with a single `afterEach(() => { vi.restoreAllMocks(); })` in the top-level describe block.

gunzip (approved):

PR #1781, opened 2026-05-21
Use GitHub app to authenticate NX release workflow

christian-calabrese:
## Summary

- switch the Nx release flow from repository/PAT-style GitHub authentication to a GitHub App token
- make checkout stop persisting credentials so subsequent git operations actually use the GitHub App token
- resolve the GitHub App bot identity dynamically and use it as the git author for release commits
- document the new action inputs and the GitHub App permission requirements
- improve push failure diagnostics when the installation can authenticate but cannot push

## Why

The release flow needs one consistent identity for branch pushes, tag pushes, pull request updates, and GitHub Releases. Using the GitHub App token end to end avoids mixing credentials from `actions/checkout` and makes permission failures easier to diagnose.

## Notes

- `release-v2.yaml` intentionally points to this branch version of `actions/nx-release` so the workflow can exercise the changes introduced by this PR.
- the PR body previously described the old docs-tag pattern experiment and is now aligned with the actual diff.

### DIFF
--- a/.github/workflows/_release-bash-docusaurus-docs-website-azure.yaml
+++ b/.github/workflows/_release-bash-docusaurus-docs-website-azure.yaml
@@ -5,7 +5,7 @@ on:
     # GitHub Actions does not evaluate path filters for tag pushes.
     # Scope production deploys with the project-specific release tag prefix.
     tags:
-      - "docs@*"
+      - "docs@[0-9]+.[0-9]+.[0-9]+"
   pull_request:
     types: [opened, synchronize, reopened, closed]
     branches:
--- a/.github/workflows/_release-bash-docusaurus-docs-website-azure.yaml
+++ b/.github/workflows/_release-bash-docusaurus-docs-website-azure.yaml
@@ -5,7 +5,7 @@ on:
     # GitHub Actions does not evaluate path filters for tag pushes.
     # Scope production deploys with the project-specific release tag prefix.
     tags:
-      - "docs@[0-9]+.[0-9]+.[0-9]+"
+      - "docs@*"
   pull_request:
     types: [opened, synchronize, reopened, closed]
     branches:
--- a/.github/workflows/release-v2.yaml
+++ b/.github/workflows/release-v2.yaml
@@ -77,7 +77,7 @@ jobs:
         run: npm ci
 
       - name: Nx Publish
-        uses: ./.dx-workflow-source/actions/nx-release
+        uses: pagopa/dx/actions/nx-release@use-regex-for-tags-release-website-to-azure
         with:
           app-slug: ${{ steps.generate-token.outputs.app-slug }}
           github-token: ${{ steps.generate-token.outputs.token }}
--- a/.github/workflows/release-v2.yaml
+++ b/.github/workflows/release-v2.yaml
@@ -26,12 +26,21 @@ jobs:
       id-token: write
       pull-requests: write
     steps:
+      - name: Generate GitHub App token
+        id: generate-token
+        uses: actions/create-github-app-token@1b10c78c7865c340bc4f6099eb2f838309f1e8c3 # 3.1.1
+        with:
+          client-id: ${{ secrets.app-client-id }}
+          private-key: ${{ secrets.app-private-key }}
+          permission-contents: write
+          permission-pull-requests: write
+
       - name: Checkout repository
         uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
         with:
           fetch-tags: "true"
           fetch-depth: 0
-          persist-credentials: false
+          token: ${{ steps.generate-token.outputs.token }}
 
       - name: Detect Release tool
         id: nx-detect
@@ -68,7 +77,7 @@ jobs:
         run: npm ci
 
       - name: Nx Publish
-        uses: pagopa/dx/actions/nx-release@use-regex-for-tags-release-website-to-azure
+        uses: ./actions/nx-release
         with:
-          app-client-id: ${{ secrets.app-client-id }}
-          app-private-key: ${{ secrets.app-private-key }}
+          app-slug: ${{ steps.generate-token.outputs.app-slug }}
+          github-token: ${{ steps.generate-token.outputs.token }}
--- a/.github/workflows/release-v2.yaml
+++ b/.github/workflows/release-v2.yaml
@@ -68,7 +68,7 @@ jobs:
         run: npm ci
 
       - name: Nx Publish
-        uses: pagopa/dx/actions/nx-release@main
+        uses: pagopa/dx/actions/nx-release@use-regex-for-tags-release-website-to-azure
         with:
           app-client-id: ${{ secrets.app-client-id }}
           app-private-key: ${{ secrets.app-private-key }}
--- a/.github/workflows/release-v2.yaml
+++ b/.github/workflows/release-v2.yaml
@@ -77,7 +77,7 @@ jobs:
         run: npm ci
 
       - name: Nx Publish
-        uses: pagopa/dx/actions/nx-release@use-regex-for-tags-release-website-to-azure
+        uses: pagopa/dx/actions/nx-release@main
         with:
           app-slug: ${{ steps.generate-token.outputs.app-slug }}
           github-token: ${{ steps.generate-token.outputs.token }}
--- a/.github/workflows/release-v2.yaml
+++ b/.github/workflows/release-v2.yaml
@@ -77,7 +77,7 @@ jobs:
         run: npm ci
 
       - name: Nx Publish
-        uses: pagopa/dx/actions/nx-release@main
+        uses: pagopa/dx/actions/nx-release@use-regex-for-tags-release-website-to-azure
         with:
           app-slug: ${{ steps.generate-token.outputs.app-slug }}
           github-token: ${{ steps.generate-token.outputs.token }}
--- a/.github/workflows/release-v2.yaml
+++ b/.github/workflows/release-v2.yaml
@@ -31,6 +31,7 @@ jobs:
         with:
           fetch-tags: "true"
           fetch-depth: 0
+          persist-credentials: false
 
       - name: Detect Release tool
         id: nx-detect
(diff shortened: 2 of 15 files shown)


.github/workflows/_release-bash-docusaurus-docs-website-azure.yaml
```diff
@@ -5,7 +5,7 @@ on:
     # GitHub Actions does not evaluate path filters for tag pushes.
     # Scope production deploys with the project-specific release tag prefix.
     tags:
-      - "docs@*"
+      - "docs@[0-9]+.[0-9]+.[0-9]+"
```
kin0992: **question**: why do we need to add a stricter constraint on the tag format?

actions/nx-release/action.yaml
```diff
@@ -66,12 +66,40 @@ runs:
...
       if: steps.release-mode.outputs.mode == 'create-pr' || steps.release-mode.outputs.mode == 'publish' || steps.release-mode.outputs.mode == 'publish-all'
       shell: bash
+      env:
+        BOT_LOGIN: ${{ steps.app-bot.outputs.login }}
+        BOT_EMAIL: ${{ steps.app-bot.outputs.email }}
       run: |
-        git config user.name "dx-pagopa-bot"
-        git config user.email "dx-pagopa-bot@users.noreply.github.com"
+        git config user.name "$BOT_LOGIN"
+        git config user.email "$BOT_EMAIL"
```
mamu0, line 93: [nitpick(non-blocking)](https://pullpo.io/cc?l=nitpick&d=non-blocking) 
This value should be tested because the `@users.noreply.github.com` was originally added to prevent GitHub from seeing two authors for the PR. I'm not sure if the same principle applies when using the GitHub App."

actions/nx-release/action.yaml
```diff
@@ -66,12 +66,40 @@ runs:
           echo "mode=create-pr" >> "$GITHUB_OUTPUT"
         fi
 
+    - name: Configure GitHub App authentication for git operations
+      if: steps.release-mode.outputs.mode == 'create-pr' || steps.release-mode.outputs.mode == 'publish' || steps.release-mode.outputs.mode == 'publish-all'
```
mamu0: [question(if-minor)](https://pullpo.io/cc?l=question&d=if-minor) 
I notice you included everything except for the `warn-pr` case, is there a specific reason for that? Otherwise, you could just check for `if: steps.release-mode.outputs.mode != 'skip'` instead.

actions/nx-release/action.yaml
```diff
@@ -66,12 +66,40 @@ runs:
...
+      shell: bash
+      env:
+        GH_APP_TOKEN: ${{ steps.generate-token.outputs.token }}
+      run: |
+        git config --local --unset-all http.https://github.com/.extraheader || true
+        git remote set-url origin "https://x-access-token:${GH_APP_TOKEN}@github.com/${GITHUB_REPOSITORY}.git"
+
+    - name: Resolve GitHub App bot identity
+      if: steps.release-mode.outputs.mode == 'create-pr' || steps.release-mode.outputs.mode == 'publish' || steps.release-mode.outputs.mode == 'publish-all'
```
mamu0, line 65: Image

.github/workflows/_release-bash-docusaurus-docs-website-azure.yaml
christian-calabrese: This was part of an earlier iteration of the PR. The stricter docs-tag constraint is no longer the scope of this change: the PR now focuses on using a GitHub App for the Nx release flow. I updated the PR body to match the actual diff and avoid that confusion.

actions/nx-release/action.yaml
christian-calabrese: Good point. I kept the standard `@users.noreply.github.com` format because that is the documented bot identity format for GitHub Apps, and the goal here is still to keep a single GitHub-recognized author for the release commits. I also added a small code comment to make that choice explicit.

actions/nx-release/action.yaml
christian-calabrese: This is intentional. The action does perform git mutations, but not in the `warn-pr` path. In `warn-pr` we only call the GitHub API to upsert the warning comment on the PR; the branch push happens only in the separate `create-pr` path. I kept the condition narrow so we only rewire git authentication and configure the bot identity for the modes that actually mutate branches, tags, or releases.

actions/nx-release/action.yaml
christian-calabrese: Image

.github/workflows/release-v2.yaml
kin0992, line 1: **question**: have you tried passing the `token` property to the checkout action?

I am going to explain this: on a side project where I use changeset to publish (where the workflow is triggered by the tag creation), I had the very same issue: workflow didn't run even though I used the GitHub app.

After some digging, I found out that if you pass a custom token to the checkout action, the git configuration is inherited from the toke, so all the git commands are executed on behalf of that user. This means that if you pass the GitHub app token, every operation (push included - and it is important the push is performed by the GitHub app) is done by the user that owns the token.
Doing so I did not have to place any custom boilerplate, but I just have a workflow like this (pseudocode):

```
- name: Generate GitHub App token
        id: app-token
        uses: actions/create-github-app-token@v3
        with:
          client-id: ${{ secrets.app_id }}
          private-key: ${{ secrets.app_private_key }}
 
- name: Checkout
        uses: actions/checkout@v6
        with:
          fetch-depth: 0
          token: ${{ steps.app-token.outputs.token || github.token }}
      .............
- name: Create release PR or publish
        uses: changesets/action@v1
        with:
          publish: pnpm release
          version: pnpm version-packages
          commit: 'Release new artifacts'
          title: 'Release new artifacts'
        env:
          GITHUB_TOKEN: ${{ steps.app-token.outputs.token }}
```
and this resolved my issues that a tag did not trigger the workflow

christian-calabrese:
The changes have been tested [here](https://github.com/pagopa/dx-automation/actions/runs/26283265172/job/77364466162) in a test repository

.github/workflows/release-v2.yaml
christian-calabrese: Done, thank you for the hint!

christian-calabrese:
Warn PR action is failing cause the interface has changed in this PR

actions/nx-release/scripts/shared.ts
```diff
@@ -141,13 +141,15 @@ export async function isPublicProject(projectName: string): Promise<boolean> {
...
-    (tag) =>
-      tag === "public" || (typeof tag === "string" && tag.endsWith(":public")),
-  );
+  const hasPublicTag =
+    Array.isArray(tags) &&
+    tags.some(
+      (tag) =>
+        tag === "public" ||
+        (typeof tag === "string" && tag.endsWith(":public")),
+    );
```
gunzip: suggestion: this kind of runtime checks sould be done using a Zod schema. can you ask the agent why it didn't despite instructions state that? I suspect it does some sort of rationalization we should counter within instructions.

actions/nx-release/action.yaml
gunzip: can you add a comment explain the expected result (ux) clarifying this control ?

actions/nx-release/action.yaml
```diff
@@ -66,19 +59,40 @@ runs:
...
+        set -euo pipefail
+
+        bot_login="${APP_SLUG}[bot]"
+        bot_user_id="$(gh api "/users/${bot_login}" --jq .id)"
+
+        echo "login=${bot_login}" >> "$GITHUB_OUTPUT"
+        echo "email=${bot_user_id}+${bot_login}@users.noreply.github.com" >> "$GITHUB_OUTPUT"
+
     - name: Configure Git user
       if: steps.release-mode.outputs.mode == 'create-pr' || steps.release-mode.outputs.mode == 'publish' || steps.release-mode.outputs.mode == 'publish-all'
```
gunzip, line 86: ditto

actions/nx-release/scripts/shared.ts
christian-calabrese: Done, however this is the rationale from copilot:

> In shared.ts, I took a shortcut. I saw a small amount of local narrowing around a couple of fields and treated that code as “best-effort metadata handling” instead of what it actually is: external input coming from CLI JSON output. In this repository, that is the wrong bar. Parsed JSON and external command output should be treated as untrusted input and validated with Zod at the boundary.

actions/nx-release/action.yaml
christian-calabrese: Done!

actions/nx-release/action.yaml
christian-calabrese: Done!

actions/nx-release/action.yaml
```diff
@@ -123,7 +142,16 @@ runs:
         # Commit and push the version changes (bump + changelog)
         git add -A
         git commit -m "$COMMIT_MESSAGE"
-        git push --force-with-lease origin "$RELEASE_BRANCH"
+        if ! git push --force-with-lease origin "$RELEASE_BRANCH" 2>/tmp/nx-release-push.err; then
```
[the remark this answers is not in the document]
christian-calabrese: Not needed

actions/nx-release/action.yaml
```diff
@@ -66,19 +59,45 @@ runs:
...
+      env:
+        BOT_LOGIN: ${{ steps.app-bot.outputs.login }}
+        BOT_EMAIL: ${{ steps.app-bot.outputs.email }}
       run: |
-        git config user.name "dx-pagopa-bot"
-        git config user.email "dx-pagopa-bot@users.noreply.github.com"
+        git config user.name "$BOT_LOGIN"
+        git config user.email "$BOT_EMAIL"
 
     - name: Warn missing version plan coverage on pull request
```
mamu0, line 95: [issue](https://pullpo.io/cc?l=issue) 
Maybe something is going wrong with the GitHub App configuration regarding the warning message when the Version Plan is missing, check the [code review error](https://github.com/pagopa/dx/actions/runs/26293354920/job/77399245041?pr=1781#step:7:27) in this PR

mamu0 (approved):

PR #1783, opened 2026-05-22
Add dx spec command for CLI schema introspection

kin0992:
## Problem

Agents interacting with the `dx` CLI need to discover all available commands, subcommands, flags, and arguments. Today this requires running `--help` on every command individually, which is slow and error-prone for automated tooling.

## Solution

Add a new `dx spec` command that outputs the complete CLI schema as structured JSON in a single call. The output is stable and versioned (`specVersion: "1"`) so agent consumers can detect breaking shape changes.

### Output shape

```json
{
  "specVersion": "1",
  "name": "dx",
  "version": "x.y.z",
  "globalOptions": [ ... ],
  "commands": [
    {
      "name": "codemod",
      "options": [],
      "arguments": [],
      "commands": [
        { "name": "apply", "arguments": [{ "name": "id", "required": true, ... }], ... }
      ]
    }
  ]
}
```

### Architecture

- `domain/spec.ts` — pure `CliSpec` / `CommandSpec` / `OptionSpec` / `ArgumentSpec` types (no framework dependency)
- `adapters/commander/spec.ts` — `extractCliSpec()` recursively maps the Commander tree to domain types, including all registered commands
- `adapters/commander/commands/spec.ts` — `makeSpecCommand()` registered last in `makeCli` so the closure captures the full command tree
- `index.ts` — GitHub PAT assertion is skipped when `spec` is detected in `argv` so the command works in credential-free environments

Closes CES-2100

### DIFF
--- a/apps/cli/src/adapters/commander/__tests__/spec.test.ts
+++ b/apps/cli/src/adapters/commander/__tests__/spec.test.ts
@@ -24,38 +24,40 @@ const buildRoot = () =>
         .default("text"),
     );
 
-describe("extractCliSpec", () => {
-  it("maps the root command name, description and version", () => {
-    const root = buildRoot();
-    const spec = extractCliSpec(root, "1.2.3");
+const registerRootMetadataAndGlobalOptionTests = () => {
+  describe("root metadata and global options", () => {
+    it("maps the root command name, description and version", () => {
+      const root = buildRoot();
+      const spec = extractCliSpec(root, "1.2.3");
 
-    expect(spec).toMatchObject({
-      description: "The CLI for DX-Platform",
-      name: "dx",
-      specVersion: "1",
-      version: "1.2.3",
-    });
-  });
-
-  it("maps global options from the root command", () => {
-    const root = buildRoot();
-    const spec = extractCliSpec(root, "1.2.3");
-
-    expect(
-      spec.globalOptions.find((o) => o.long === "--verbose"),
-    ).toStrictEqual({
-      choices: [],
-      defaultValue: false,
-      description: "Enable verbose output",
-      flags: "-v, --verbose",
-      long: "--verbose",
-      optional: false,
-      required: false,
-      short: "-v",
+      expect(spec).toMatchObject({
+        description: "The CLI for DX-Platform",
+        name: "dx",
+        specVersion: "1",
+        version: "1.2.3",
+      });
     });
 
-    expect(spec.globalOptions.find((o) => o.long === "--output")).toStrictEqual(
-      {
+    it("maps global options from the root command", () => {
+      const root = buildRoot();
+      const spec = extractCliSpec(root, "1.2.3");
+
+      expect(
+        spec.globalOptions.find((o) => o.long === "--verbose"),
+      ).toStrictEqual({
+        choices: [],
+        defaultValue: false,
+        description: "Enable verbose output",
+        flags: "-v, --verbose",
+        long: "--verbose",
+        optional: false,
+        required: false,
+        short: "-v",
+      });
+
+      expect(
+        spec.globalOptions.find((o) => o.long === "--output"),
+      ).toStrictEqual({
         choices: ["text", "json"],
         defaultValue: "text",
         description: "Output mode",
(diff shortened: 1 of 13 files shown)


apps/cli/src/adapters/commander/commands/__tests__/spec.test.ts
```diff
@@ -0,0 +1,107 @@
...
+  description: "The CLI for DX-Platform",
+  globalOptions: [],
+  name: "dx",
+  specVersion: "1",
+  version: "0.0.0",
+});
+
+describe("makeSpecCommand", () => {
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  let stdoutSpy: ReturnType<typeof vi.spyOn<any, any>>;
```
kin0992: issue: avoid `any`

apps/cli/src/adapters/commander/__tests__/spec.test.ts
```diff
@@ -0,0 +1,178 @@
...
+
+    const verbose = spec.globalOptions.find((o) => o.long === "--verbose");
+    expect(verbose).toBeDefined();
+    expect(verbose?.flags).toBe("-v, --verbose");
+    expect(verbose?.short).toBe("-v");
+    expect(verbose?.description).toBe("Enable verbose output");
+    expect(verbose?.defaultValue).toBe(false);
+    expect(verbose?.required).toBe(false);
+    expect(verbose?.optional).toBe(false);
+    expect(verbose?.choices).toEqual([]);
```
kin0992: **nitpick**: use the `.toEqual` or `toStrictEqual` so it is clear the output shape

lucacavallaro (commented):
**thought:** I am wondering why `help` is not enough for this use case. Do you have any supportive data or recommended best practices that support this solution?

Maybe you can use [evals](https://developers.openai.com/api/docs/guides/agent-evals) to rank the agent ability to find DX commands with `help` and `spec`

gunzip:
sounds interesting, but it looks like this needs follow up

apps/cli/src/adapters/commander/index.ts
```diff
@@ -55,10 +57,13 @@ export const makeCli = (
...
   program.addCommand(makeCodemodCommand(cliDeps));
-  program.addCommand(makeInitCommand(deps));
+  program.addCommand(makeInitCommand(deps.requireGitHubAuth));
   program.addCommand(makeSavemoneyCommand());
   program.addCommand(makeInfoCommand(deps));
-  program.addCommand(makeAddCommand(deps));
+  program.addCommand(makeAddCommand(deps.requireGitHubAuth));
+
+  // spec is registered last so the closure captures the complete command tree.
+  program.addCommand(makeSpecCommand(() => extractCliSpec(program, version)));
```
gunzip, line 66: thoughts: the spec command is registered on program before `extractCliSpec(program, version)` runs. Since `program.commands` then includes spec, the JSON output is self-referential. Agents consuming the spec will see a spec command entry they don't need.

Possibile fix (?): either filter spec from the output in `extractCliSpec`, or snapshot the spec before registering the spec command on program.

apps/cli/src/adapters/commander/commands/init.ts
```diff
@@ -280,40 +281,39 @@ export const confirmGitHubRepoCreation = (
...
-          process.chdir(payload.repoName);
-        })
-        .andThen((payload) =>
-          confirmGitHubRepoCreation(payload).andThen<SummaryInput, Error>(
-            (confirmed) =>
-              confirmed
-                ? handleNewGitHubRepository(gitHubService)(payload)
-                : okAsync({ gitHubRepoCreationSkipped: true, payload }),
-          ),
+      await requireGitHubAuth()
```
gunzip: **thoughts**: It looks like `dx init` now requires GitHub PAT before checking local tooling, is this right?

The old flow checked Terraform/Corepack installation before GitHub credentials. The new flow requires the PAT first. A user without Terraform would now see a "GitHub PAT required" error instead of the more actionable "Terraform not installed" message. 

Consider running `checkInitPreconditions()` before the `requireGitHubAuth()` call.

kin0992:
> **thought:** I am wondering why `help` is not enough for this use case. Do you have any supportive data or recommended best practices that support this solution?
> 
> Maybe you can use [evals](https://developers.openai.com/api/docs/guides/agent-evals) to rank the agent ability to find DX commands with `help` and `spec`

@lucacavallaro the rationale behind this command is that with `--help` the agent must run many commands (e.g. `dx init --help`, `dx add environment --help` and so on) while with spec the agent can have the whole specifications running just a single command.

I would not run any benchmark in this PR, but we could totally do in a separate task (that we need to create)

lucacavallaro (approved):

PR #1792, opened 2026-05-26
Add `dx-tasks` package and `run-dx-task` Github Action

lucacavallaro:
`dx-tasks` is a new package that contains our pure TypeScript implementations of our custom scripts used in GitHub Workflows and other abstractions like Nx Plugins and DX CLI.

this change implements `terraformPlan` and a basic `dispatcher` that is used to make these tasks runnable for workflows

the plan is to share the code between `Nx` and `GitHub Workflow`, to keep it DRY - later we will backport this logic to `infra_plan.yaml` deprecating the current action


actions/dx-tasks-action/package.json
```diff
@@ -0,0 +1,28 @@
+{
+  "name": "dx-tasks-action",
+  "version": "0.1.0",
```
kin0992: **note**: this should start from `0.0.0` and then nx release will handle the bump

actions/dx-tasks-action/rspack.config.ts
kin0992, line 1: **question**: why rspack and not tsup like other actions?

lucacavallaro: Replaced now with [`tsdown`](https://tsdown.dev/)

[`tsup`](https://github.com/egoist/tsup) is not actively maintained anymore, so we have to find another solution for bundling our code.

I tried using `esbuild` directly (since `tsup` was just a wrapper on esbuild), but I was not satisfied by the Nx integration, then I tried `rspack` that has first class support with a dedicated plugin and finally, following the suggestion of @gunzip, I replaced it with `tsdown` that is similar to `tsup` and officially sponsored/supported by void0 (the same authors of vitest and oxc, tools we already like and use).

In a later PR I will replace all instances of `tsup` with `tsdown` (which also supports monorepos, so we can have only one root configuration for bundling).

packages/dx-tasks/package.json
```diff
@@ -0,0 +1,48 @@
+{
```
gunzip, line 1: issue: this package misses a README.md that explains how to run tasks locally and list all availlable tasks

packages/dx-tasks/src/terraform-plan.ts
```diff
@@ -0,0 +1,86 @@
...
+    .entries()
+    .reduce(
+      (acc, [key, value]) =>
+        `${acc}-${key}${value === true ? "" : `=${value}`} `,
+      "",
+    );
+
+  const result = await runCommand("terraform", ["plan"], modulePath, env);
+
+  if (result.signal) {
```
gunzip, line 70: **issue**: missing exit code check in terraformPlan, silent Terraform failures

The `terraformPlan` function checks for process killed-by-signal (`result.signal`) but never checks `result.exitCode`. terraform plan exits with:

 - 0 — no changes
 - 1 — error (invalid config, state lock failure, provider issues)  
 - 2 — changes detected (normal)

When Terraform fails with exit code 1, the function quietly masks/prints the error output but never throws — the GitHub Action succeeds even when Terraform has failed, silently masking real infrastructure errors.

The original shell implementation (actions/sanitize-terraform-plan/run_plan.sh) explicitly handles this:

```sh
 if [[ "$PLAN_EXIT_CODE" -eq 1 ]]; then
   echo "::error::Terraform plan failed with a critical error."
   exit 1
 fi
```

The TypeScript port has zero references to exitCode, and there is no test coverage for exitCode: 1 (tests only pass exitCode: 0 or exitCode: null).

Suggested fix: After the signal check, add an exit code check that throws on exit code 1 while allowing codes 0 and 2 to continue normally.

packages/dx-tasks/src/default-dispatcher.ts
```diff
@@ -0,0 +1,25 @@
...
+  type TaskDefinition,
+  type TaskDispatcher,
+} from "./dispatcher.ts";
+import {
+  terraformPlan,
+  type TerraformPlanPayload,
+  payloadSchema as terraformPlanPayloadSchema,
+} from "./terraform-plan.ts";
+
+export const terraformPlanTask: TaskDefinition<TerraformPlanPayload> = {
```
kin0992, line 14: **note**: it looks like `terraformPlanTask` is used in this file only, so we could remove the `export`

packages/dx-tasks/package.json
```diff
@@ -0,0 +1,48 @@
...
+  "name": "@pagopa/dx-tasks",
+  "version": "0.2.1",
```
kin0992: **issue**: start with version `0.0.0`

kin0992 (changes requested):

packages/dx-tasks/src/default-dispatcher.ts
lucacavallaro: it is exported as part of the dispatcher API (consumers can create their own dispatchers, choosing a subset of these tasks)

packages/dx-tasks/src/terraform-plan.ts
lucacavallaro: it currently rejects on errors and the error is handled by the orchestrator (nx, github actions ..) 

I addressed it in a separate change, you will see it with the next pr

packages/dx-tasks/package.json
lucacavallaro: Added a `README.md`, note that this is just a library meant to be used by task runner/orchestrators (Nx, github actions, other abstractions...) 

it exports tasks and a simpler dispatcher, but it does not allow running tasks directly

actions/run-dx-task/.gitignore
krusty93, line 1: thought: can we have a single `.gitignore` with this exception in the entire `actions` folder instead of duplicating it everytime? It can be done in another PR

packages/dx-tasks/README.md
```diff
@@ -0,0 +1,54 @@
+# @pagopa/dx-tasks
+
+Utility task helpers and the built-in task dispatcher for DX packages.
```
gunzip: suggestion: clarify this description

actions/run-dx-task/.gitignore
lucacavallaro: I think we can stop targeting `dist/` and write the action bundle in the project root

packages/dx-tasks/README.md
lucacavallaro: it is ok now?

kin0992 (approved):

gunzip (approved):

PR #1797, opened 2026-05-26
Redesign docs sidebar navigation

krusty93:
The DX docs sidebar had become visually noisy and hard to scan. This PR prototypes a cleaner, Terramate-inspired navigation model while preserving the existing DX color identity.

## What changed

- replaces the autogenerated sidebar with a manually grouped structure built around non-clickable categories and predictable open and closed sections
- restyles spacing, typography, separators, indentation, and open-state rails to make the hierarchy easier to scan without changing site colors
- splits the DX CLI docs into dedicated requirements, installation, and usage pages so the new navigation has focused leaf nodes
- adds experimental product SVG icons for top-level categories
- includes a minor release plan for `docs`

## Notes for review

- the sidebar structure now relies on native Docusaurus category behavior instead of custom caret rendering, which keeps the implementation simpler and more stable
- the icon experiment is easy to roll back by disabling `useExperimentalSidebarIcons` in `apps/website/sidebars.ts` or removing `sidebar-icons.css` from `apps/website/src/css/custom.css`


krusty93:
Example:
image

apps/website/sidebars.ts
```diff
@@ -1,31 +1,307 @@
...
-  /*
   tutorialSidebar: [
-    'intro',
-    'hello',
     {
-      type: 'category',
-      label: 'Tutorial',
-      items: ['tutorial-basics/create-a-document'],
+      className: categoryClass("overview"),
+      collapsed: false,
```
mamu0: [suggestion(non-blocking)](https://pullpo.io/cc?l=suggestion&d=non-blocking) 
Great work on the new sidebar layout! Just one thought: upon loading the page, multiple sections are expanded by default, drawing attention away from the collapsed ones.
It might be cleaner to have them all collapsed initially so users can choose what to open. What do you think?

apps/website/sidebars.ts
```diff
@@ -1,31 +1,307 @@
...
+            },
+          ],
+          label: "Pull Requests",
+          type: "category",
+        },
+        {
+          className: categoryClass("workflows"),
+          collapsed: true,
+          items: [
+            { id: "pipelines/index", label: "Overview", type: "doc" },
```
mamu0: [suggestion(non-blocking)](https://pullpo.io/cc?l=suggestion&d=non-blocking) 
I see that 'Overview' is listed as the first item in nearly all sections. To streamline the sidebar and reduce redundancy, how about mapping the Overview content to the section title click instead of having a separate page?

apps/website/sidebars.ts
krusty93: I was inspired by [terramate](https://terramate.io/docs/) docs, which keeps open all the sections except for the less important ones. So I did the same with the ones I think are less important. It's just a matter of taste, the result still looks good:
image

If you all prefer to keep them closed, it's ok to me!

apps/website/sidebars.ts
krusty93: Two reasons:
- making easier to expand/reduce a section
- making clear that the item doesn't include information

Otherwise changing behavior according to the clicked area might be misleading to me. We can rename some `Overview` pages though

apps/website/sidebars.ts
mamu0: I like it this way!

apps/website/sidebars.ts
mamu0: Renaming them is ok to me.
Maybe we can delete some of them, if I remember correctly, we only added them as dummy pages for when users clicked on the section title.

krusty93: Done

apps/website/sidebars.ts
```diff
@@ -1,31 +1,310 @@
...
+            },
+            "github/pull-requests/format",
+            "github/pull-requests/acceptance-criteria",
+            "github/pull-requests/auto-merge",
+            "github/pull-requests/changeset",
+            "github/pull-requests/version-plan",
+            {
+              id: "github/pull-requests/code-review/index",
+              label: "Code Review",
+              type: "doc",
```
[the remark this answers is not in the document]
krusty93: fixed

mamu0 (approved):
:shipit:

PR #1807, opened 2026-05-28
Remove codecov.yml

lucacavallaro:
We don't use `codecov` anymore, this configuration was just a test


gunzip (approved):
lgtm but it breaks ci

PR #1810, opened 2026-05-29
Isolate build from AWS credentials in MCP Server deploy

gunzip:
## Summary

Split the single `build-and-deploy` job in both AWS MCP Server deploy workflows into separate **build** and **deploy** jobs to enforce a security boundary between untrusted build steps and CSP credentials.

### Problem

The Docker build step was executing with AWS OIDC credentials already configured in the environment (via `csp-login`). If the Dockerfile, a dependency, or a supply-chain compromised action were malicious, it could exfiltrate or abuse those credentials.

### Solution

- **Build job**: Builds the Docker image and pushes it to GHCR. No CSP login, no `ROLE_ARN` secret. Uses `id-token: write` only for Sigstore attestation.
- **Deploy job**: Authenticates to AWS (OIDC), copies the image from GHCR to ECR using `crane`, then deploys to Lambda.

This pattern is consistent with the approach already used by `release-azure-appsvc-v1`, `release-azure-staticapp-v1`, and `function_app_deploy`.

### Affected workflows

- `_release-aws-mcp-server-deploy-dev-v1.yaml`
- `_release-aws-mcp-server-deploy-v1.yaml`

### DIFF
--- a/.github/workflows/_release-aws-mcp-server-deploy-dev-v1.yaml
+++ b/.github/workflows/_release-aws-mcp-server-deploy-dev-v1.yaml
@@ -12,35 +12,31 @@ on:
       - "packages/mcp-prompts/**"
   workflow_dispatch:
 
+env:
+  IMAGE_NAME: dx/mcp-server
+
 jobs:
-  build-and-deploy:
-    name: Build and Deploy MCP on AWS Lambda
+  build:
+    name: Build Docker Image
     runs-on: ubuntu-latest
-    environment: app-dev-cd
-    env:
-      AWS_REGION: eu-central-1
-      ROLE_ARN: ${{ secrets.ROLE_ARN }}
-      LAMBDA_FUNCTION_NAME: dx-d-euc1-mcp-server-lambda-01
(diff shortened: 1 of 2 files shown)


christian-calabrese (approved):

lucacavallaro (approved):

gunzip:
Closing since this is defense-in-depth, not a fix for an exploitable vulnerability. The split adds complexity (two jobs, GHCR→ECR crane copy, binary download risk) for a threat that requires a zero-day in the Docker/buildx toolchain.

 In a standard docker build (no `--privileged`, no host volume mounts, no `--build-arg` passing secrets):

   - The build container is isolated from the host
   - Host environment variables are not accessible
   - The host filesystem is not accessible

So under normal operation, a malicious Dockerfile cannot reach the AWS credentials that csp-login sets on the runner. The original single-job workflow was not practically vulnerable.

The residual risk is narrow:
   - A vulnerability in docker/build-push-action itself (it runs as a JS action on the runner, with access to env vars), but it's pinned to a SHA
   - A buildx container escape (theoretical, would be a major CVE)

PR #1818, opened 2026-06-04
Update dependencies across the project

gunzip:
Upgrade various dependencies in the project to their latest versions, ensuring compatibility and improved functionality. This includes updates to testing libraries, AWS SDK, Azure monitoring tools, and other essential packages. Additionally, make minor adjustments to configuration files for better performance and compatibility.


kin0992 (approved):

krusty93 (approved):

apps/mcpserver/tsconfig.app.json
```diff
@@ -1,11 +1,7 @@
 {
   "extends": "./tsconfig.json",
   "include": ["./src/**/*"],
-  "exclude": [
-    "node_modules",
-    "**/*.spec.ts",
-    "**/*.test.ts"
-  ],
+  "exclude": ["node_modules", "**/*.spec.ts", "**/*.test.ts"],
```
lucacavallaro: **suggestion(non-blocking):** I noticed that `include` / `exclude` are not written in the same way across the monorepo

I think we should stick to a general rule/pattern

Something like

```json
{
  "include": ".src/**/*.ts",
  "exclude": ["**/__tests__/**"]
}
```

1. **always use** `include`, with explicit glob like `./src/**/*.ts`
2. when use `exclude`, try to minimize the list (if we exclude `__tests__` maybe we don't need to add `*.spec.ts`)

`exclude` excludes file starting from the glob specified in `include`, so they should contain only files selected by `include` (in this example, `node_modules` is not necessary, since we are using `src/**/*.ts` as base)

gunzip: done!

PR #1830, opened 2026-06-09
Update TypeScript instructions

lucacavallaro:


gunzip (approved):

lucacavallaro:
Changed my mind about these instructions, it is better to enforce them via `eslint`

PR #1831, opened 2026-06-09
Update pnpm to v11

lucacavallaro:


pnpm-workspace.yaml
```diff
@@ -6,6 +6,22 @@ packages:
...
+allowBuilds:
+  '@swc/core': true
+  core-js: false
+  cpu-features: true
+  esbuild: false
+  nx: true
+  protobufjs: false
+  sharp: false
+  ssh2: true
+  unrs-resolver: false
```
gunzip, line 23: isn't there a default? why specify both true and false?

lucacavallaro: No, is one of the changes with v11 https://pnpm.io/it/blog/releases/11.0#allowbuilds-replaces-the-old-build-settings

If you don't specify `true` or `false`, you will get a warning at install time

I accurately revised these dependencies manually, and enabled only the useful-mandatory builds for us

lucacavallaro: (I think it is a better approach, it forces you to think about your dependencies :D and these allowlist can be distributed by pnpm config dependencies)

gunzip (approved):

PR #1833, opened 2026-06-10
Add Telemetry in Infra plan and apply

mamu0:
This PR fixes the `setup-telemetry` action and updates both the `infra_plan.yaml` and `infra_apply.yaml` workflows to include the telemetry setup for exporting data to Application Insights.

Examples:

- `dx` plan **(UPDATED)**: https://github.com/pagopa/dx/actions/runs/27347929284
- `dx-playground` plan **(UPDATED)**: https://github.com/pagopa/dx-playground/actions/runs/27339975207
- `dx-playground` apply: https://github.com/pagopa/dx-playground/actions/runs/27271001513
- `dx-playground` plan in **non default** application insight: https://github.com/pagopa/dx-playground/actions/runs/27279574701

Result in default application insight:
image

Resolves: CES-1928 CES-1929


actions/setup-telemetry/action.yaml
```diff
@@ -5,7 +5,6 @@ inputs:
   connection_string:
     description: "Azure Application Insights connection string"
     required: false
-    default: "InstrumentationKey=e0ff8094-78fa-45e5-a21d-e62b453dc5d1;IngestionEndpoint=https://italynorth-0.in.applicationinsights.azure.com/;LiveEndpoint=https://italynorth.livediagnostics.monitor.azure.com/;ApplicationId=ce469d55-2ff7-4dfd-a249-cc787291e672"
```
krusty93, line 8: issue: This connection string should stay here (or in the yaml) as client repositories don't have it (and we can't do nothing)

actions/setup-telemetry/src/index.ts
```diff
@@ -60,28 +79,28 @@ async function run(): Promise<void> {
...
       { cause: err },
     );
   }
 
-  const eventsFile = `${SESSION_DIR}/events.ndjson`;
-
   // Initialize empty events file
   try {
-    await fs.writeFile(eventsFile, "", { flag: "wx" });
+    await fs.writeFile(eventsFile, "", { flag: "w" });
```
krusty93, line 105: question: what's this change?

.github/workflows/_release-terraform-apply-prod-resources.yaml
```diff
@@ -20,3 +20,4 @@ jobs:
       override_github_environment: infra-prod
       use_labels: true
       use_private_agent: true
+      connection_string: "InstrumentationKey=e0ff8094-78fa-45e5-a21d-e62b453dc5d1;IngestionEndpoint=https://italynorth-0.in.applicationinsights.azure.com/;LiveEndpoint=https://italynorth.livediagnostics.monitor.azure.com/;ApplicationId=ce469d55-2ff7-4dfd-a249-cc787291e672"
```
lucacavallaro: **question:** I noticed that the CS is already hardcoded into the action code (as `DEFAULT_CONNECTION_STRING`). Why is this hardcoded here as well?

krusty93 (changes requested):
Great it resumed back to work! 🙏🏻 

If I remember correctly, the `shutdown` calls the `flush` internally. Or, why the `flush` stopped to work? Or again, why does the `shutdown` not work [here](https://github.com/pagopa/dx/compare/CES-1928-instrumentare-con-application-insights-la-pipeline-di-terraform-plan?expand=1#diff-d5b818ef6f341283df70ddfc75057041e50e34dbd27f1ca25c83e6a1d2c9a336R323)?

I still see few issues:
1. the connection string must be fixed: users shouldn't and don't need to change it; it's our telemetry, not theirs

2. I see some parsing errors in almost all the examples linked:
```
Post Setup Telemetry Session
Post job cleanup.
Skipping malformed JSON telemetry line
Skipping malformed JSON telemetry line
Telemetry flushed
```
I think they're related to the custom span [you set](https://github.com/pagopa/dx/pull/1833/changes#diff-ae911cd5a54ded7aa219543aba3574eb16a37e08ab9dd46407c65d9c54771896R369) that doesn't appear in [AI](https://portal.azure.com/#blade/AppInsightsExtension/DetailsV2Blade/DataModel/%7B%22eventId%22:%224de51314-64ae-11f1-aafe-7ced8d027e0b%22,%22timestamp%22:%222026-06-10T09:24:18.963Z%22%7D/ComponentId/%7B%22Name%22:%22dx-d-itn-common-appi-01%22,%22ResourceGroup%22:%22dx-d-itn-common-rg-01%22,%22SubscriptionId%22:%2235e6e3b2-4388-470e-a1b9-ad3bc34326d1%22%7D). I don't remember exactly, but [this](https://github.com/pagopa/dx/compare/CES-1928-instrumentare-con-application-insights-la-pipeline-di-terraform-plan?expand=1#diff-20d259a4310481c4b635aa147f2767b7b8a59ca77af97b40108ccabbe1b3784fR23) could be the required fix

3. I strongly suggest to bring back the [integration with GitHub SDK](https://github.com/pagopa/dx/compare/CES-1928-instrumentare-con-application-insights-la-pipeline-di-terraform-plan?expand=1#diff-d5b818ef6f341283df70ddfc75057041e50e34dbd27f1ca25c83e6a1d2c9a336R9) as current logs aren't reliable at all. Of course it can be done in a subsequent PR

.github/workflows/_release-terraform-apply-prod-resources.yaml
mamu0: Is different, the default one is for the DEV application insight, this is the PROD one

actions/setup-telemetry/action.yaml
mamu0: i've moved [here](https://github.com/pagopa/dx/pull/1833/changes#diff-74db3c2d7ea13a219652312196b9c4104e5b5ee5b4de03571928d2377fdeb54dR16) this default, this way it was easier for me to create the input for the workflow without a default

actions/setup-telemetry/src/index.ts
mamu0: Using `wx` creates the file only if it does not exist and throws an error if it does, whereas `w` creates or truncates the file, ensuring that every session starts with an empty file.
This completely prevents residual lines, making the output fully deterministic.

mamu0:
@krusty93 Thanks for the feedback!
> 1. the connection string must be fixed: users shouldn't and don't need to change it; it's our telemetry, not theirs

About the connection string: I agree that if this is DX-owned telemetry, users should not choose the target Application Insights.

What is not fully clear to me is where we want to enforce that contract. Today the action has a fallback/default connection string, while the reusable workflows can also pass one explicitly. If we want a single DX telemetry sink for every environment and workflow, then I think the cleaner contract is either:

1. keep the connection string fully internal to `setup-telemetry` and remove the workflow input entirely, **OR**
2. make `connection_string` required in the action and pass a fixed value from the reusable workflows.

Option 1 seems simpler if the target is always the same for all infra plan/apply workflows. Could you confirm that this telemetry must always go to the same DX-owned Application Insights, regardless of caller/environment?

actions/setup-telemetry/src/post.ts
```diff
@@ -236,19 +261,10 @@ async function post(): Promise<void> {
...
-    };
-    const loggerProvider = logs.getLoggerProvider() as {
-      forceFlush?: (options?: { timeoutMillis?: number }) => Promise<void>;
-    };
-    await Promise.all([
-      tracerProvider.forceFlush?.({ timeoutMillis: FLUSH_TIMEOUT_MS }),
-      loggerProvider.forceFlush?.({ timeoutMillis: FLUSH_TIMEOUT_MS }),
-    ]);
+    await tracerProvider.shutdown();
+    await loggerProvider.shutdown();
```
kin0992: **note**: if you do not have the order requirement (so the execution order of the `shutdown` functions is not relevant), you can run in parallel
```typescript
await Promise.all([tracerProvider.shutdown(), loggerProvider.shutdown()]);
```

actions/setup-telemetry/package.json
```diff
@@ -23,12 +23,12 @@
...
   "license": "UNLICENSED",
   "dependencies": {
+    "@azure/monitor-opentelemetry-exporter": "catalog:opentelemetry",
     "@opentelemetry/api": "catalog:opentelemetry",
     "@opentelemetry/api-logs": "catalog:opentelemetry",
-    "@opentelemetry/sdk-metrics": "catalog:opentelemetry",
     "@opentelemetry/resources": "catalog:opentelemetry",
+    "@opentelemetry/sdk-logs": "catalog:opentelemetry",
     "@opentelemetry/sdk-trace-base": "catalog:opentelemetry",
-    "@pagopa/azure-tracing": "workspace:^",
```
kin0992, line 30: **question**: why the azure-tracing has been removed? is it non-compatible with this use case?
if so, what issues did you found with it?

.github/workflows/_release-terraform-apply-prod-resources.yaml
lucacavallaro: why does the default one point to `dev`? our dev environment is ephemeral, in my opinion the default should be `prod`

krusty93:
> @krusty93 Thanks for the feedback!
> 
> > 1. the connection string must be fixed: users shouldn't and don't need to change it; it's our telemetry, not theirs
> 
> About the connection string: I agree that if this is DX-owned telemetry, users should not choose the target Application Insights.
> 
> What is not fully clear to me is where we want to enforce that contract. Today the action has a fallback/default connection string, while the reusable workflows can also pass one explicitly. If we want a single DX telemetry sink for every environment and workflow, then I think the cleaner contract is either:
> 
> 1. keep the connection string fully internal to `setup-telemetry` and remove the workflow input entirely, **OR**
> 2. make `connection_string` required in the action and pass a fixed value from the reusable workflows.
> 
> Option 1 seems simpler if the target is always the same for all infra plan/apply workflows. Could you confirm that this telemetry must always go to the same DX-owned Application Insights, regardless of caller/environment?

- Agree with option 1
I can confirm (it was a team made decision taken during a stand up and a brief Slack follow up)

actions/setup-telemetry/action.yaml
mamu0: it has been reverted [999c5ee](https://github.com/pagopa/dx/pull/1833/commits/999c5ee7e70f2df6572f4a8249461715fe9f9008)

.github/workflows/_release-terraform-apply-prod-resources.yaml
mamu0: As decided in [this comment](https://github.com/pagopa/dx/pull/1833#issuecomment-4678260324) with @krusty93, I reverted to the default inside **action**, without requiring a configuration from the reusable workflow.

krusty93: BTW be sure to replace the existing connection string with the prod one before merging the PR

actions/setup-telemetry/src/index.ts
krusty93: Wouldn't be better if it fails if you use it improperly like initializing it twice?

actions/setup-telemetry/package.json
mamu0: `@pagopa/azure-tracing` is not incompatible in general. I removed it here because this GitHub Action post step needs explicit ownership of the trace/log providers and exporters lifecycle. The previous global-provider path was not reliably waiting for all Azure Monitor sends before Node exited, while the direct exporters plus `shutdown()` make the flush lifecycle explicit for this one-shot action.

actions/setup-telemetry/src/post.ts
mamu0: thanks for the suggestion, i've added this here [a757d82](https://github.com/pagopa/dx/pull/1833/commits/a757d82cf2354685b138e4111ec74d7b6eb0ad6d)

mamu0:
cc. @krusty93

> 1. the connection string must be fixed: users shouldn't and don't need to change it; it's our telemetry, not theirs

I reverted the connection string to its original state, meaning no input is required for the reusable workflow. It is now handled solely within the action, which defaults to Application Insights in PROD-DEVEX.

> 2. I see some parsing errors in almost all the examples linked:

The parsing errors you identified in the post setup job have been resolved. Here are two examples of successful executions: [dx](https://github.com/pagopa/dx/actions/runs/27347929284), [dx-playground](https://github.com/pagopa/dx-playground/actions/runs/27339975207)

> 3. I strongly suggest to bring back the [integration with GitHub SDK](https://github.com/pagopa/dx/compare/CES-1928-instrumentare-con-application-insights-la-pipeline-di-terraform-plan?expand=1#diff-d5b818ef6f341283df70ddfc75057041e50e34dbd27f1ca25c83e6a1d2c9a336R9) as current logs aren't reliable at all. Of course it can be done in a subsequent PR

Agree on bringing back the GitHub SDK integration, but I would keep it as a follow-up PR. This PR is focused on restoring export reliability and fixing malformed telemetry records.

.github/workflows/_release-terraform-apply-prod-resources.yaml
mamu0: done here [583d937](https://github.com/pagopa/dx/pull/1833/commits/583d937b5fd70635a1bcdeae7d9f25be8d44f173)

actions/setup-telemetry/src/index.ts
mamu0: Good point. I agree that initializing telemetry twice in the same job should fail.

I kept `w` for the events file because file existence is not a reliable signal for double initialization: on reused/self-hosted workspaces it can also be stale state from a previous run, and in that case we want a clean deterministic session.

To cover both cases, I added in this commit [21f4e9d](https://github.com/pagopa/dx/pull/1833/commits/21f4e9d66440f0d63a04602eb00620d7299f47db) an explicit guard on `OTEL_EVENT_FILE`: if `setup-telemetry` is called again after a session has already been initialized in the same job, it now fails with a clear error. The file is still truncated with `w` only for the valid first initialization.

krusty93: Maybe it's the wrong commit as the only thing I see is the git ref?

mamu0: Sorry, now is correct

krusty93:
> cc. @krusty93
> 
> > 1. the connection string must be fixed: users shouldn't and don't need to change it; it's our telemetry, not theirs
> 
> I reverted the connection string to its original state, meaning no input is required for the reusable workflow. It is now handled solely within the action, which defaults to Application Insights in DEV-DEVEX.
> 

You mean PROD-DEVEX?

> > 2. I see some parsing errors in almost all the examples linked:
> 
> The parsing errors you identified in the post setup job have been resolved. Here are two examples of successful executions: [dx](https://github.com/pagopa/dx/actions/runs/27347929284), [dx-playground](https://github.com/pagopa/dx-playground/actions/runs/27339975207)
> 

What was the issue?

mamu0:
> You mean PROD-DEVEX?

Yes

> What was the issue?

The "malformed JSON" warning was actually a Zod schema crash. `telemetryLineSchema` used `z.discriminatedUnion("type")` but defined type as `.optional().default(...)`, causing a Duplicate discriminator value "undefined" error. This error was caught by the broad `try/catch` block, masking the real issue.

**Fix**: Made type required in the schema, handled backward compatibility by pre-inferring type before validation, and isolated `JSON.parse` into its own `try/catch` [30a95da](https://github.com/pagopa/dx/pull/1833/commits/30a95dab47f5a5fd8dca6ae1fba1c9f800d3a4ea).

krusty93 (approved):

PR #1834, opened 2026-06-11
Add VSCode launch and task configuration for dx CLI

krusty93:
Introduce a launch configuration and a build task for the dx CLI in VSCode, enabling streamlined debugging and building processes. It supports breakpoints, variables evaluation, call stack and more.

From now on, `F5` is the way to go.

Apart from dx-cli, this setup can be replicated to any typescript project, including systems in production

### DIFF
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -1,0 +1,20 @@
+{
+  "version": "0.2.0",
+  "configurations": [
+    {
+      "name": "dx CLI",
+      "type": "node",
+      "request": "launch",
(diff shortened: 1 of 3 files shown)


krusty93:
Breakpoint with local variable evaluation and custom expressions (`watch window`):

image

Conditional breakpoint setup. It supports conditions, a custom log when the breakpoint is hit, and an hit counter:
image

Evaluation of an expression inline (I can see the trimmed value before the application actually executed the `trim()`). The `watch` window is locally persistent across executions:
image

lucacavallaro (approved):

PR #1835, opened 2026-06-11
Add doctor path flag

kin0992:
Allow `dx doctor` to inspect a repository from an explicit path while keeping the full doctor check set enabled. This makes the command easier to run from automation or parent directories without changing the current validation behavior.

Resolves CES-2133

### DIFF
--- a/.nx/version-plans/version-plan-1781173004636.md
+++ b/.nx/version-plans/version-plan-1781173004636.md
@@ -1,0 +1,5 @@
+---
+"@pagopa/dx-cli": patch
+---
+
+Add doctor command repository path flag
--- a/apps/cli/README.md
+++ b/apps/cli/README.md
@@ -75,14 +75,20 @@ Verify the repository setup according to the DevEx guidelines.
 
 ```bash
 dx doctor
+dx doctor --path apps/cli
+dx doctor -p apps/cli
 ```
 
 This command will:
 
 - Check if you're in a valid Git repository
 - Validate that required monorepo scripts are present in package.json
-- Check that the `turbo.json` file exists
-- Verify that the installed `turbo` version meets the minimum requirements
+- Verify that `.pre-commit-config.yaml` is present
+- Check that `nx.json` exists and that the installed Nx version meets the minimum requirements
+- Verify that pnpm workspaces are configured
+
+Use `-p, --path` to inspect a repository other than the current working
+directory.
 
 **Example output:**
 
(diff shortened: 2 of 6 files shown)


.nx/version-plans/version-plan-1781173004636.md
```diff
@@ -0,0 +1,5 @@
+---
+"@pagopa/dx-cli": patch
+---
+
+Add doctor command repository path flag
```
gunzip, line 5: question: in which cases do we want to run doctor against a repository which is not the cwd?

kin0992: Actually, this is an alternative access point. Instead of iterating through every repository and run `dx doctor`, with this PR it could be possible to fetch upfront the repo paths and iterate through them to run `dx doctor -p <path>`.

If we do not think this is necessary, I could close the PR

gunzip: I'm not deployed. How and when do you find this useful?

kin0992: The only use case I found useful is when I want to use the DX CLI not inside a repo (like I am testing the init command, then I want to run the doctor command on the new or an existing repo). Or in general when I am not in the repository folder, but I still need to run the doctor command, having the `-p` flag is useful (I can also use the tabs to autocomplete paths).

But still I think this can improve for DX, I'm not sure it improves AX.

gunzip: let's see what others think about this @pagopa/engineering-team-devex

krusty93: I do like the path flag as it enables (or makes simpler) automation

lucacavallaro: in my opinion, this option makes sense only if it is available to the whole dx-cli and not only for doctor

a similar pattern is adopted by `terraform CLI`, as it has a `-chdir` option that applies to all subcommands

`terrafrom -chdir=/path/to/module plan`
`terraform -chdir=/path/to/module console`

(notice that `-chdir` is used after the main command "`terraform`" not not after the subcommand)

mamu0: Maybe is unnecessary for an agent that can simply execute a `cd` command behind the scenes, but i think it could be helpful to speed up manual operations, and it would be easier to use in an automated context

kin0992:
we have decided not to proceed right now and evaluate a similar solution in future if needed

PR #1836, opened 2026-06-11
Wire info command to presenter

kin0992:
Route the info command output through the shared command presenter so text and JSON modes use the same CLI output path. Remove the old domain-level logger output helper and cover the JSON presenter envelope with a command test.

Closes CES-2139


kin0992:
Command outcome:

Screenshot 2026-06-11 at 15 05 07

gunzip (approved):

PR #1838, opened 2026-06-11
Integrate Application Insights and OpenTelemetry for CLI observability

krusty93:
Implement OpenTelemetry logging for improved observability, including root span tracing for CLI commands. Enhance telemetry capabilities by concurrently fetching tool versions and adding Azure Monitor service name.


krusty93:
Few screenshots:

Operation duration and status:
image


Log events using logtape open telemetry sink (`console.error/info/debug/etc.`):
image

Each span logs the dependencies:
image

gunzip:
Since the CLI is a public package available to anyone, any tracking mechanism must be opt-in and accompanied by a privacy policy. The same applies to pipelines. Alternatively, we must restrict tracking to internal usage only

krusty93:
> Since the CLI is a public package available to anyone, any tracking mechanism must be opt-in and accompanied by a privacy policy. The same applies to pipelines. Alternatively, we must restrict tracking to internal usage only

restricted to pagopa users only, via github

apps/cli/src/index.ts
```diff
@@ -28,17 +37,49 @@ import { listCodemods } from "./use-cases/list-codemods.js";
...
+};
+
 const configureLogging = async (verbose: boolean): Promise<void> => {
   const level = verbose ? "debug" : "info";
+  // Wire logtape logs to the same OTel LoggerProvider that useAzureMonitor()
+  // registered globally. When telemetry is disabled the provider is a no-op,
+  // so this is always safe to call unconditionally.
+  const otelSink = getOpenTelemetrySink({
+    loggerProvider: logs.getLoggerProvider() as unknown as LoggerProvider,
+  });
```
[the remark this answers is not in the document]
krusty93: fixed

pnpm-workspace.yaml
```diff
@@ -48,30 +48,31 @@ catalog:
...
+    "@azure/monitor-opentelemetry-exporter": ^1.0.0-beta.42
+    "@logtape/otel": 1.3.8
+    "@opentelemetry/api": ^1.9.1
+    "@opentelemetry/api-logs": ^0.218.0
+    "@opentelemetry/instrumentation": ^0.218.0
+    "@opentelemetry/instrumentation-undici": ^0.28.0
+    "@opentelemetry/resources": ^2.7.1
+    "@opentelemetry/sdk-metrics": ^2.7.1
+    "@opentelemetry/sdk-trace-base": ^2.7.1
+    "@opentelemetry/sdk-logs": ^0.218.0
```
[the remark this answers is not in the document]
krusty93: fixed

apps/cli/src/adapters/azure-monitor/instrumentation.ts
```diff
@@ -0,0 +1,93 @@
...
+  if (azureMonitorEnabled) {
+    return;
+  }
+  azureMonitorEnabled = true;
+
+  useAzureMonitor({
+    azureMonitorExporterOptions: {
+      connectionString:
+        process.env.APPLICATIONINSIGHTS_CONNECTION_STRING ||
+        "InstrumentationKey=e0ff8094-78fa-45e5-a21d-e62b453dc5d1;IngestionEndpoint=https://italynorth-0.in.applicationinsights.azure.com/;LiveEndpoint=https://italynorth.livediagnostics.monitor.azure.com/;ApplicationId=ce469d55-2ff7-4dfd-a249-cc787291e672",
```
mamu0, line 78: [question(non-blocking)](https://pullpo.io/cc?l=question&d=non-blocking) 
The Application Insights connection string is currently hardcoded into the public npm package. Although the CLI only utilizes it for PagoPA members, anyone could extract it and send spoofed telemetry data directly to that endpoint. Is this an accepted risk, or would we prefer to inject it at build or runtime?

krusty93: Accepted risk, we can rotate or block connections at any time. Moreover we have a daily cap

kin0992: **note**: we could use managed identity. I did some experiments with that a few weeks ago, and the steps to use Managed Identity should be:
- force our Application Insights to use Managed Identity by adding `local_authentication_disabled = true`
- add `Monitoring Metrics Publisher` role to the instances that need to push telemetry data
- use the `@pagopa/azure-tracing` to initialize Azure Monitor using the Managed Identity (the package already supports Managed Identity authentication)

Please note that with this current approach, the Connection String will still be publicly available, but it won't be enough to send data.

kin0992: if it is not possible to use the `@pagopa/azure-tracing` package, then we should add here the logic to use the Managed Identity instead (through the `@azure/identity` package)

krusty93: The issue in using IAM is that dx cli is a client app, so it doesn't have any role but the user identity. This requires giving access to all the users.
An alternative is using service principals, but it requires saving a secret locally

apps/cli/src/adapters/azure-monitor/instrumentation.ts
```diff
@@ -0,0 +1,93 @@
...
+
+// Register the ESM module hook so that subsequent dynamic imports are
+// intercepted by OpenTelemetry's module instrumentation.
+register("import-in-the-middle/hook.mjs", import.meta.url, registerOptions);
+
+// Set service name before useAzureMonitor() initialises the Resource so
+// that Azure Monitor picks it up as the cloud_RoleName.
+process.env["OTEL_SERVICE_NAME"] = "dx-cli";
+process.env["ATTR_SERVICE_NAMESPACE"] = "dx";
+
```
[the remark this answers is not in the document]
gunzip: what about this comment @krusty93  ?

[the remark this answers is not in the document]
krusty93: It was right, I've implemented it.

For reference: https://opentelemetry.io/docs/specs/semconv/resource/service/
We have:
- service.namespace: `dx`
- service.name: `dx.{cli/infra_plan/infra_apply, etc}`.
- service.instance_id: `guid` (mapped as application id on AI)

mamu0 (approved):

PR #1844, opened 2026-06-12
Refactor devcontainer configuration and add mise.toml for tool management

gunzip:
Refactor the devcontainer setup by removing unnecessary features and integrating mise.toml for streamlined tool management. This change simplifies the configuration and enhances the development environment.

Closes CES-2127

### DIFF
--- a/.devcontainer/devcontainer.json
+++ b/.devcontainer/devcontainer.json
@@ -1,47 +1,13 @@
 {
   "image": "mcr.microsoft.com/devcontainers/base:debian@sha256:aa263c6fdd7c1bc990eead3922224fe8afbf5d85eaf3b6b9260786ab94d74933",
   "features": {
-    "ghcr.io/devcontainers-extra/features/pre-commit:2": {},
-    "ghcr.io/pagopa/devcontainer-features/acli:1": {},
-    "ghcr.io/devcontainers/features/azure-cli:1": {},
-    "ghcr.io/devcontainers/features/github-cli:1": {
-      "extensions": "seachicken/gh-poi"
-    },
-    "ghcr.io/devcontainers/features/terraform:1.4.3": {
-      "version": "latest",
-      "tflint": "latest",
-      "installTFsec": true,
-      "installTerraformDocs": true
-    },
-    "ghcr.io/pagopa/devcontainer-features/trivy:1": {},
-    "ghcr.io/devcontainers/features/node:2": {
-      "version": "24",
-      "installYarnUsingApt": false,
-      "pnpmVersion": "none",
-      "nvmInstallPath": "/opt/nvm"
-    },
-    "ghcr.io/devcontainers/features/go:1.3.4": {
-      "golangciLintVersion": "2.11.4"
-    },
+    "ghcr.io/devcontainers-extra/features/mise:1": {},
     "ghcr.io/devcontainers/features/docker-outside-of-docker:1.10.0": {
       "moby": false
     },
-    "ghcr.io/devcontainers-extra/features/act:1": {},
-    "ghcr.io/pagopa/devcontainer-features/shellcheck:1": {},
-    "ghcr.io/devcontainers-extra/features/nx-npm:1": {},
-    "ghcr.io/devcontainers/features/aws-cli:1": {},
-    "ghcr.io/devcontainers/features/copilot-cli:1.1.3": {},
     "ghcr.io/devcontainers-extra/features/apt-get-packages:1": {
-      "packages": "ripgrep,xdg-utils"
+      "packages": "xdg-utils"
     }
-  },
-  "postStartCommand": {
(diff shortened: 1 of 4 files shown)


mise.toml
```diff
@@ -0,0 +1,37 @@
...
+pre-commit = "4.6.0"
+qdns = "latest"
+ripgrep = "latest"
+shellcheck = "0.11.0"
+terraform = "1.15.2"
+terraform-docs = "0.24.0"
+tflint = "0.63.1"
+tfsec = "1.28.14"
+trivy = "0.71.0"
+uv = "latest"
```
lucacavallaro, line 22: **suggestion:** apart from `copilot`, I would not use `latest` on other tools

fix them to their latest stable available, or (better) `major.minor`

(tools like `azure-cli` can release a new major changing their API breaking our pipelines)

mise.toml
```diff
@@ -0,0 +1,37 @@
+[tools]
+acli = "latest"
+act = "latest"
+aws-cli = { version = "latest", symlink_bins = "true" }
+azure-cli = { version = "latest", uvx_args = "--prerelease=allow" }
+copilot = "latest"
+github-cli = "latest"
+go = "1.26.4"
+jq = "latest"
+node = "24.14.1"
```
lucacavallaro: **question:** can we specify only `24.14`? mise can handle that?

(is it necessary, for all tool, to specify the exact version? I think `major.minor` can be more useful if they follow semver)

mise.toml
```diff
@@ -0,0 +1,37 @@
...
+terraform-docs = "0.24.0"
+tflint = "0.63.1"
+tfsec = "1.28.14"
+trivy = "0.71.0"
+uv = "latest"
+
+[env]
+COREPACK_ENABLE_DOWNLOAD_PROMPT=0
+NX_DAEMON=true
+EDITOR="code --reuse-window --wait"
```
lucacavallaro: **issue:** this env is useful only for `code` users, and should not be set here

(for example, in my system `EDITOR` have a different value and I don't want to override that)

lucacavallaro: this value should be declared only in the context of the development container

mise.toml
```diff
@@ -0,0 +1,37 @@
...
+
+[tasks.setup-git]
+run = """
+#!/usr/bin/env bash
+gh extension install seachicken/gh-poi
```
lucacavallaro: **issue:** this setting in my opinion should be moved to the `postinstall` hook

https://mise.jdx.dev/hooks.html#preinstall-postinstall-hook

the `postinstall` hook is a good place to also run `corepack enable`

mise.toml
```diff
@@ -0,0 +1,37 @@
...
+git config --global --add safe.directory "$PWD"
```
lucacavallaro: **issue:** this command is relevant only in development container environment - so it should be removed from this task

.devcontainer/devcontainer.json
```diff
@@ -1,48 +1,14 @@
...
-    "ghcr.io/devcontainers-extra/features/nx-npm:1": {},
-    "ghcr.io/devcontainers/features/aws-cli:1": {},
-    "ghcr.io/devcontainers/features/copilot-cli:1.1.3": {},
     "ghcr.io/devcontainers-extra/features/apt-get-packages:1": {
-      "packages": "ripgrep,xdg-utils"
+      "packages": "xdg-utils"
     }
   },
-  "postStartCommand": {
-    "set-workspace-as-safe-dir": "git config --global --add safe.directory ${containerWorkspaceFolder}"
```
lucacavallaro, line 39: **suggestion:** keep this hook (it is needed only on container)

.devcontainer/devcontainer.json
```diff
@@ -1,48 +1,14 @@
...
-  },
-  "containerEnv": {
-    "COREPACK_ENABLE_DOWNLOAD_PROMPT": "0",
-    "NX_DAEMON": "true",
-    "EDITOR": "code --reuse-window --wait"
```
lucacavallaro, line 44: **suggestion:** keep this variable, is relevant only on development containers managed by vscode

.devcontainer/devcontainer.json
```diff
@@ -58,6 +24,7 @@
         "unifiedjs.vscode-mdx",
         "EditorConfig.EditorConfig",
         "bruno-api-client.bruno",
+        "hverlin.mise-vscode",
```
lucacavallaro: **thought:** I noticed that it is a 3rd-party extension, not an official one from the mise project. Do we really need that?

mise.toml
gunzip: I'd keep "latest" for every tool we don't care about breaking changes (e.g., tool which aren't used within pipelines). Look up the updated list.

lucacavallaro: looks good

lucacavallaro (approved):

PR #1852, opened 2026-06-15
Use package build scripts in DX Metrics Dockerfiles

gunzip:
## Summary

Build the DX Metrics Docker images by running each package build script from inside the package directory instead of re-entering pnpm workspace resolution from the pruned image workspace.

## Why

The current Dockerfiles install dependencies from a reduced workspace that intentionally copies only the DX Metrics app and its local shared package. A follow-up pnpm --filter ... build command triggers another workspace resolution pass from the repository root, which can fail on unrelated workspace-only packages that are not present in the image context.

Running npm run build from the target package keeps the build aligned with the script already defined in package.json, avoids duplicating the build command in the Dockerfile, and removes the workspace-resolution failure mode for both dx-metrics-import and dx-metrics.


mamu0 (approved):
LGTM

PR #1854, opened 2026-06-15
Add flags for add environment

kin0992:
Make dx add environment accept the agreed CLI flags and use them as prefilled answers while keeping the conditional initialization prompts interactive.

Add coverage for the new flag contract and the prefilled-answer prompt behavior.

Resolves CES-2134

### DIFF
--- a/apps/cli/src/adapters/commander/commands/add.ts
+++ b/apps/cli/src/adapters/commander/commands/add.ts
@@ -208,6 +208,20 @@ export const getEnvironmentInitialAnswers = async (
           );
         });
 
+  // Fail fast on an empty (or whitespace-only) key file: the file contents are
+  // never trimmed, so an empty value would otherwise skip the recovery prompt
+  // and surface as an opaque schema error (or a silently invalid key) later.
+  if (privateKey !== undefined && privateKey.trim().length === 0) {
+    throw new Error(
+      `Private key file at "${options.privateKeyPath}" is empty.`,
+    );
+  }
+
+  // Prefill runner-app credentials only when at least one credential flag was
+  // provided. Unspecified fields stay `undefined` on purpose so the generator
+  // prompts for the missing pieces and validates the complete set later. With
+  // no credential flag at all, leave this `undefined` to keep initialization
+  // fully interactive.
   const runnerAppCredentials =
     options.runnerAppId === undefined &&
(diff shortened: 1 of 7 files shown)


apps/cli/src/adapters/plop/generators/environment/prompts.ts
```diff
@@ -69,116 +72,377 @@ export const payloadSchema = z.object({
...
+  workspace: z
+    .object({
+      domain: workspaceSchema.shape.domain.optional(),
+    })
+    .optional(),
+});
+
+export type InitialAnswers = z.infer<typeof initialAnswersSchema>;
+
+const basePromptAnswersSchema = z.object({
```
gunzip, line 118: suggestion: It's not very clear for a naive reader what's the difference between `basePromptAnswersSchema` and `initialAnswersSchema`, I'd add a comment to clarify the intended usage.

apps/cli/src/adapters/commander/commands/add.ts
```diff
@@ -38,6 +50,188 @@ import {
 } from "../presenters/index.js";
 import { runAddEnvironmentPreconditions } from "./init.js";
 
+const appendOptionValue = (
+  value: string,
+  previous: string[] = [],
+): string[] => [...previous, value];
+
+const locationOptionSchema = z
```
kin0992, line 58: **suggestion**: I think this could be simplified.

I don't see any reason why this option cannot be a simple string, (like italynorth).

kin0992: it's ok for multi-account

apps/cli/src/adapters/commander/commands/add.ts
```diff
@@ -38,6 +50,188 @@ import {
...
+    options.privateKeyPath === undefined
+      ? undefined
+      : await readFile(options.privateKeyPath, "utf8").catch((cause) => {
+          throw new Error(
+            `Failed to read private key file at "${options.privateKeyPath}".`,
+            { cause },
+          );
+        });
+
+  const runnerAppCredentials =
```
gunzip, line 225: suggestion: add a comment

kin0992: updated

gunzip (approved):

PR #1856, opened 2026-06-16
Enable Nx Docker release foundation

christian-calabrese:
This PR introduces the technical foundation for Docker releases with Nx, without yet adding support for individual projects or updating the documentation.

It also adds the nx-docker-release-tools package, which enables the creation of an additional latest Docker image tag alongside the versioned tags.

Resolves CES-2152


packages/nx-docker-release-tools/package.json
```diff
@@ -0,0 +1,49 @@
...
+  "files": [
+    "dist"
+  ],
+  "bin": {
+    "dx-docker-release-publish-with-latest": "./dist/publish-with-latest.js"
+  },
+  "scripts": {
+    "build": "tsup",
+    "dev": "tsup --watch",
+    "typecheck": "tsc --noEmit",
```
gunzip:
```suggestion
```

inferred by nx

packages/nx-docker-release-tools/package.json
```diff
@@ -0,0 +1,49 @@
...
+    "test": "vitest run",
+    "test:coverage": "vitest run --coverage",
+    "lint": "eslint --fix src",
+    "lint:check": "eslint src",
```
gunzip: inferred by nx

packages/nx-docker-release-tools/README.md
```diff
@@ -0,0 +1,21 @@
+# @pagopa/nx-docker-release-tools
+
+Reusable CLI utilities for Docker publish steps in Nx release flows.
+
+## CLI
+
+### dx-docker-release-publish-with-latest
+
+Reads `tmp/<projectRoot>/.docker-version`, pushes the version tag, tags the same image as `latest`, and pushes `latest`.
```
gunzip: question: why `tmp`?

.dockerignore
```diff
@@ -4,6 +4,12 @@
 
 # node
 node_modules
+**/node_modules
```
gunzip: There are cases where we want to keep `node_modules` within docker images

nx.json
```diff
@@ -8,6 +8,16 @@
...
+    "nx-release-publish": {
+      "options": {
+        "command": "npx --yes --package=file:./packages/nx-docker-release-tools dx-docker-release-publish-with-latest --project-root {projectRoot}"
+      }
+    },
+    "typecheck": {
+      "dependsOn": ["^build"],
+      "inputs": ["{projectRoot}/src/**/*.ts"],
+      "cache": true
+    },
```
gunzip:
```suggestion
```

packages/nx-docker-release-tools/src/publish-with-latest.ts
```diff
@@ -0,0 +1,231 @@
+/**
+ * Docker publish utility for Nx release flows.
+ *
+ * The CLI reads Nx generated `.docker-version` metadata, pushes the versioned
```
gunzip: isn't `.docker-version` commonly used to pin Docker version?

packages/nx-docker-release-tools/README.md
christian-calabrese: It's how the docker plugin for nx works ([source 1](https://github.com/nrwl/nx/blob/71bbc709f475b263ff2c3a9056bfb63de2a06963/packages/docker/src/release/version-utils.ts#L18), [source 2](https://github.com/nrwl/nx/blob/71bbc709f475b263ff2c3a9056bfb63de2a06963/packages/docker/src/release/version-utils.ts#L147-L149))

packages/nx-docker-release-tools/src/publish-with-latest.ts
christian-calabrese: The docker plugin for nx writes the name of the image to be pushed, for example: `ghcr.io/pagopa/dx-slc-dockerapp2:0.1.25`

packages/nx-docker-release-tools/eslint.config.js
```diff
@@ -0,0 +1,3 @@
+import lintRules from "@pagopa/eslint-config";
+
+export default [...lintRules];
```
kin0992: **suggestion**:

```suggestion
export default lintRules;
```

packages/nx-docker-release-tools/package.json
```diff
@@ -0,0 +1,46 @@
+{
+  "name": "@pagopa/nx-docker-release-tools",
+  "version": "0.1.0",
```
kin0992: **suggestion**: start with version `0.0.0`, then nx will handle the semver with version plan


```suggestion
  "version": "0.0.0",
```

packages/nx-docker-release-tools/package.json
```diff
@@ -0,0 +1,46 @@
...
+    "cli"
+  ],
+  "files": [
+    "dist"
+  ],
+  "bin": {
+    "dx-docker-release-publish-with-latest": "./dist/publish-with-latest.js"
+  },
+  "scripts": {
+    "build": "tsup",
```
kin0992: **question**: aren't we dropping `tsup` and move to `tsdown`?
cc @lucacavallaro

.github/workflows/release-v2.yaml
```diff
@@ -76,6 +77,13 @@ jobs:
         shell: bash
         run: npm ci
 
+      - name: Container registry Login
+        uses: docker/login-action@74a5d142397b4f367a81961eba4e8cd7edddf772 # v3.4.0
```
kin0992: **question**: is there any reason that prevent us to use v4 of this action?

packages/nx-docker-release-tools/tsup.config.ts
kin0992, line 1: **note**: if we move to `tsdown` then also this file needs some changes

kin0992 (changes requested):
please update pr title to follow the conventions

packages/nx-docker-release-tools/README.md
```diff
@@ -0,0 +1,21 @@
+# @pagopa/nx-docker-release-tools
```
lucacavallaro, line 1: **question:** what is the purpose of this package? how should be used during a release? how is integrated with nx?

packages/nx-docker-release-tools/tsup.config.ts
```diff
@@ -0,0 +1,20 @@
+import { defineConfig } from "tsup";
```
lucacavallaro: **issue:** `tsup` must not be used, since it is abandoned

if you need to bundle the code -> use `rolldown` instead (search for examples in this repo)

if you don't need to bundle the code -> just use `tsc` (by just removing the `build` script from the manifest and using the inferred build task from Nx)

nx.json
```diff
@@ -8,6 +8,11 @@
       "outputs": ["{projectRoot}/dist/**", "{projectRoot}/bin/**"],
       "cache": true
     },
+    "nx-release-publish": {
```
lucacavallaro: **issue**: `nx-release-publish` is Nx task name used for publishing and must not be overridden by global config like this.

`nx-release-publish` in this repo is currently implemented for:
- publishable js code (`tag:npm:public`), like DX CLI and savemoney
- publishable terraform module (`tag:terraform:public`)

overriding the command field for these task can break other release flow

nx.json
```diff
@@ -8,6 +8,11 @@
...
+      "options": {
+        "command": "npx --yes --package=file:./packages/nx-docker-release-tools dx-docker-release-publish-with-latest --project-root {projectRoot}"
```
lucacavallaro: **issue:** avoid using `npx` for running internal commands

I recommend two alternatives

A) **keeping a dedicated package for the script**: add the package as dependency, export the script as `bin` and then call `pnpm script-name`

B) **removing the package, keeping only the script**: use `node /path/to/script.ts`

nx.json
```diff
@@ -98,10 +109,25 @@
...
+            "--label org.opencontainers.image.url=$(repo=${GITHUB_REPOSITORY:-}; repo=${repo:-$(git config --get remote.origin.url 2>/dev/null || git remote get-url origin 2>/dev/null || true)}; repo=${repo:-$(jq -r 'if (.repository|type)==\"string\" then .repository elif (.repository|type)==\"object\" then .repository.url else empty end // empty' package.json 2>/dev/null)}; repo=$(printf '%s' \"$repo\" | sed -E 's#^git\\+##; s#\\.git$##; s#^https://github.com/##; s#^git@github.com:##; s#^ssh://git@github.com/##; s#^git://github.com/##'); printf '%s' \"https://github.com/$repo\")",
+            "--label org.opencontainers.image.revision={commitSha}",
```
lucacavallaro: **question:** isn't there a way to centralize this config? 

in my opinion we can't ask our users to copy this snippet in their project

it is hard to understand and impossible to maintain

lucacavallaro (changes requested):

packages/nx-docker-release-tools/tsup.config.ts
lucacavallaro: before moving to tsdown, we should make sure if bundling is actually needed

nx.json
christian-calabrese: The DX CLI would configure this and it's centralized in the monorepo NX config

.github/workflows/release-v2.yaml
```diff
@@ -76,6 +77,13 @@ jobs:
...
+        with:
+          registry: ghcr.io
```
krusty93: question: is ghcr.io only enough?

packages/nx-docker-release-tools/src/publish-with-latest.ts
krusty93: Never seen this file, I cannot find it on documentation neither

krusty93: Perhaps must be added to .gitignore?

packages/nx-docker-release-tools/README.md
```diff
@@ -0,0 +1,21 @@
...
+npx --yes --package=@pagopa/nx-docker-release-tools dx-docker-release-publish-with-latest --project-root apps/my-docker-app
+```
+
+Supported args:
+
+- `--project-root <path>` (required)
+
+Environment behavior:
+
+- `NX_DRY_RUN=true` prints intended commands and exits without pushing.
```
krusty93: nitpick: DRY_RUN does not make me think to push operation, I see them as different things. DRY_RUN makes me think about input and environment validation

.dockerignore
krusty93: Such as?

nx.json
krusty93: Mmmh I'd suggest to look for other solutions. Can these be set only in CD workflows?

packages/nx-docker-release-tools/README.md
krusty93: The purpose is not clear to me neither

nx.json
lucacavallaro: maybe we can (xor):
- write a new Nx plugin that writes this config
- introduce a lint rule to the dockerfile linter that enforce the presence of LABEL in OCI images

even if this configuration is enforced by the DX CLI it is hard to understand, due to their inline scripts

packages/nx-docker-release-tools/README.md
christian-calabrese: `If your release process does not require latest, this package is not needed.`

In practice, the plugin does not yet support creating the latest tag. In some cases, this is still required for our use cases (for example, the GitHub Runner image). This package fills that gap by automatically creating and managing the additional latest tag alongside the versioned image tags.

packages/nx-docker-release-tools/package.json
christian-calabrese: I substituted it with tsc

packages/nx-docker-release-tools/README.md
christian-calabrese: It is inherited from NX

packages/nx-docker-release-tools/package.json
lucacavallaro: @christian-calabrese `tsc` is already configured via Nx, so it should not be specified in a custom build script

instead, remove the `build` command form this file and rename `tsconfig.json` to `tsconfig.lib.json`

in this way (after a `nx reset`), nx will infer the `build` target for this project (with the global, optimized, configuration)

packages/nx-docker-release-tools/package.json
```diff
@@ -0,0 +1,45 @@
...
+  "bin": {
+    "dx-docker-release-publish-with-latest": "./dist/publish-with-latest.js"
+  },
+  "scripts": {
+    "build": "tsc -p tsconfig.json",
+    "dev": "tsc -w -p tsconfig.json",
+    "test": "vitest run",
+    "test:coverage": "vitest run --coverage",
+    "format": "prettier --write .",
+    "format:check": "prettier --check ."
```
lucacavallaro: suggestion: remove these scripts, they are already inferred by `Nx`

packages/nx-docker-release-tools/README.md
krusty93: my question is: why do we need a package to set the tag `latest`? what's wrong with `docker build -t <image>:latest`, or equivalent?

PR #1869, opened 2026-06-19
Fix `build-workspace` action

lucacavallaro:
On `Nx` branch, `get-workspace-path` returned an absolute path instead
of a relative one.

With this change, it is now normalized to always be relative.


gunzip:
this breaks CI

gunzip (approved):

PR #1874, opened 2026-06-22
Add tests for renew TLS certificate action

mamu0:
This PR adds integration and end-to-end (E2E) tests for the `renew-tls-certificate` GitHub Action.
These E2E tests will trigger automatically whenever a pull request modifying the action is opened or synchronized.

Resolves: CES-2174


.github/workflows/_validate-renew-tls-certificate-e2e.yaml
```diff
@@ -0,0 +1,111 @@
...
+      ACME_DIRECTORY_URL: https://acme-staging-v02.api.letsencrypt.org/directory
+
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
+
+      - name: Cloud Provider Login
+        uses: pagopa/dx/actions/csp-login@main
+
+      - name: Create temporary Key Vault
```
gunzip: question: why key vault handling lies outside `test-e2e.sh`?

actions/renew-tls-certificate/le_create_acme_test_account.py
gunzip, line 1: suggestion: rename this file to something like `le_create_acme_test_account.py` since it's used only for testing or put everything test related within a new directory` __tests__`

actions/renew-tls-certificate/test-e2e.sh
```diff
@@ -0,0 +1,91 @@
+#!/usr/bin/env bash
```
gunzip: suggestion: add a module comment that share what this file does

actions/renew-tls-certificate/test-smoke.sh
```diff
@@ -0,0 +1,74 @@
+#!/usr/bin/env bash
```
gunzip, line 1: suggestion: add a module comment that share what this file does

actions/renew-tls-certificate/test.sh
```diff
@@ -0,0 +1,74 @@
...
+  cd "$WORK_DIR"
+
+  "$VENV_DIR/bin/python" "$ACTION_DIR/le_renew_ssl_certificate_generate_csr.py" \
+    --common-name test.dx.pagopa.it \
+    --out csr.der \
+    --keyout csr.key \
+    --rsa-key-size 2048 \
+    --quiet
+
+  "$VENV_DIR/bin/python" - <<'PY'
```
gunzip: suggestion: put these python scripts outside bash script so we can format, lint and test them individually

actions/renew-tls-certificate/action.yaml
```diff
@@ -36,6 +36,11 @@ inputs:
   lets_encrypt_registration_json:
     description: "Let's Encrypt account registration info in JSON format"
     required: true
+  lets_encrypt_directory_url:
```
krusty93: issue: tests shouldn't alter the public contract

.github/workflows/_validate-renew-tls-certificate-e2e.yaml
mamu0: It was only temporary to verify a specific step and has already been cleaned up

actions/renew-tls-certificate/action.yaml
mamu0: It was only temporary to verify a specific step and has already been cleaned up

actions/renew-tls-certificate/test.sh
```diff
@@ -0,0 +1,58 @@
...
+    --common-name test.dx.pagopa.it \
+    --out csr.der \
+    --keyout csr.key \
+    --rsa-key-size 2048 \
+    --quiet
+
+  "$VENV_DIR/bin/python" "$ACTION_DIR/test_validate_csr.py" csr.der test.dx.pagopa.it
+)
+
+PYTHONPATH="$ACTION_DIR" "$VENV_DIR/bin/python" - <<'PY'
```
gunzip: suggestion: extract this python code in it's own module.py

actions/renew-tls-certificate/package.json
```diff
@@ -14,6 +14,29 @@
     "azure",
     "key-vault"
   ],
-  "scripts": {},
+  "scripts": {
+    "test": "bash test.sh",
```
krusty93: suggestion: 
```suggestion
    "test:integration": "bash test.sh",
```

actions/renew-tls-certificate/test.sh
mamu0: done [c74d72d](https://github.com/pagopa/dx/pull/1874/commits/c74d72df3925d7e8feef3df3ae60e37d2aa8b9ad)

actions/renew-tls-certificate/le_create_acme_test_account.py
```diff
@@ -0,0 +1,84 @@
...
+import base64
+import json
+import urllib.request
+
+import cryptography.hazmat.primitives.hashes
+import cryptography.hazmat.primitives.serialization
+import cryptography.hazmat.primitives.asymmetric.padding
+import jwcrypto.jwk
+
+DEFAULT_DIRECTORY_URL = "https://acme-v02.api.letsencrypt.org/directory"
```
krusty93: question: shouldn't be staging?

actions/renew-tls-certificate/README.md
```diff
@@ -6,18 +6,19 @@ Both the check and the renewal are executed in a **single job**, keeping the cal
...
+| `arm_client_id`                  | Yes      | —                            | Azure Client ID (Workload Identity / Service Principal)                                     |
+| `arm_subscription_id`            | Yes      | —                            | Azure Subscription ID                                                                       |
+| `arm_tenant_id`                  | Yes      | —                            | Azure Tenant ID                                                                             |
+| `lets_encrypt_private_key_json`  | Yes      | —                            | Let's Encrypt account private key in JWK JSON format                                        |
+| `lets_encrypt_registration_json` | Yes      | —                            | Let's Encrypt account registration info in JSON format                                      |
+| `lets_encrypt_directory_url`     | No       | Production ACME v2 directory | ACME directory URL used to request the certificate                                          |
```
krusty93: Not needed anymore

.github/workflows/_validate-renew-tls-certificate-e2e.yaml
```diff
@@ -0,0 +1,59 @@
...
+
+      - name: Cache Nx
+        uses: actions/cache@668228422ae6a00e4ad889ee87cd7109ec5666a7 # v5.0.4
+        with:
+          path: .nx
+          key: ${{ runner.os }}-nx-cache-${{ hashFiles('**/pnpm-lock.yaml') }}
+          restore-keys: |
+            ${{ runner.os }}-nx-cache-
+
+      - name: Run renew TLS certificate E2E test
```
krusty93: question: why do we need this script instead of using the workflow template?

mamu0: It seemed like the simplest and most effective way to test the PR's code without pointing directly to the local action

actions/renew-tls-certificate/README.md
mamu0: done [0050f65](https://github.com/pagopa/dx/pull/1874/commits/0050f65ef5aac764b5edb8208660ed71554ee13b)

actions/renew-tls-certificate/le_create_acme_test_account.py
mamu0: changed here [0050f65](https://github.com/pagopa/dx/pull/1874/commits/0050f65ef5aac764b5edb8208660ed71554ee13b)

actions/renew-tls-certificate/package.json
```diff
@@ -14,6 +14,30 @@
...
+        "cache": false,
+        "dependsOn": [],
+        "inputs": [
+          "{projectRoot}/action.yaml",
+          "{projectRoot}/le_*.py",
+          "{projectRoot}/test_smoke_imports.py",
+          "{projectRoot}/test_validate_csr.py",
+          "{projectRoot}/le_renew_ssl_certificate_python_requirements.txt",
+          "{projectRoot}/test-e2e.sh"
+        ],
```
gunzip:
```suggestion
```

We can just rely on NX defaults which is 'all project files'

actions/renew-tls-certificate/package.json
```diff
@@ -14,6 +14,30 @@
...
     "key-vault"
   ],
-  "scripts": {},
+  "scripts": {
+    "test:integration": "bash test.sh",
+    "test:e2e": "bash test-e2e.sh"
+  },
+  "nx": {
+    "targets": {
+      "test:e2e": {
```
gunzip: question: why do we need this nx task? can't we just rely on the npm tasks above?

.github/workflows/_validate-renew-tls-certificate-e2e.yaml
gunzip: I've reviewed though copilot, and I agree that it would be better to reuse action code if possible.

`test-e2e.sh` manually creates the ACME account, CSR, cert, PFX, and Key Vault import, duplicating `action.yaml`. That makes the test harder to read and less useful: changes to the composite action wiring could break while this E2E still passes.

Suggested simplification:

1. Add an optional `lets_encrypt_directory_url` input to `action.yaml`, defaulting to production (it's ok for me to introduce this parameter for testing purpose)
2. In `.github/workflows/_validate-renew-tls-certificate-e2e.yaml`, create the staging ACME account, then call the local action directly with `uses: ./actions/renew-tls-certificate`.
3. Use a per-run CN like `renew-tls-test-${{ github.run_id }}-${{ github.run_attempt }}.dev.dx.pagopa.it` so the Key Vault cert name is naturally unique and the suffix logic in `test-e2e.sh:18-23` disappears.

This would cut most of the bespoke shell orchestration and turn the E2E into a real test of the GitHub Action surface.

mamu0: That is exactly how I had it originally, but adding that input was considered an issue. If we are okay with adding that parameter to the action, I can revert the changes and point to the local action instead.

gunzip: Spot on, we're ok with that. Using 'production' as the default keeps the UX intact and ensures a non-breaking change. Weighing the trade-off: it's better to ensure a solid test than to add complexity, all for a minimal user impact.

.github/workflows/_validate-renew-tls-certificate-e2e.yaml
```diff
@@ -0,0 +1,59 @@
...
+env:
+  ARM_TENANT_ID: ${{ secrets.ARM_TENANT_ID }}
+  ARM_USE_OIDC: true
+  ARM_USE_AZUREAD: true
+
+jobs:
+  renew-tls-certificate-e2e-test:
+    name: Run renew TLS certificate E2E test
+    runs-on: ubuntu-latest
+    environment: automation-dev-cd
```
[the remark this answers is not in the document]
krusty93: Shall we add the fork gate instead?

.github/workflows/_validate-renew-tls-certificate-e2e.yaml
krusty93: Can we dynamically set the URL according to the current context instead of adding a new user-POV-useless parameter?

e.g. the reusable workflow checks for an env var set by the caller..

actions/renew-tls-certificate/test-smoke.sh
krusty93, line 1: thought: perhaps the file should be named `test-smoke.sh`?

.github/workflows/_validate-renew-tls-certificate-e2e.yaml
mamu0: Yes, that can be done.
However, we should keep in mind that users are more likely to adopt the reusable workflow `release-certificate-v1.yaml` rather than the action itself.
Adding an extra input to the action would simply make it more customizable, but PagoPA users will continue to utilize the workflow, which won't expose this parameter.

actions/renew-tls-certificate/test-smoke.sh
mamu0: done here [c7b8dca](https://github.com/pagopa/dx/pull/1874/commits/c7b8dca58020cc27abe4d815c714aff4b6e807a2)

.github/workflows/_validate-renew-tls-certificate-e2e.yaml
```diff
@@ -0,0 +1,59 @@
...
+        uses: actions/cache@668228422ae6a00e4ad889ee87cd7109ec5666a7 # v5.0.4
+        with:
+          path: .nx
+          key: ${{ runner.os }}-nx-cache-${{ hashFiles('**/pnpm-lock.yaml') }}
+          restore-keys: |
+            ${{ runner.os }}-nx-cache-
+
+      - name: Run renew TLS certificate E2E test
+        timeout-minutes: 20
+        run: pnpm nx run renew-tls-certificate:test:e2e
```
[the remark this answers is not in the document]
mamu0: done [c7b8dca](https://github.com/pagopa/dx/pull/1874/commits/c7b8dca58020cc27abe4d815c714aff4b6e807a2)

actions/renew-tls-certificate/package.json
mamu0: removed here [c7b8dca](https://github.com/pagopa/dx/pull/1874/commits/c7b8dca58020cc27abe4d815c714aff4b6e807a2)

actions/renew-tls-certificate/package.json
mamu0: resolved here [c7b8dca](https://github.com/pagopa/dx/pull/1874/commits/c7b8dca58020cc27abe4d815c714aff4b6e807a2)

.github/workflows/_validate-renew-tls-certificate-e2e.yaml
mamu0: done here [c7b8dca](https://github.com/pagopa/dx/pull/1874/commits/c7b8dca58020cc27abe4d815c714aff4b6e807a2)

actions/renew-tls-certificate/le_renew_ssl_certificate_acme_tiny.py
```diff
@@ -19,8 +19,14 @@
 import jwcrypto.jwk
 
 DEFAULT_DIRECTORY_URL = "https://acme-v02.api.letsencrypt.org/directory"
+DIRECTORY_URL_ENV_VAR = "LETS_ENCRYPT_DIRECTORY_URL"
 DEFAULT_DNS_TTL_SEC = 30
 
+
+def get_default_directory_url():
+    return os.getenv(DIRECTORY_URL_ENV_VAR, DEFAULT_DIRECTORY_URL)
```
krusty93, line 27: question: What does happen if I set `DEFAULT_DIRECTORY_URL`, leaving `DIRECTORY_URL_ENV_VAR` unset?

mamu0: I'm not sure I entirely follow your question, but `DEFAULT_DIRECTORY_URL` is the existing constant that always holds the value `https://acme-v02.api.letsencrypt.org/directory`, whereas `DIRECTORY_URL_ENV_VAR` contains the value passed from the `LETS_ENCRYPT_DIRECTORY_URL` environment variable.

In this line, it fetches the environment variable, and if it is not found, it falls back to the default constant.

krusty93: so, `DEFAULT_DIRECTORY_URL` is not an environment variable that can be overridden by the caller, doesn't it?

mamu0: Exactly, that is the fallback value it defaults to

krusty93 (approved):

PR #1875, opened 2026-06-23
Add DX CLI init skill

kin0992:
## Summary
- add a repo-local `dx-cli` skill under `.github/skills/`
- make the skill start from the live `dx spec` output and scope it to `dx init`
- document that agents must invoke the CLI directly instead of running separate precondition checks, including the current `--publish` limitation for fully prompt-free runs

## Notes
- the skill includes a dedicated `init` workflow reference plus an Apache 2.0 `LICENSE.txt`
- `pnpm nx release plan:check` reports that these changes do not touch released projects, so no new version plan file is required


kin0992:
replaced by #1898

PR #1877, opened 2026-06-23
Align Azure authorization with repo owner

kin0992:
When a repository is initialized under an organization other than `pagopa`, `dx add environment` still opens the Azure authorization pull request against `pagopa/eng-azure-authorization`. This breaks the bootstrap flow for repositories created under `pagopa-dx` or any other supported owner.

This change threads the selected GitHub owner through the authorization request and uses it to target `<owner>/eng-azure-authorization` while keeping the authorization repository name fixed. It also extends the CLI tests to cover non-default owners and adds the required patch version plan for `@pagopa/dx-cli`.

## What changed

- add `repoOwner` to the authorization input contract and pass it from `dx add environment`
- update the Azure authorization adapter to use the propagated owner for file reads, branch creation, file updates, and pull request creation
- add regression coverage for owner-aware command, use-case, and adapter behavior
- add the CLI patch release plan

### DIFF
--- a/.nx/version-plans/version-plan-1782204343110.md
+++ b/.nx/version-plans/version-plan-1782204343110.md
@@ -1,0 +1,5 @@
+---
+'@pagopa/dx-cli': patch
+---
+
+Make Azure authorization follow the selected GitHub owner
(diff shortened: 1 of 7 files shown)


gunzip (approved):
lgtm, let's see what copilot says

PR #1879, opened 2026-06-23
Add support for IP masking in Application Insights configuration

krusty93:
Introduce a feature to enable IP masking in Application Insights telemetry data. Update the Azure module version and add necessary variables to support this functionality. Adjust the Application Insights resource configuration to utilize the new masking option.

Close CES-2164

### DIFF
--- a/infra/modules/azure_core_infra/_modules/application_insights/appi.tf
+++ b/infra/modules/azure_core_infra/_modules/application_insights/appi.tf
@@ -7,7 +7,7 @@ resource "azurerm_application_insights" "appi" {
   }))
   location             = var.location
   resource_group_name  = var.resource_group_name
-  disable_ip_masking   = var.mask_ip_address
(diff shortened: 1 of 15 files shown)


infra/modules/azure_core_infra/_modules/application_insights/appi.tf
```diff
@@ -7,7 +7,7 @@ resource "azurerm_application_insights" "appi" {
   }))
   location             = var.location
   resource_group_name  = var.resource_group_name
-  disable_ip_masking   = true
+  disable_ip_masking   = var.mask_ip_address
```
christian-calabrese: [issue](https://pullpo.io/cc?l=issue) 
I think we should negate the variable due to the naming

```suggestion
  disable_ip_masking   = !var.mask_ip_address
```

christian-calabrese (approved):

PR #1883, opened 2026-06-23
Support ESLint 9 alongside ESLint 10

kin0992:
Widen the eslint and @eslint/js peer ranges to ^9.0.0 || ^10.0.0 so
React Native applications (which are pinned to ESLint 9) can use this
shared config. @eslint/js moves from a regular dependency to a peer
dependency so the consumer's ESLint major dictates the matching
@eslint/js major.

Refs CES-1940

Co-authored-by: Copilot <223556219+Copilot@users.noreply.github.com>

### DIFF
--- a/packages/eslint-config/package.json
+++ b/packages/eslint-config/package.json
@@ -16,7 +16,6 @@
     "index.js"
   ],
   "dependencies": {
-    "@eslint/js": "^10.0.1",
     "@vitest/eslint-plugin": "^1.6.13",
     "eslint-config-prettier": "^10.1.5",
     "eslint-plugin-perfectionist": "^5.7.0",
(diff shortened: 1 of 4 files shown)


packages/eslint-config/package.json
```diff
@@ -25,11 +24,13 @@
...
   },
   "devDependencies": {
+    "@eslint/js": "^10.0.1",
     "eslint": "catalog:",
     "prettier": "catalog:"
   },
   "peerDependencies": {
-    "eslint": "^10.0.0",
+    "@eslint/js": "^9.0.0 || ^10.0.0",
+    "eslint": "^9.0.0 || ^10.0.0",
```
lucacavallaro, line 33: question: why `||` and not `>=9.0.0`

packages/eslint-config/package.json
```diff
@@ -25,11 +24,13 @@
     "typescript-eslint": "^8.59.1"
   },
   "devDependencies": {
+    "@eslint/js": "^10.0.1",
```
lucacavallaro, line 27: question: why was this dep moved to `devDependencies`? it is used at runtime

packages/eslint-config/package.json
kin0992: I decided to use the `||` instead of the `>=` operator to make sure the package is compatible with just version 9 and 10 (when version 11 is out, the DX package will not pop a warning/error - I'm not sure which one, honestly - during the install/update).

It's a way to explicitly define which versions are supported by the package.

packages/eslint-config/package.json
lucacavallaro: nvm, since it is a peerdep we don't need it anymore in `dependencies`

lucacavallaro (approved):

PR #1884, opened 2026-06-23
Add /jest subpath export to @pagopa/eslint-config

kin0992:
Extract the shared non-test configuration into base.js and add a new
jest.js entry point so React Native applications (and any other
consumer using Jest) can opt into the equivalent recommended-style
rules without the Vitest plugin's false positives. The default import
keeps shipping the Vitest preset, so existing consumers do not need
to migrate.

eslint-plugin-jest is declared as an optional peer dependency so it
is only installed by consumers that actually opt into the new
subpath.

An explicit subpath was preferred over auto-detecting the test runner
at config-load time. Detection would have to inspect the consumer's
installed packages or package.json while the flat config is being
built, which is non-deterministic: a project may legitimately have
both Vitest and Jest present (during a migration, or across different
projects in a monorepo), leaving the choice ambiguous, and the
resulting lint behaviour would silently change with the dependency
tree. A subpath keeps the selection explicit and reproducible — each
eslint.config.js states exactly which preset it wants — and lets the
matching plugin be a clean opt-in peer dependency instead of a
conditional runtime require.

Refs CES-1939, CES-1942

Co-authored-by: Copilot <223556219+Copilot@users.noreply.github.com>

### DIFF
--- a/.nx/version-plans/version-plan-1782213453124.md
+++ b/.nx/version-plans/version-plan-1782213453124.md
@@ -1,0 +1,5 @@
+---
+'@pagopa/eslint-config': minor
+---
+
+Support ESLint 9 alongside ESLint 10
--- a/.nx/version-plans/version-plan-1782215083050.md
+++ b/.nx/version-plans/version-plan-1782215083050.md
@@ -1,0 +1,5 @@
+---
+'@pagopa/eslint-config': minor
+---
+
+Add @pagopa/eslint-config/jest subpath for Jest / React Native consumers
--- a/packages/eslint-config/README.md
+++ b/packages/eslint-config/README.md
@@ -45,3 +45,31 @@ It supports both **ESLint 9** and **ESLint 10**.
      }
    }
    ```
+
+## Test-runner support
+
+The default entry point ships rules for [Vitest](https://vitest.dev/). A separate
+`@pagopa/eslint-config/jest` subpath provides the equivalent rules for
+[Jest](https://jestjs.io/), which is the default test runner for React Native apps.
+
+### Jest (React Native)
+
+Install `eslint-plugin-jest` alongside the regular peer dependencies:
+
+```shell
+pnpm add -D eslint-plugin-jest
+```
+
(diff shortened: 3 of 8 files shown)


dmnplb:
Hi @kin0992,
just to be sure, by analysing what has already been done in [this PR](https://github.com/pagopa/io-app/pull/7992), we should replace the following import:
```js
// Before
import pagopaConfig from "@pagopa/eslint-config";
// After
import pagopaConfig from "@pagopa/eslint-config/jest";
```
Then remove the `vitest` filter and leave the rest untouched. Is that correct?

kin0992:
> Hi @kin0992, just to be sure, by analysing what has already been done in [this PR](https://github.com/pagopa/io-app/pull/7992), we should replace the following import:
> 
> ```js
> // Before
> import pagopaConfig from "@pagopa/eslint-config";
> // After
> import pagopaConfig from "@pagopa/eslint-config/jest";
> ```
> 
> Then remove the `vitest` filter and leave the rest untouched. Is that correct?

Hi @dmnplb 😄
yes the idea is that importing `import pagopaConfig from "@pagopa/eslint-config";` you would use vitest as it is the tool in the adopt ring in the tech radar.
But since there are requirements to use jest too (as you previously pointed out), by importing `import pagopaConfig from "@pagopa/eslint-config/jest";` you would load the whole Typescript configuration enabled also by the "default" import, but the vitest rules are replaced by the Jest rules.

Then, you can extend your config with other rules too, like the React Native ones.

In case you have some feedback on the Jest rules, we are all ears 🙏

dmnplb:
Thanks for the explanation.

> Then, you can extend your config with other rules too, like the React Native ones.
> In case you have some feedback on the Jest rules, we are all ears 🙏

I don't have any specific feedback. We just need to run the `jest` rules against the codebase to see if the errors can be easily fixed. It would be better if we could find a way to test them before the official release, to avoid any back-and-forth

gunzip (approved):

PR #1895, opened 2026-06-24
Enhance Trivy summary reporting in static analysis workflow

krusty93:
Improve the reporting of Trivy findings in the static analysis workflow by summarizing results in a dedicated markdown file and posting them as comments on pull requests when issues are detected. This change enhances visibility and ensures that necessary findings are addressed before enforcement.

Close CES-2159

### DIFF
--- a/.github/workflows/static_analysis.yaml
+++ b/.github/workflows/static_analysis.yaml
@@ -339,34 +339,6 @@ jobs:
             add_to_summary ""
           fi
 
-          # Trivy custom policies output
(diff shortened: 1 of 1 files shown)


gunzip:
can you add a before and after screenshot?

krusty93:
> can you add a before and after screenshot?

Old
image

New
image

The difference than before is that the trivy report will not be posted in the same comment of the terraform locks warning; moreover, it now updates always the same comment instead of posting a new one at every run (like pre commit does)

gunzip (approved):

PR #1899, opened 2026-06-25
Remove `uat` environment

lucacavallaro:
This environment was never used


krusty93 (approved):
thought: we should create the command `remove environment`

PR #1902, opened 2026-06-26
Compare MCP server full-workspace Docker build

christian-calabrese:
Compare the MCP server Docker strategy that copies the whole monorepo and builds inside the image.

Summary
- enable the official @nx/docker target for apps/mcpserver with repo-root context
- keep the Dockerfile inside the app while forcing Docker to use the repo root as build context
- copy the whole workspace and run pnpm install + pnpm nx build inside Docker
- document the cache trade-off directly in the Dockerfile comments

Validation
- docker build produced apps-mcpserver:poc-a
- the final image contains the dist/cli.js entrypoint payload

Notes
- No version plan: PoC branch for Docker strategy evaluation, not intended for release
- Draft because this branch is meant to be compared with the prune-based alternative

### DIFF
--- a/apps/mcpserver/Dockerfile
+++ b/apps/mcpserver/Dockerfile
@@ -15,11 +15,8 @@ COPY . .
 #    the project from sources entirely inside the container.
 RUN pnpm install --filter @pagopa/dx-mcpserver...
 
-# 5. Run the normal Nx build inside Docker.
-RUN pnpm nx build @pagopa/dx-mcpserver
-
-# 6. Materialize a production-only payload through an Nx task so the same
-#    packaging step can be exercised outside Docker as well.
+# 5. Materialize the production-only payload through the Nx task pipeline.
+#    The package target depends on build, so Docker only needs a single Nx call.
 RUN pnpm nx package @pagopa/dx-mcpserver
 
 # --- Final Image ---
--- a/apps/mcpserver/Dockerfile
+++ b/apps/mcpserver/Dockerfile
@@ -1,32 +1,34 @@
 FROM public.ecr.aws/docker/library/node:24-alpine@sha256:2bdb65ed1dab192432bc31c95f94155ca5ad7fc1392fb7eb7526ab682fa5bf14 AS base
+# This PoC keeps the Dockerfile in charge of dependency installation and build
+# so we can measure the caching impact of copying the whole workspace.
 # 1. Enable pnpm
 RUN corepack enable
 
 # 2. Set working directory
 WORKDIR /app
 
-# 3. Copy monorepo configuration files
-COPY ./pnpm-lock.yaml ./pnpm-workspace.yaml ./package.json ./nx.json ./tsconfig.json ./
+# 3. Copy the full workspace on purpose: any touched package can invalidate
+#    the build cache, which is the behavior this PoC wants to expose.
+COPY . .
 
-# 4. Copy source code for apps and packages
-COPY ./apps ./apps
-COPY ./packages ./packages
-
-# 5. Install ALL dependencies for mcpserver and its workspace dependencies
+# 4. Install the app plus every workspace package it consumes so Nx can build
+#    the project from sources entirely inside the container.
 RUN pnpm install --filter @pagopa/dx-mcpserver...
-# 6. Build the mcpserver app
+
+# 5. Run the normal Nx build inside Docker.
 RUN pnpm nx build @pagopa/dx-mcpserver
-# 7. Prune development-only dependencies for the final image
+
+# 6. Materialize a production-only payload for the runtime stage.
 RUN pnpm --filter @pagopa/dx-mcpserver deploy --legacy --prod /app/deploy
 
 # --- Final Image ---
 FROM public.ecr.aws/docker/library/node:24-alpine@sha256:2bdb65ed1dab192432bc31c95f94155ca5ad7fc1392fb7eb7526ab682fa5bf14
 WORKDIR /app
 
-# Copy the AWS Lambda Adapter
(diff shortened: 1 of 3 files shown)


apps/mcpserver/Dockerfile
```diff
@@ -1,32 +1,34 @@
...
+#    the project from sources entirely inside the container.
 RUN pnpm install --filter @pagopa/dx-mcpserver...
-# 6. Build the mcpserver app
+
+# 5. Run the normal Nx build inside Docker.
 RUN pnpm nx build @pagopa/dx-mcpserver
-# 7. Prune development-only dependencies for the final image
+
+# 6. Materialize a production-only payload for the runtime stage.
 RUN pnpm --filter @pagopa/dx-mcpserver deploy --legacy --prod /app/deploy
```
gunzip: you can extract this task into a nx (npm in this case) task

`"package": "pnpm --filter @pagopa/dx-mcpserver deploy --legacy --prod /app/deploy"`

then call

`RUN nx package @pagopa/dx-mcpserver`

this lets you try the task without running docker and simplifies dockerfile

christian-calabrese:
Benchmark notes for the two Docker PoCs.

Environment
- Measured locally on macOS with Rancher Desktop.
- Base images were already available locally.
- These numbers are best read as relative comparisons between the two approaches, not as absolute CI timings.
- Two views are reported below:
  - Workspace-ready benchmark: the host workspace was already bootstrapped.
  - Clean-state / CI-like benchmark: a fresh temporary worktree with no host `node_modules`; for PoC B, `pnpm install --frozen-lockfile` ran against an empty local pnpm store (`--store-dir .bench-store`) before `nx prune`; Docker builds used `--no-cache`.

Scenario definitions
- Workspace-ready benchmark:
  - PoC A: Docker build only, because dependency install, project build and production packaging all happen inside the Dockerfile.
  - PoC B: Nx prune outside Docker, then Docker build from the pruned payload.
  - Warm rebuild: re-run the same flow without changing any input file.
  - Root-noise rebuild: re-run after touching an unrelated file at the repository root.
- Runtime validity: not just file presence. The image is checked by importing the compiled runtime module inside the container and verifying that it loads successfully.

Workspace-ready results

| Scenario | PoC A: full-workspace Docker build | PoC B: pruned Docker build |
| --- | ---: | ---: |
| Cold path | 78.850s | prune 9.190s + docker 15.540s = 24.730s |
| Warm rebuild | 0.430s | prune 0.820s + docker 2.680s = 3.500s |
| Root-noise rebuild | 87.255s | prune 1.290s + docker 1.820s = 3.110s |
| Runtime validity | PASS | PASS |

Clean-state / CI-like results

| Scenario | PoC A: full-workspace Docker build | PoC B: pruned Docker build |
| --- | ---: | ---: |
| Clean-state total | 110.670s | install 55.200s + prune 43.770s + docker 14.230s = 113.200s |
| Runtime validity | PASS | PASS |

How to read these numbers
- PoC A is still the simplest pipeline and still the fastest when absolutely nothing changes, because Docker can reuse every layer.
- PoC A still degrades badly when an unrelated root-level file changes, because `COPY . .` invalidates the build context and forces install/build/package work again.
- PoC B is now runtime-valid on the current branch.
- PoC B still has a much better cold path and a much better root-noise path once the workspace is already bootstrapped, because the Docker context is narrow and the Nx prune step isolates the app payload.
- The clean-state benchmark changes the picture: if we model a CI run starting from checkout only, PoC B must first pay for host bootstrap (`pnpm install --frozen-lockfile`) and for the prune/build step before Docker even starts.
- In this clean-state run, total end-to-end time is effectively a wash, with PoC B slightly slower: 113.200s vs 110.670s.
- PoC B currently relies on the temporary sanitize workaround for the current Nx gap around pnpm `catalog:` specifiers in pruned outputs.

Current conclusion
- If the target is the current CI shape in this repo, where the workflow does checkout and then Docker build, PoC A is still the cleaner fit and is slightly faster from a clean start.
- If the target is a pipeline that is willing to prepare or cache the workspace on the host in exchange for a narrower Docker context and better isolation from monorepo-wide changes, PoC B remains attractive.
- So PoC B is no longer blocked on runtime correctness, but its real payoff depends on whether the pipeline can amortize the extra host bootstrap it requires.

apps/mcpserver/package.json
```diff
@@ -41,6 +41,7 @@
   },
   "scripts": {
     "start": "tsx src/cli.ts",
+    "package": "pnpm --dir ../.. --filter @pagopa/dx-mcpserver deploy --legacy --prod apps/mcpserver/deploy",
```
lucacavallaro: **question:** is the --dir required?

apps/mcpserver/Dockerfile
```diff
@@ -1,33 +1,36 @@
...
-# 6. Build the mcpserver app
+
+# 5. Run the normal Nx build inside Docker.
 RUN pnpm nx build @pagopa/dx-mcpserver
-# 7. Prune development-only dependencies for the final image
-RUN pnpm --filter @pagopa/dx-mcpserver deploy --legacy --prod /app/deploy
+
+# 6. Materialize a production-only payload through an Nx task so the same
+#    packaging step can be exercised outside Docker as well.
+RUN pnpm nx package @pagopa/dx-mcpserver
```
lucacavallaro: **suggestion**: always prefer task pipelines over multiple `nx run` calls. Nx should be aware that `package` depends on `build`, users should not have to remember the order of operations

apps/mcpserver/package.json
lucacavallaro: **suggestion:** avoid calling `pnpm` from `pnpm`, this kind tasks can be implemented as `nx targets`

PR #1903, opened 2026-06-26
Compare MCP server pruned Docker build

christian-calabrese:
Compare the MCP server Docker strategy that consumes a pruned payload and installs runtime dependencies inside the image.

Summary
- enable the official @nx/docker target for apps/mcpserver with project-local context
- add Nx prune targets for lockfile pruning, workspace module copy, workspace config copy and manifest sanitization
- build the runtime image from dist only and keep the Docker context minimal with a dedicated .dockerignore
- document the pruning and multi-arch trade-offs directly in the Docker assets

Validation
- node ./scripts/sanitize-pruned-output.js
- docker build produced apps-mcpserver:poc-b
- the final image contains the dist/cli.js entrypoint payload

Notes
- No version plan: PoC branch for Docker strategy evaluation, not intended for release
- The install path still uses --no-frozen-lockfile because raw Nx prune output is not sufficient in this workspace
- Draft because this branch is meant to be compared with the full-workspace alternative


apps/mcpserver/Dockerfile
```diff
@@ -1,33 +1,26 @@
...
-RUN pnpm install --filter @pagopa/dx-mcpserver...
-# 6. Build the mcpserver app
-RUN pnpm nx build @pagopa/dx-mcpserver
-# 7. Prune development-only dependencies for the final image
-RUN pnpm --filter @pagopa/dx-mcpserver deploy --legacy --prod /app/deploy
+# Install from the pruned payload rather than from the monorepo root.
+# The explicit cleanup keeps the final image close to the runtime-only payload
+# instead of shipping the temporary pnpm store and pruning metadata.
+RUN pnpm install --prod --no-frozen-lockfile && \
+	pnpm store prune && \
```
gunzip, line 17: shound't we use `nx prune` instead of `pnpm prune` (which cleans up deps)?

christian-calabrese:
Benchmark notes for the two Docker PoCs.

Environment
- Measured locally on macOS with Rancher Desktop.
- Base images were already available locally.
- These numbers are best read as relative comparisons between the two approaches, not as absolute CI timings.
- Two views are reported below:
  - Workspace-ready benchmark: the host workspace was already bootstrapped.
  - Clean-state / CI-like benchmark: a fresh temporary worktree with no host `node_modules`; for PoC B, `pnpm install --frozen-lockfile` ran against an empty local pnpm store (`--store-dir .bench-store`) before `nx prune`; Docker builds used `--no-cache`.

Scenario definitions
- Workspace-ready benchmark:
  - PoC A: Docker build only, because dependency install, project build and production packaging all happen inside the Dockerfile.
  - PoC B: Nx prune outside Docker, then Docker build from the pruned payload.
  - Warm rebuild: re-run the same flow without changing any input file.
  - Root-noise rebuild: re-run after touching an unrelated file at the repository root.
- Runtime validity: not just file presence. The image is checked by importing the compiled runtime module inside the container and verifying that it loads successfully.

Workspace-ready results

| Scenario | PoC A: full-workspace Docker build | PoC B: pruned Docker build |
| --- | ---: | ---: |
| Cold path | 78.850s | prune 9.190s + docker 15.540s = 24.730s |
| Warm rebuild | 0.430s | prune 0.820s + docker 2.680s = 3.500s |
| Root-noise rebuild | 87.255s | prune 1.290s + docker 1.820s = 3.110s |
| Runtime validity | PASS | PASS |

Clean-state / CI-like results

| Scenario | PoC A: full-workspace Docker build | PoC B: pruned Docker build |
| --- | ---: | ---: |
| Clean-state total | 110.670s | install 55.200s + prune 43.770s + docker 14.230s = 113.200s |
| Runtime validity | PASS | PASS |

How to read these numbers
- PoC A is still the simplest pipeline and still the fastest when absolutely nothing changes, because Docker can reuse every layer.
- PoC A still degrades badly when an unrelated root-level file changes, because `COPY . .` invalidates the build context and forces install/build/package work again.
- PoC B is now runtime-valid on the current branch.
- PoC B still has a much better cold path and a much better root-noise path once the workspace is already bootstrapped, because the Docker context is narrow and the Nx prune step isolates the app payload.
- The clean-state benchmark changes the picture: if we model a CI run starting from checkout only, PoC B must first pay for host bootstrap (`pnpm install --frozen-lockfile`) and for the prune/build step before Docker even starts.
- In this clean-state run, total end-to-end time is effectively a wash, with PoC B slightly slower: 113.200s vs 110.670s.
- PoC B currently relies on the temporary sanitize workaround for the current Nx gap around pnpm `catalog:` specifiers in pruned outputs.

Current conclusion
- If the target is the current CI shape in this repo, where the workflow does checkout and then Docker build, PoC A is still the cleaner fit and is slightly faster from a clean start.
- If the target is a pipeline that is willing to prepare or cache the workspace on the host in exchange for a narrower Docker context and better isolation from monorepo-wide changes, PoC B remains attractive.
- So PoC B is no longer blocked on runtime correctness, but its real payoff depends on whether the pipeline can amortize the extra host bootstrap it requires.

christian-calabrese:
The sanitize script in this PoC is a temporary workaround for the current Nx gap around pnpm `catalog:` specifiers in pruned outputs. It should be removed as soon as the upstream fix lands. Tracking issue: https://github.com/nrwl/nx/issues/35419. Proposed upstream fix: https://github.com/nrwl/nx/pull/35805.

apps/mcpserver/Dockerfile
christian-calabrese: From what we decided, the actual prune is executed by nx outside of docker (you can see it in the project.json configuration).
This `pnpm store prune`, instead, cleans the pnpm local store cache to avoid including it in the final docker image.

nx.json
```diff
@@ -98,9 +98,20 @@
...
+      "exclude": [
+        "apps/dx-metrics/Dockerfile",
+        "apps/dx-metrics-import/Dockerfile"
+      ],
       "options": {
-        "buildTarget": "docker:build",
+        "buildTarget": {
+          "name": "docker:build",
+          "args": [
+            "--platform linux/arm64"
```
krusty93, line 109:
```suggestion
            "--platform linux/arm64,linux/amd64"
```

PR #1905, opened 2026-06-26
Add Terraform Plan in CI for DX CLI Bootstrap workflow

mamu0:
This PR updates the bootstrap workflow template generated by the dx CLI to support bootstrap updates within a pull request.

Now, whenever the bootstrap configuration is modified in a PR, the workflow triggers in CI and posts the terraform plan output as a comment.

This relies on infra_plan.yaml. To ensure proper least-privilege access, a new CI environment for the bootstrap has been introduced, along with its dedicated managed identity.

Resolves: CES-2178


gunzip:
it looks like CI fails and changes have a conflict

apps/cli/src/adapters/azure/cloud-account-service.ts
```diff
@@ -263,30 +281,27 @@ export class AzureCloudAccountService implements CloudAccountService {
...
-      );
+        resourceGroupName,
+      });
+      const ciIdentity = await this.#createBootstrapperIdentity({
+        cloudAccountId: cloudAccount.id,
+        identityName: ciIdentityName,
+        msiClient,
+        parameters,
+        resourceGroupName,
+      });
```
kin0992: **note**: it looks like you can run this in pararllel, as you do later, using `Promise.all`

apps/cli/src/adapters/azure/cloud-account-service.ts
```diff
@@ -70,6 +71,23 @@ const bootstrapIdentityRoleDefinitionIds = [
...
 
+const bootstrapCiIdentityRoleDefinitionIds = [
+  // Plan needs read access for refresh and blob access for Terraform backend locks/state.
+  builtInRoleDefinitionIds.reader,
+  builtInRoleDefinitionIds.storageBlobDataContributor,
+] as const;
+
+type BootstrapperIdentity = {
+  clientId: string;
+  name: string;
```
kin0992: **thought**: in case the `name` has a defined format, we could use [TypeScript template literals](https://www.typescriptlang.org/docs/handbook/2/template-literal-types.html)

For example, if the BootstrapIdentity name follows the following pattern: `${prefix}-${env}-${location}-bootstrap-id-01`, we could change its type from string to `${string}-${string}-${string}-bootstrap-id-${string}`.
In this case, the string `hello` is not a valid bootstrap identity name (right now it would since the `name` property type is string.

Even better, if our domain defines the types for environment and location (like `type Environment = 'dev' | 'prod' | 'uat'` and `type Location = ...`, the identity name type could be `${string}-${Environment}-${Location}-bootstrap-id-${string}`.

The difference between `${string}-${string}-${string}-bootstrap-id-${string}` and `${string}-${Environment}-${Location}-bootstrap-id-${string}` is that at compile time you would have the following behaviours:

```typescript
type Type0 = `${string}-${string}-${string}-bootstrap-id-${string}`:
type Type1 = `${string}-${Environment}-${Location}-bootstrap-id-${string}`;

const name0OK: Type0 = "a-b-c-bootstrap-id-01";
const name0KO: Type0 = "hello"; // type error

const name1OK: Type1 = "a-dev-italynorth-bootsrap-id-01";
const name1KO: Type1 = "a-b-c-bootstrap-id-01"; // type error: "b" does not satisfy "dev | prod | uat"
const anotherName1KO: Type1 = "a-b-c-botstrap-id-01"; // typo in bootsrap

Of course, this is some knowledge share; it is not mandatory.
But if our codebase provides such types and we have some format constraints, I would evaluate adding those.

mamu0: done here [2296ad7](https://github.com/pagopa/dx/pull/1905/commits/2296ad7b58bbe4c47eeaff16b62a837afa571cab)

apps/cli/src/adapters/azure/cloud-account-service.ts
mamu0: done [2296ad7](https://github.com/pagopa/dx/pull/1905/commits/2296ad7b58bbe4c47eeaff16b62a837afa571cab)

kin0992 (approved):
Well done! 👏

Just to integrate a previous comment: https://github.com/pagopa/dx/pull/1925

that PR does provide a stricter type, so it is not possible to pass any string as `environmentShort` because of the stricter type.
It is not mandatory to merge it. If you think it is ok, feel free to merge it; otherwise you can close it

mamu0:
> Just to integrate a previous comment: #1925

@kin0992  Thank you so for the contribution! I have integrated your suggestion into this PR [fd64f62](https://github.com/pagopa/dx/pull/1905/commits/fd64f62ada25db321eca3b33f58a37c4607ed9c6)

PR #1906, opened 2026-06-29
Update Go dependencies

lucacavallaro:
Update providers dependencies to their latest version (using `go get -u`)

Also refreshes Go test dependency metadata for module checks and CodeQL analysis.

Tools dependencies (`./tools`) were not touched because will be soon migrated to the new [Tool dependencies](https://go.dev/doc/modules/managing-dependencies#tools) format, which was introduced by Go 1.24

Closes CES-2170


providers/azure/go.mod
```diff
@@ -1,28 +1,28 @@
 module github.com/pagopa-dx/terraform-provider-azure
 
-go 1.24.0
+go 1.25.8
```
krusty93: Why not 1.26.4?

kin0992 (approved):

PR #1908, opened 2026-06-29
Report precise CLI precheck failures in init and add commands

kin0992:
## What

The CLI prechecks for `init` and `add environment` collapsed several distinct failures into a single generic message, hiding the real cause unless `--verbose` was used. This surfaces the actual failure in normal mode while keeping the original cause intact for verbose output.

- **`dx init` Terraform step**: reports which step (`init`/`apply`) failed plus the underlying Execa `shortMessage`/`stderr`, instead of the static "Failed to create GitHub repository."
- **Azure precheck**: a failing `az account show` still suggests `az login`, but a failing `az group list` now surfaces the real subscription/permission/network cause.
- **`add environment`**: a genuinely missing git remote now resolves to "no detectable repository" instead of an opaque exit-1 error; other git failures (including exit 1 with stderr) surface their real cause.

## Why

All three were the same anti-pattern: a specific failure masked by a static message in non-verbose mode, making prechecks hard to diagnose.

## Tests

All CLI tests, lint, typecheck, and build pass. New tests cover each failure path.

Resolves CES-2156, CES-2157, CES-2158


apps/cli/src/adapters/commander/commands/init.ts
```diff
@@ -165,24 +166,37 @@ const azureAccountSchema = z.object({
...
+  const { stdout } = await tf$`az account show`.catch((cause: unknown) => {
+    throw asError(AZ_LOGIN_REQUIRED_MESSAGE)(cause);
+  });
   // `az account show` reads the cached CLI context, but `az group list`
   // fails when the current session token has expired.
-  await tf$`az group list`;
+  await tf$`az group list`.catch((cause: unknown) => {
+    throw mapAzAccessError(cause);
+  });
   const parsed = JSON.parse(stdout);
```
gunzip: nitpick: this may throw

apps/cli/src/adapters/commander/commands/init.ts
```diff
@@ -257,22 +304,33 @@ const createRemoteRepositoryWithPresenter =
...
+      task: () => Promise<unknown>,
+    ) => {
       try {
-        await repo$`terraform init`;
-        await repo$`terraform apply -auto-approve`;
-      } catch (error) {
-        if (error instanceof ExecaError) {
-          logger.error(error.shortMessage);
-        }
+        await task();
```
gunzip: question: what's happening here? code is not explicit as it should

apps/cli/src/adapters/github/github-repo.ts
```diff
@@ -5,8 +5,22 @@ import { githubRepoSchema } from "../../domain/github-repo.js";
...
 
 export const getGithubRepo = async (): Promise<GitHubRepo | undefined> => {
-  const result = await $`git config --get remote.origin.url`;
+  const result = await readGitRemoteOriginUrl();
   const repoUrl = result.stdout.trim();
+  const stderr = (result.stderr ?? "").trim();
+
+  // `git config --get` exits 1 with no output when the key is unset: there is
+  // simply no origin remote, which is a valid "not in a repo we manage" state.
+  if (result.exitCode === 1 && repoUrl === "" && stderr === "") {
```
gunzip: nitpick: you can give this condition a name to make it more explicit

gunzip (approved):

PR #1917, opened 2026-06-30
Upgrade JavaScript dependencies

gunzip:
Upgrade Javascript packages.


kin0992 (commented):
**question**: are you going to update also the remaining outdated dependencies, or this PR aims only to upgrade just the `dx-metrics` dependencies?

Usually I use `pnpm -ir update` to see the outdated dependencies that has `minor` and `patch` upgrade. With the `--latest` flag I see the `major` upgrades as well

krusty93 (approved):

pnpm-workspace.yaml
```diff
@@ -30,57 +30,57 @@ allowBuilds:
...
-    "@azure/storage-blob": ^12.31.0
+    "@azure/storage-blob": ^12.33.0
 
   github-actions:
     "@actions/core": ^3.0.1
     "@actions/github": ^9.1.1
   opentelemetry:
     "@azure/monitor-opentelemetry": 1.18.1
-    "@azure/monitor-opentelemetry-exporter": ^1.0.0-beta.42
+    "@azure/monitor-opentelemetry-exporter": 1.0.0-beta.42
```
kin0992, line 69: **question**: why here the caret `^` has been removed?

gunzip: This is the Copilot's explaination:

Per semver rules, prerelease identifiers are compared as strings: alphabetically, `"preview" > "dev" > "beta"`. Therefore, the range `^1.0.0-beta.42` theoretically includes all `1.0.0-*` prereleases that are alphabetically higher, not just the beta series. I tested this in isolation:

```json
"@azure/monitor-opentelemetry-exporter": "^1.0.0-beta.42"

```

→ `pnpm update` → resolves to `1.0.0-preview.6` (deprecated!)

In the actual repo, however, `@azure/monitor-opentelemetry@1.18.1` (which is pinned exactly) declares an exact peer dependency (not a range) on `@azure/monitor-opentelemetry-exporter@1.0.0-beta.42`. Since the pnpm catalog must provide a single shared version that also satisfies this constraint, pnpm cannot follow the caret toward `preview.6` (as it would break the peer dependency). Consequently, it resolves exactly to `1.0.0-beta.42`, which is the only version compatible with all consumers in the workspace.

Since this is no longer a "free" resolution but a pin enforced by an external constraint, pnpm normalizes the catalog entry by removing the caret and writing the exact version — honestly reflecting that it is no longer an independently upgradable range.

kin0992 (approved):

PR #1928, opened 2026-07-02
Run TLS e2e test with cron

mamu0:
This PR updates the TLS renewal E2E tests to run on a cron schedule every Sunday at 02:00. Additionally, the smoke tests will now trigger automatically as part of the validate workflow:
image

Resolves: CES-2189


krusty93:
question: is Slack notification enabled in case of failure?

mamu0:
> question: is Slack notification enabled in case of failure?

@krusty93 I added it in this commit [0642f0e](https://github.com/pagopa/dx/pull/1928/commits/0642f0eadd07f8529df1ea6d6297fb78788225bb)

krusty93 (approved):

PR #1934, opened 2026-07-03
Add publish failure diagnostics

christian-calabrese:
## Summary
- add verbose pre-publish logging around the Nx release publish step
- capture git state and per-project pnpm dry-run diagnostics only when publish fails
- keep the release flow unchanged apart from the extra diagnostics

## Why
This is a temporary diagnostic change to capture the real publish failure in CI/CD without patching Nx or introducing custom scripts.

### DIFF
--- a/actions/nx-release/action.yaml
+++ b/actions/nx-release/action.yaml
@@ -169,10 +169,133 @@ runs:
       run: |
         set -euo pipefail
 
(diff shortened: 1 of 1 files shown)


christian-calabrese:
Superseded by #1933, which contains the same diagnostics change from the clean default-branch worktree.

PR #1826, opened 2026-06-09
Add SaveMoney pricing module

mamu0:
This PR is the final stack branch that will merge into `main`. It introduces the internal Azure Retail Prices pricing module used by the stacked SaveMoney cost-at-risk work.

The full stack adds potential monthly cost-at-risk values for analyzed custom resources:
image

And a `cost at risk` entry in the summary:
image

Stack order: #1939 -> #1938 -> #1937 -> this PR -> `main`.

depends-on #1937

Resolves CES-2097

PR #1937, opened 2026-07-03
Wire custom cost at risk

mamu0:
Uses the pricing module from #1826 during Azure analysis and reports custom cost-at-risk separately from Advisor savings.

depends-on: #1938


.nx/version-plans/version-plan-1783348058718.md
```diff
@@ -0,0 +1,5 @@
+---
+'@pagopa/dx-savemoney': patch
+---
+
+Separate custom cost-at-risk estimates from Advisor savings and wire pricing enrichment into Azure analyzers.
```
gunzip, line 5: suggestion: I'd elaborate a little bit more this entry making it understandable by users (ie. what do you mean with 'separate'?)

mamu0: I can improve it, but this version plan is mostly indicative. The actual one that will "replace" it is [here](https://github.com/pagopa/dx/pull/1939/changes#diff-0e84025b6e56e9edee3ee55b3667829a93ffbeae967c55da3267582fdec4a280R1-R9).

This is because the current one is just a `patch`, whereas the linked one refers to the actual `minor` release.

gunzip (approved):

PR #1938, opened 2026-07-03
Add SaveMoney pricing toggle

mamu0:
Adds a CLI flag to disable best-effort pricing lookups and keeps explicit source overrides working.

depends-on #1939

### DIFF
--- a/.nx/version-plans/version-plan-1780067373607.md
+++ b/.nx/version-plans/version-plan-1780067373607.md
@@ -1,0 +1,5 @@
+---
+"@pagopa/dx-cli": patch
+---
+
+- Add `--no-pricing` flag to the `savemoney` command to disable custom cost-at-risk pricing lookups.
--- a/.nx/version-plans/version-plan-1780067373607.md
+++ b/.nx/version-plans/version-plan-1780067373607.md
@@ -2,4 +2,4 @@
 "@pagopa/dx-cli": patch
 ---
 
-- Add `--no-pricing` flag to the `savemoney` command to disable custom cost-at-risk pricing lookups.
+- Add `--no-pricing` flag to the `savemoney` command so users can skip custom cost-at-risk pricing lookups when they need a faster offline run or want to avoid external Retail Prices API calls.
--- a/apps/cli/src/adapters/commander/commands/savemoney.ts
+++ b/apps/cli/src/adapters/commander/commands/savemoney.ts
@@ -54,11 +54,18 @@ export const makeSavemoneyCommand = () =>
         const finalConfig: AzureConfig = {
           ...config,
           filterTags,
-          preferredLocation: options.location || config.preferredLocation,
+          preferredLocation: resolveStringOption(
+            options.location,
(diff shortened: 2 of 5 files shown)


apps/cli/src/adapters/commander/commands/savemoney.ts
```diff
@@ -120,3 +129,17 @@ export function parseTagsOption(
   }
   return result;
 }
+
+/**
+ * Resolves the source filter preserving the difference between an omitted
+ * option (respect config/defaults) and an explicit `--source all` override.
+ */
+export function resolveSourcesOption(
```
kin0992, line 137: **thought**: I think there is something a little bit wired to types here.

I am going to explain this better with a draft PR to show my doubts there

kin0992: you can find more details on PR #1954 and #1958

mamu0: I see the point: after `loadConfig()` the Zod default should make `sources` always available, so the TypeScript type could be stricter than `sources?: AzureSource[]`.

I would keep this out of this PR though, because changing `AzureConfig`, `schema.ts`, and `loadAzureConfig()` touches the SaveMoney package model and expands the scope beyond the CLI pricing toggle. Also, the stricter tuple approach in #1958 needs care: `z.tuple([sourceEnum, sourceEnum]).rest(sourceEnum)` requires at least two sources, while `["custom"]` is a valid documented config.

So I’d keep #1938 focused and handle the config typing cleanup in a separate SaveMoney PR, using a non-empty tuple with one required element if we decide to go that way.

kin0992: looks good to me, but I'd track this on Jira, because there is a smell there: decoder defines a stricter type (so passing a non empty array to the decode function would throw), but in our types (like domain types) we are allowing them to be undefined or empty, which to me looks like a misalignment

.nx/version-plans/version-plan-1780067373607.md
```diff
@@ -0,0 +1,5 @@
+---
+"@pagopa/dx-cli": patch
+---
+
+- Add `--no-pricing` flag to the `savemoney` command to disable custom cost-at-risk pricing lookups.
```
gunzip: suggestion: I'd try to elaborate a little bit more in order to clarify the expected result for users

gunzip: for example: why and when a user should users disable this feature?

mamu0: done [c2a50bc](https://github.com/pagopa/dx/pull/1938/commits/c2a50bc926f882d96e3a3cbe915968516492fb1d)

apps/cli/src/adapters/commander/commands/savemoney.ts
mamu0: i've created a task on Jira (CES-2217)

PR #1941, opened 2026-07-06
Fix integration test environment configuration

krusty93:
Update the environment configuration to use the correct variables for naming and application settings, ensuring proper integration test setup.

### DIFF
--- a/infra/modules/azure_container_app/tests/integration.tftest.hcl
+++ b/infra/modules/azure_container_app/tests/integration.tftest.hcl
@@ -14,7 +14,7 @@ variables {
     env_short       = "d"
     location        = "italynorth"
     domain          = "int"
-    app_name        = "ca"
+    app_name        = "def"
     instance_number = "01"
   }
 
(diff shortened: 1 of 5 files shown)


infra/modules/azure_container_app/tests/integration.tftest.hcl
```diff
@@ -14,7 +14,7 @@ variables {
     env_short       = "d"
     location        = "italynorth"
     domain          = "int"
-    app_name        = "ca"
+    app_name        = "def"
```
kin0992, line 17: **question**: is it right `def`?

I see later you used `dev`, is this a typo?

krusty93: yep, it stands for `default` but I'm afraid of hitting the length limit here. Below instead I set `dev` as the `use_case` tested there is `development`.

BTW Names are not important here, these are just ephemeral resources

gunzip (approved):

kin0992 (approved):

PR #1946, opened 2026-07-07
Add module.json files for Terraform modules and update nx.json for GitHub publishing

krusty93:
Introduce module.json files for various Terraform modules to facilitate releases via NX. Update nx.json to configure GitHub publishing options for the project.

Close CES-2190

### DIFF
--- a/.nx/version-plans/version-plan-1783411380197.md
+++ b/.nx/version-plans/version-plan-1783411380197.md
@@ -1,0 +1,36 @@
+---
+github_selfhosted_runner_on_container_app_jobs: patch
+github_selfhosted_runner_on_codebuild: patch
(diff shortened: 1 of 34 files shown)


gunzip:
shouldn't we remove package.json in each module as well?

krusty93:
> shouldn't we remove package.json in each module as well?

According to the spec linked in the Jira ticket, it is suggested to open another PR to remove `package.json`s and updating the GitHub workflow as well: https://github.com/pagopa/dx/blob/d7fc64a6c8ad71214118f430f560d5f0f20b1ee4/docs/superpowers/specs/2026-05-07-nx-terraform-publish-design.md#migration-and-rollout

gunzip (approved):

PR #1953, opened 2026-07-07
Publish executor fails on 'nothing to commit' race with legacy subrepo sync

krusty93:
## Root cause

Run https://github.com/pagopa/dx/actions/runs/28875477370/job/85649279947 (publish for `azure_app_service` + `azure_app_service_exposed` @ 4.0.1) failed at `git commit` with `nothing to commit, working tree clean`, even though the tag `4.0.1` didn't exist yet remotely.

This happens because the **legacy** `.github/workflows/_release-bash-modules-to-subrepo.yaml` also triggers on the same push (any change under `infra/modules/**`) and its `push-to-subrepo` action pushes the module's current content to the subrepo's `main` branch **unconditionally** — it only conditions the *tag* push on tag-existence, not the content push to `main`. So by the time `nx-release-publish` ran, `main` already matched the module's current tree (pushed moments earlier by the legacy workflow), leaving nothing to commit — and the executor treated that as a hard failure instead of a no-op, so it never created/pushed the `4.0.1` tag.

Both pilot modules' subrepos DO already have the right content on `main` (from the legacy workflow), but are missing the `4.0.1` tag — that's why the user sees updated README/module.json in the subrepos despite the failed run.

Separately found and fixed a latent bug: the commit message template embedded literal `"` characters (`` `git commit -m "Release ${version}"` ``), which `execa` split into two arguments instead of one message string.

## Fix

- `git commit` failures containing `nothing to commit` are now treated as a no-op (content already up to date) instead of throwing — tag creation + push still happen afterwards.
- Commit message is now passed as a single interpolated argument, avoiding the quoting bug.

## Not in scope

This does not remove the race itself (legacy bash workflow still pushes to `main` unconditionally in parallel with the new flow) — that's covered by the planned Phase D retirement of the legacy workflow. This PR just makes the new publish flow resilient to it in the meantime.

## Verification

- `pnpm nx test @pagopa/nx-terraform-plugin` — 109/109 passing (added 2 new tests: no-op on "nothing to commit", real commit failures still throw).
- `pnpm nx run-many --target=lint,typecheck,build --projects=@pagopa/nx-terraform-plugin` — clean.
- Version plan included (patch bump for `@pagopa/nx-terraform-plugin`).

## Next step after merge

Re-trigger the publish for the two pilot modules via `workflow_dispatch` recovery mode (`RELEASE_MODE=publish-all`) on `_release-changesets-pull-request.yaml`/`release-v2.yaml` once this is merged, so `4.0.1` tags get created for both.

Close CES-2190

### DIFF
--- a/packages/nx-terraform-plugin/dist/executors/publish/publish.js
+++ b/packages/nx-terraform-plugin/dist/executors/publish/publish.js
@@ -81,8 +81,7 @@ const publishToGithub = async (input) => {
 				GIT_AUTHOR_NAME: "PagoPA DX Bot",
 				GIT_COMMITTER_EMAIL: "pagopa-dx-bot@pagopa.it",
 				GIT_COMMITTER_NAME: "PagoPA DX Bot"
-			},
-			shell: true
+			}
 		});
 		const safe$ = $$1({ reject: false });
 		await $$1`git init -b main`;
@@ -100,7 +99,11 @@ const publishToGithub = async (input) => {
 			await clearExportWorkingTree(tempExportDir);
 			await copyModuleDirectoryContents(sourceModuleDirectory, tempExportDir);
 			await $$1`git add --all`;
-			await $$1`git commit -m "Release ${input.version}"`;
+			const commitResult = await safe$`git commit -m ${`Release ${input.version}`}`;
+			if (commitResult.exitCode !== 0) {
+				const commitOutput = `${commitResult.stdout}${commitResult.stderr}`;
+				if (!commitOutput.includes("nothing to commit")) throw new Error(`Failed to commit release ${input.version} for ${repoUrl}: ${commitOutput}`);
+			}
 			await $$1`git tag -f ${input.version}`;
 			await $$1`git push origin main`;
 			await $$1`git push origin refs/tags/${input.version} --force`;
--- a/packages/nx-terraform-plugin/dist/executors/publish/publish.js
+++ b/packages/nx-terraform-plugin/dist/executors/publish/publish.js
@@ -81,7 +81,8 @@ const publishToGithub = async (input) => {
 				GIT_AUTHOR_NAME: "PagoPA DX Bot",
 				GIT_COMMITTER_EMAIL: "pagopa-dx-bot@pagopa.it",
 				GIT_COMMITTER_NAME: "PagoPA DX Bot"
-			}
+			},
+			shell: true
 		});
 		const safe$ = $$1({ reject: false });
 		await $$1`git init -b main`;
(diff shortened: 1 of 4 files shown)


packages/nx-terraform-plugin/src/adapters/github/__tests__/publisher.test.ts
```diff
@@ -445,6 +445,66 @@ it("creates and force-pushes a tag named after version", async () => {
...
+    "git ls-remote --exit-code --tags origin refs/tags/1.2.3",
+    "git ls-remote --exit-code --heads origin main",
+    "git fetch origin main",
+    "git checkout -B main origin/main",
+    "git add --all",
+    "git commit -m Release 1.2.3",
+    "git tag -f 1.2.3",
+    "git push origin main",
+    "git push origin refs/tags/1.2.3 --force",
+  ]);
```
gunzip, line 485: suggestion: this test is fragile as is testing too much

packages/nx-terraform-plugin/dist/executors/publish/publish.js
```diff
@@ -100,7 +100,11 @@ const publishToGithub = async (input) => {
 			await clearExportWorkingTree(tempExportDir);
 			await copyModuleDirectoryContents(sourceModuleDirectory, tempExportDir);
 			await $$1`git add --all`;
-			await $$1`git commit -m "Release ${input.version}"`;
+			const commitResult = await safe$`git commit -m ${`Release ${input.version}`}`;
```
gunzip: question: have you tested this change? I suspect that the previous version was the right syntax to use with execa

packages/nx-terraform-plugin/src/adapters/github/__tests__/publisher.test.ts
```diff
@@ -445,6 +445,66 @@ it("creates and force-pushes a tag named after version", async () => {
   ]);
 });
 
+it("still tags and pushes when there is nothing to commit (content already matches remote)", async () => {
+  const { commands } = createGitCommandHarness({
+    onCommand: (command) => {
+      if (command === "git commit -m Release 1.2.3") {
```
gunzip: suggestion: afaik should be quoted
```suggestion
      if (command === "git commit -m \"Release 1.2.3\"") {
```

gunzip:
please don't. use conventional commits for PR titles

packages/nx-terraform-plugin/src/adapters/github/__tests__/publisher.test.ts
krusty93: it's a workaround that I need to make sure the workflow works. Changes are already on the subrepo, so the commit is empty and the current workflow fails. So the new flow (applied to two modules only) is broken now.
Once done, I'll revert this change

packages/nx-terraform-plugin/dist/executors/publish/publish.js
krusty93: I'm testing it now

krusty93:
> please don't. use conventional commits for PR titles

it's copilot, I manually edited it. Maybe it's lack of instructions

christian-calabrese (approved):

PR #1954, opened 2026-07-08
Some types enhacements

kin0992:
> [!NOTE]
> This PR is just a review on another PR

### DIFF
--- a/apps/cli/src/adapters/commander/commands/__tests__/savemoney.test.ts
+++ b/apps/cli/src/adapters/commander/commands/__tests__/savemoney.test.ts
@@ -77,25 +77,22 @@ describe("parseTagsOption", () => {
 
 describe("resolveSourcesOption", () => {
   it("uses config sources when --source is omitted", () => {
-    expect(resolveSourcesOption(undefined, ["custom"])).toEqual(["custom"]);
+    expect(resolveSourcesOption(["custom"])).toEqual(["custom"]);
   });
 
-  it("uses both sources when --source is omitted and config has no sources", () => {
-    expect(resolveSourcesOption(undefined, undefined)).toEqual([
-      "advisor",
-      "custom",
-    ]);
+  it.skip("uses both sources when --source is omitted and config has no sources", () => {
+    expect(resolveSourcesOption([])).toEqual(["advisor", "custom"]);
   });
 
   it("lets explicit --source all override a narrowed config", () => {
-    expect(resolveSourcesOption("all", ["custom"])).toEqual([
+    expect(resolveSourcesOption(["custom"], "all")).toEqual([
       "advisor",
       "custom",
     ]);
   });
 
   it("uses the explicit single source", () => {
-    expect(resolveSourcesOption("advisor", ["custom"])).toEqual(["advisor"]);
+    expect(resolveSourcesOption(["custom"], "advisor")).toEqual(["advisor"]);
   });
 });
 
--- a/apps/cli/src/adapters/commander/commands/__tests__/savemoney.test.ts
+++ b/apps/cli/src/adapters/commander/commands/__tests__/savemoney.test.ts
@@ -13,7 +13,9 @@ import {
   makeSavemoneyCommand,
   parseSourceOption,
   parseTagsOption,
+  resolveNumberOption,
   resolveSourcesOption,
+  resolveStringOption,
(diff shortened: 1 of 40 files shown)


apps/cli/src/adapters/commander/commands/savemoney.ts
```diff
@@ -46,7 +46,7 @@ export const makeSavemoneyCommand = () =>
       const { verbose } = this.optsWithGlobals<GlobalOptions>();
       try {
         // Load configuration from YAML (includes subscriptionIds, location, timespanDays, thresholds)
-        const config: AzureConfig = await loadConfig(options.config);
+        const config = await loadConfig(options.config);
```
kin0992, line 49: **nitpick**: type here is redundant since `loadConfig` function has a well-defined signature

nothing wrong here though

apps/cli/src/adapters/commander/commands/savemoney.ts
```diff
@@ -135,11 +135,11 @@ export function parseTagsOption(
  * option (respect config/defaults) and an explicit `--source all` override.
  */
 export function resolveSourcesOption(
-  option: SourceOption | undefined,
-  configSources: AzureSource[] | undefined,
+  configSources: AzureSource[],
```
kin0992, line 138: **note**: swapped argument order here and change syntax to `option?`.

Rationale: the `AzureSource` type defines it to have a default (or at least [its documentation says that](https://github.com/pagopa/dx/blob/daf530554d48775ada4e242280b0d2ad5d130ac7/packages/savemoney/src/azure/types.ts#L30-L37)) and the zod schema that parses the data to a type actually [set a default value](https://github.com/pagopa/dx/blob/daf530554d48775ada4e242280b0d2ad5d130ac7/packages/savemoney/src/schema.ts#L119-L122), so it looks impossible to get an undefined value here. Moreover, it looks that this should be a NonEmpty array, and so the type can be even more stricter (`[AzureSource, ...AzureSource[]]`, which looks a punch in the eye reading it, but it means `An array of at least one AzureSource element` -> a non empty array.

Then, the argument have been swapped just to simplify the syntax to call the function. Having the required arguments first and the optional at the end allows you to omit the optional arguments, so you can call the function like this
```typescript
resolveSourceOption([]);
```
and this can be useful for tests.

packages/savemoney/src/azure/types.ts
```diff
@@ -40,7 +40,7 @@ export type AzureConfig = BaseConfig & {
    *
    * Defaults to `["advisor", "custom"]` when omitted (i.e. all sources).
    */
-  sources?: AzureSource[];
+  sources: AzureSource[];
```
kin0992, line 43: **note**: following what [explained in another comment](https://github.com/pagopa/dx/pull/1954/changes#r3542264830), this cannot be optional

packages/savemoney/src/azure/config.ts
```diff
@@ -73,6 +73,7 @@ export async function loadAzureConfig(
 
   return {
     preferredLocation: "italynorth",
+    sources: ["advisor", "custom"],
```
kin0992, line 76: **note**: since there is no way to have sources undefined, it must return a value; so here it returns the default values

apps/cli/src/adapters/commander/commands/__tests__/savemoney.test.ts
```diff
@@ -77,25 +77,22 @@ describe("parseTagsOption", () => {
...
+    expect(resolveSourcesOption(["custom"])).toEqual(["custom"]);
   });
 
-  it("uses both sources when --source is omitted and config has no sources", () => {
-    expect(resolveSourcesOption(undefined, undefined)).toEqual([
-      "advisor",
-      "custom",
-    ]);
+  it.skip("uses both sources when --source is omitted and config has no sources", () => {
+    expect(resolveSourcesOption([])).toEqual(["advisor", "custom"]);
```
kin0992, line 84: **note**: this is skipped because i am not sure that can be valuable according to the comments made on this PR.

but I may have lost some context here

apps/cli/src/adapters/commander/commands/savemoney.ts
kin0992: you can see an example of sticter types on PR #1958

PR #1958, opened 2026-07-08
Stricten types

kin0992:
> [!NOTE]
> This PR is just a review on another PR

### DIFF
--- a/packages/savemoney/src/azure/types.ts
+++ b/packages/savemoney/src/azure/types.ts
@@ -40,7 +40,7 @@ export type AzureConfig = BaseConfig & {
    *
    * Defaults to `["advisor", "custom"]` when omitted (i.e. all sources).
    */
-  sources: AzureSource[];
+  sources: [AzureSource, ...AzureSource[]];
   subscriptionIds: string[];
   /**
    * Analysis thresholds. Defaults from DEFAULT_THRESHOLDS are used when not provided.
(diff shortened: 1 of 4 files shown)


kin0992:
With this PR, which is an improvement of #1954, at type level it is not possible to have an empty (or undefined) array, which follows the zod decoder.

packages/savemoney/src/schema.ts
```diff
@@ -117,8 +117,8 @@ const AzureSectionSchema = z
      * Azure Advisor recommendations, or `["custom"]` to skip Advisor.
      */
     sources: z
-      .array(z.enum(["advisor", "custom"]))
-      .nonempty()
+      .tuple([sourceEnum, sourceEnum])
+      .rest(sourceEnum)
```
kin0992, line 121: **note**: this is neccessary due to a simplification made by zod

it looks like zod simplifies the `z.array().nonempty()` to a simple typescript array. please note that `string[]` allows you to pass an empty array.

to make sure the typescript type, inferred by the zod schema, follows the non-emptiness requirement, then we should parse it as a tuple

packages/savemoney/src/azure/types.ts
```diff
@@ -40,7 +40,7 @@ export type AzureConfig = BaseConfig & {
    *
    * Defaults to `["advisor", "custom"]` when omitted (i.e. all sources).
    */
-  sources: AzureSource[];
+  sources: [AzureSource, ...AzureSource[]];
```
kin0992, line 43: **note**: an alternative here could be `Config["azure"]["sources"];`

kin0992, line 43: or, another way could be to define a `type AzureSources = [AzureSource, ...AzureSource[]]` and so the AzureSources type represents the non-empty array

PR #1960, opened 2026-07-08
Authenticate git push to subrepo in publish executor

krusty93:
## Why

Since PR #1953 merged and the release workflow ran for real against a pilot module, `git push origin main` to the module's GitHub subrepo failed:

```
fatal: could not read Username for 'https://github.com': No such device or address
```

Root cause: `publishToGithub` (in `packages/nx-terraform-plugin/src/adapters/github/publisher.ts`) only wired the GitHub App installation token into Octokit (`ensureGitHubRepository`), used to create the subrepo if missing. The git CLI subprocess (init/fetch/checkout/push) never received any credentials. Anonymous HTTPS git reads succeed against public repos, but `git push` requires auth, and in non-interactive CI git can't prompt for a username.

## What changed

- `octokit.ts`: export the existing `getGitHubToken()` helper so it can be reused outside the Octokit client.
- `publisher.ts`:
  - `git remote add origin` now embeds the token as the `x-access-token` user (`https://x-access-token:<token>@github.com/...`) when available, falling back to the plain URL otherwise. All error messages/logs keep using the token-free URL, and `git` itself redacts credentials from its own error output (verified locally against a real auth failure), so the token is never printed.
  - Fixed a related latent bug in the same commit step: with execa's `shell: true`, an interpolated `` git commit -m ${commitMessage} `` gets word-split by the shell on whitespace, breaking multi-word release messages (e.g. "Release 1.2.3" turning into a bogus extra arg). The message is now quoted literally in the template.
- Updated/added unit tests in `publisher.test.ts` covering both the token-embedded and token-free remote URL cases, and updated existing assertions to the quoted commit-message form.
- Added a version plan (`patch` bump for `@pagopa/nx-terraform-plugin`).

## Validation

- `pnpm nx run-many --target=test,lint,typecheck,build --projects=@pagopa/nx-terraform-plugin` — 111/111 tests passing, lint/typecheck/build clean.
- Manually verified git redacts embedded credentials from its own error output (tested locally with a fake token against an unreachable repo).

Not yet validated against a real GitHub Actions run — this should be watched closely on the next release for these pilot modules, since it's possible the GitHub App installation isn't authorized with write access on the `pagopa-dx` org yet (in which case the next failure would be a clear permission error instead of the current confusing username error).

Close CES-2190

### DIFF
--- a/packages/nx-terraform-plugin/src/adapters/github/publisher.ts
+++ b/packages/nx-terraform-plugin/src/adapters/github/publisher.ts
@@ -5,7 +5,7 @@ import { cp, mkdtemp, readdir, rm } from "node:fs/promises";
 import { tmpdir } from "node:os";
 import { basename, join } from "node:path";
 
-import { ensureGitHubRepository } from "./octokit.ts";
+import { ensureGitHubRepository, getGitHubToken } from "./octokit.ts";
 
 export interface PublishToGithubInput {
   description: string;
@@ -88,7 +88,23 @@ export const publishToGithub = async (
     const safe$ = $({ reject: false });
 
     await $`git init -b main`;
-    await $`git remote add origin ${repoUrl}`;
+
+    // Reads/writes to the subrepo need credentials. Anonymous HTTPS can read a
+    // public repo, but `git push` requires auth or git prompts for a username
+    // (fatal in CI). Embed the GitHub App installation token as the
+    // `x-access-token` user in the remote URL when available. Keep `repoUrl`
+    // (token-free) for logs and error messages so the token never leaks, and
+    // use `safe$` for the remote add so a failure can't echo the token URL.
+    const token = getGitHubToken();
+    const authenticatedRepoUrl =
+      token === undefined
+        ? repoUrl
+        : `https://x-access-token:${token}@github.com/${input.githubOwner}/${repo}.git`;
+    const remoteAdd =
+      await safe$`git remote add origin ${authenticatedRepoUrl}`;
+    if (remoteAdd.exitCode !== 0) {
+      throw new Error(`Failed to add git remote origin for ${repoUrl}`);
+    }
 
     const remoteTag =
       await safe$`git ls-remote --exit-code --tags origin refs/tags/${input.version}`;
@@ -118,6 +134,10 @@ export const publishToGithub = async (
       // contents (e.g. a concurrent legacy sync pushed the same tree first).
       // In that case there's nothing to commit, but we must still tag and
       // push so the release completes instead of failing.
+      //
+      // `shell: true` makes execa join argv into a single string that the
+      // shell re-parses, so an interpolated value must be quoted here or it
+      // gets word-split on whitespace (e.g. "Release 1.2.3" -> two args).
       const commitResult =
         await safe$`git commit -m "Release ${input.version}"`;
       if (commitResult.exitCode !== 0) {
--- a/packages/nx-terraform-plugin/src/adapters/github/publisher.ts
+++ b/packages/nx-terraform-plugin/src/adapters/github/publisher.ts
@@ -92,14 +92,20 @@ export const publishToGithub = async (
     // Reads/writes to the subrepo need credentials. Anonymous HTTPS can read a
     // public repo, but `git push` requires auth or git prompts for a username
     // (fatal in CI). Embed the GitHub App installation token as the
-    // `x-access-token` user in the remote URL when available. Keep `repoUrl`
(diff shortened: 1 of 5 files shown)


packages/nx-terraform-plugin/src/adapters/github/publisher.ts
```diff
@@ -88,7 +88,23 @@ export const publishToGithub = async (
...
+    // public repo, but `git push` requires auth or git prompts for a username
+    // (fatal in CI). Embed the GitHub App installation token as the
+    // `x-access-token` user in the remote URL when available. Keep `repoUrl`
+    // (token-free) for logs and error messages so the token never leaks, and
+    // use `safe$` for the remote add so a failure can't echo the token URL.
+    const token = getGitHubToken();
+    const authenticatedRepoUrl =
+      token === undefined
+        ? repoUrl
+        : `https://x-access-token:${token}@github.com/${input.githubOwner}/${repo}.git`;
```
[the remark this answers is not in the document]
krusty93: applied

packages/nx-terraform-plugin/src/adapters/github/publisher.ts
```diff
@@ -88,7 +88,29 @@ export const publishToGithub = async (
...
+    const token = getGitHubToken();
+    let authenticatedRepoUrl = repoUrl;
+    if (token !== undefined) {
+      const url = new URL(repoUrl);
+      url.username = "x-access-token";
+      url.password = token;
+      authenticatedRepoUrl = url.toString();
+    }
+    const remoteAdd =
+      await safe$`git remote add origin ${authenticatedRepoUrl}`;
```
kin0992: **nitpick**: this could be extracted to a function and have a smoother flow like

```typescript
const getAuthenticatedUrl = (repoUrl: string) => {
    const token = getGitHubToken();
    if (token) {
        const url = new URL(repoUrl);
        url.username = ...
        url.password = ...
        return url.toString()
    }
    return repoUrl;
}
```
and the caller would be 
```typescript
const authenticatedRepoUrl = getAuthenticatedUrl(repoUrl);
const remoteAdd =
      await safe$`git remote add origin ${authenticatedRepoUrl}`;
...
```

kin0992 (approved):

packages/nx-terraform-plugin/src/adapters/github/publisher.ts
```diff
@@ -88,7 +88,29 @@ export const publishToGithub = async (
...
+    if (remoteAdd.exitCode !== 0) {
+      throw new Error(`Failed to add git remote origin for ${repoUrl}`);
+    }
 
     const remoteTag =
       await safe$`git ls-remote --exit-code --tags origin refs/tags/${input.version}`;
```
gunzip, line 109: **Suggestion: simplify with `gh auth setup-git` instead of embedding the token in the URL**

The entire token-in-URL block could be replaced by a single `gh auth setup-git` call before the remote is added:

```typescript
// gh auth setup-git installs a git credential helper that reads GH_TOKEN /
// GITHUB_TOKEN and provides credentials for github.com HTTPS remotes.
// This keeps the URL clean (no embedded secret), avoids URL manipulation,
// and covers every subsequent git operation (ls-remote, fetch, push)
// without any extra work.
if (getGitHubToken() !== undefined) {
  await $`gh auth setup-git`;
}
await $`git remote add origin ${repoUrl}`;
```

Benefits over the current approach:
- **No token in git config or process list.** Embedding credentials in the remote URL writes them into `.git/config` and makes them visible in `ps`/`git remote -v` output. `gh auth setup-git` uses a credential helper so the token is never stored in the URL.
- **Covers all git operations automatically.** `ls-remote`, `fetch`, and `push` all go through the same credential helper — nothing to wire up per-call.
- **Less code.** The `URL` manipulation, `getGitHubToken` export, and extra test cases are no longer needed.
- **`gh` is already available in the CI environments this runs in**, and `GH_TOKEN`/`GITHUB_TOKEN` are the same variables `gh` reads by default.

`gh auth setup-git` writes a credential helper entry to the user-level `.gitconfig` (or to the scope passed via `--hostname`), so it works fine even inside the temporary `git init` directory.

krusty93: done

packages/nx-terraform-plugin/src/adapters/github/publisher.ts
krusty93: removed due to the other suggestion

PR #1963, opened 2026-07-08
Activate Nx Docker plugin in this repository

christian-calabrese:
Activates `@pagopa/nx-dx-docker-plugin` in this repository and includes the plugin package introduced through #1962.

## Why a DX plugin

`@nx/docker` remains responsible for `docker:run` and the standard Nx Docker version flow. The DX plugin owns `docker:build` and `docker:push` because we need parity with `docker/metadata-action`: complete OCI metadata, reproducibility flags, multi-platform builds, and branch/SHA/semver aliases. Extending the inferred `@nx/docker` targets is fragile because Nx merges colliding inferred targets by replacement rather than deeply merging their options; dedicated targets provide one predictable source of truth.

Tags, timestamps, and GitHub metadata are computed by the executors at task runtime, not while creating the project graph. Nx caches inferred targets by file inputs, not CI environment variables, so graph-time computation could reuse stale metadata across runs.

A custom `nx-release-publish` executor is used only for projects configured with `nx.release.docker.repositoryName`. The standard Nx publisher pushes the primary version tag only and does not invoke `docker:push`; the DX executor also publishes major, minor, and `latest` aliases while reusing Nx's generated Docker version.

## Conventions and compatibility

- projects with a `Dockerfile` receive language-agnostic `docker:build` and `docker:push` targets without required plugin options
- registry, default branch, repository metadata, target names, and multi-platform defaults follow DX conventions; descriptive metadata remains optionally overridable
- existing image names are preserved through the single `nx.release.docker.repositoryName` setting
- Docker builds default to the workspace-root context so monorepo Dockerfiles can copy shared files; project-specific contexts remain explicit overrides
- the plugin is emitted as CommonJS for compatibility with Nx versions whose plugin loader mutates the imported namespace
- built plugin output is committed because Nx must load the plugin to create the project graph before any build target can run

The PR also configures `{versionActionsVersion}`, aligns existing E2E Docker workflows with the new project-root contexts, and adds version plans for the affected applications and plugin.

#1964 adds the website documentation and targets this branch before this integration PR merges into `main`.

Depends on #1964


christian-calabrese:
Demo video [here](https://drive.google.com/file/d/1qY3By5tVvObFoZP1mRZyf6Pdx3sGjOB9/view?usp=sharing)!

kin0992 (approved):

gunzip:
## Copilot review findings

### `fix(docker): Preserve stable aliases for prereleases`

`packages/nx-dx-docker-plugin/src/docker-image.ts:179-184` generates major/minor aliases for prerelease versions. Publishing `0.2.11-rc.0` therefore also publishes `0.2`, replacing the stable channel with a release candidate. Prereleases should emit only the complete prerelease tag.

**Confidence:** High  
**Impact:** High — users pulling the stable minor tag can unexpectedly receive prerelease images.

### `fix(docker): Apply project overrides to push targets`

`packages/nx-dx-docker-plugin/src/index.ts:130-149` applies project-level context and platform overrides only to `docker:build`. The generated `docker:push` target keeps the defaults, so it can rebuild from the repository root, fail to resolve Dockerfile paths such as `COPY src/go.mod`, or target unsupported architectures.

**Confidence:** High  
**Impact:** High — affected push targets can fail or publish images with unintended platforms and build inputs.

### `fix(release): Make the MCP Docker publisher runnable`

`apps/mcpserver/package.json:5-15` replaces the npm publisher with a Docker-only publisher that expects a locally tagged image. The existing release workflow builds only the application target, does not build/tag the Docker image, does not log in to the registry, and lacks package write permissions. As a result, `nx release publish` cannot push the image and the previous npm publication no longer occurs.

**Confidence:** High  
**Impact:** High — the MCP release workflow will fail after merging a version packages PR and will stop publishing the npm package.

### `fix(docker): Retain scopes in image repository names`

`packages/nx-dx-docker-plugin/src/docker-image.ts:60-66,85-86` removes npm scopes when deriving image names. Distinct projects such as `@team-a/api` and `@team-b/api` therefore resolve to the same image repository and can overwrite each other's tags.

**Confidence:** High  
**Impact:** Medium — scoped projects sharing a suffix cannot safely publish to the same registry prefix.

— **GitHub Copilot**

gunzip:
@christian-calabrese I've added copilot findings, feel free to dismiss them if useless.

package.json
```diff
@@ -8,6 +8,7 @@
     "@nx/js": "22.7.5",
     "@nx/plugin": "22.7.5",
     "@nx/vitest": "22.7.5",
+    "@pagopa/nx-dx-docker-plugin": "workspace:^",
```
mamu0, line 11: [suggestion(non-blocking)](https://pullpo.io/cc?l=suggestion&d=non-blocking) 
I noticed that `-dx-` was included in the plugin name, whereas it wasn't used for the Terraform plugin (which is also provided by DX). To maintain a consistent naming convention, we could remove it here as well, resulting in `@pagopa/nx-docker-plugin`.

This is non-blocking, so feel free to evaluate whether to make the change. If you don't think it's necessary, you can go ahead and resolve this comment.

PR #1975, opened 2026-07-10
Add shared glossary for AIEPFD plugin

kin0992:
Add a plugin-level glossary for `plugins/aiepfd` and have the `prd-alchemist` skill read it before proceeding.


plugins/aiepfd/references/glossary.md
```diff
@@ -0,0 +1,104 @@
+# AIEPFD glossary
+
+Shared domain glossary for skills in the `aiepfd` plugin. Read this file before
+interpreting or producing plugin artifacts so terms are used consistently across
+skills.
+
+This glossary mirrors the terminology in the Confluence page _Glossario del
```
gunzip: thoughts: as every other template is in english we may leave even this glossary in english

lucacavallaro (commented):
@kin0992, could you please explain the rationale behind this change? The glossary has ~2,000 tokens, which will be loaded upfront with every skill invocation.

I noticed that some terms in the glossary are very common (e.g., CI, CI/CD), so we might not need to instruct the model on these concepts.

Have you conducted any evals against these skills to measure the performance impact?

kin0992:
> @kin0992, could you please explain the rationale behind this change? The glossary has ~2,000 tokens, which will be loaded upfront with every skill invocation.
> 
> I noticed that some terms in the glossary are very common (e.g., CI, CI/CD), so we might not need to instruct the model on these concepts.
> 
> Have you conducted any evals against these skills to measure the performance impact?

@lucacavallaro the goal of this PR is to provide a common glossary to the skills of the plugin, so that is it not necessary to duplicate it inside them.
At the moment no evals have been executed except those during development (local only).

krusty93:
> > @kin0992, could you please explain the rationale behind this change? The glossary has ~2,000 tokens, which will be loaded upfront with every skill invocation.
> > I noticed that some terms in the glossary are very common (e.g., CI, CI/CD), so we might not need to instruct the model on these concepts.
> > Have you conducted any evals against these skills to measure the performance impact?
> 
> @lucacavallaro the goal of this PR is to provide a common glossary to the skills of the plugin, so that is it not necessary to duplicate it inside them. At the moment no evals have been executed except those during development (local only).

Perhaps we can drastically reduce the dimension by removing industry standard terms? I don't think it's meaningful to have an exact copy of the glossary actually on Confluence

lucacavallaro:
> > > @kin0992, could you please explain the rationale behind this change? The glossary has ~2,000 tokens, which will be loaded upfront with every skill invocation.
> > > I noticed that some terms in the glossary are very common (e.g., CI, CI/CD), so we might not need to instruct the model on these concepts.
> > > Have you conducted any evals against these skills to measure the performance impact?
> > 
> > 
> > @lucacavallaro the goal of this PR is to provide a common glossary to the skills of the plugin, so that is it not necessary to duplicate it inside them. At the moment no evals have been executed except those during development (local only).
> 
> Perhaps we can drastically reduce the dimension by removing industry standard terms? I don't think it's meaningful to have an exact copy of the glossary actually on Confluence

maybe we don't need the shared glossary at all, each skill can include (if required) only the terms it uses

plugins/aiepdf/references/glossary.md
```diff
@@ -0,0 +1,104 @@
+# AIEPFD glossary
+
+Shared domain glossary for skills in the `aiepfd` plugin. Read this file before
```
mamu0:
```suggestion
# AIEPDF glossary

Shared domain glossary for skills in the `aiepdf` plugin. Read this file before
```

plugins/aiepdf/skills/dr-blacksmith/SKILL.md
```diff
@@ -11,6 +11,12 @@ document. Product, Engineering, Design, control functions, and delivery
 stakeholders use it to review feasibility, align system behavior and contracts,
 and guide implementation.
 
+## Shared glossary
```
mamu0: [question(non-blocking)](https://pullpo.io/cc?l=question&d=non-blocking) 
I've seen this shared glossary into `dr-blacksmith`, `jira-magister`, `prd-alchemist`, and `uc-engraver`, but not into `confluence-librarian`, is this omission intentional?

krusty93:
I suggest to put this PR in draft and re-evaluate it in two month since now while iterating on skills improvement

PR #1986, opened 2026-07-15
Deploy Docker workloads after release

christian-calabrese:
## Summary

- deploy DX Metrics directly from the exact GHCR image published by the project release
- promote the released ARM64 MCP Server image from GHCR to ECR, then deploy that ECR image to the Lambda environments
- remove the redundant DX Metrics Import image workflow: the release process already publishes it, including `latest` for stable releases only
- trigger deployments from the matching Nx project release tags
- preserve the `release-azure-containerapp-v1.yaml` reusable workflow contract: existing build inputs and `sha-*` deployment behavior remain the default; `docker_image_tag` opt-in skips the build and deploys the released GHCR image

Deployments use the explicit release version, never `latest`, so release candidates cannot be deployed accidentally as a stable image.

## Validation

- `pnpm exec prettier --check .github/workflows/release-azure-containerapp-v1.yaml .github/workflows/_release-docker-dx-metrics-v1.yaml .github/workflows/_release-aws-mcp-server-deploy-v1.yaml .github/workflows/_release-aws-mcp-server-deploy-dev-v1.yaml`
- `git diff --check`
- VS Code YAML diagnostics


.github/workflows/_release-aws-mcp-server-deploy-dev-v1.yaml
```diff
@@ -23,37 +24,60 @@ jobs:
...
+          image_tag="${RELEASE_TAG##*@}"
+          if [[ -z "$image_tag" ]]; then
+            echo "::error::A release image tag is required."
+            exit 1
+          fi
+
+          echo "image_tag=$image_tag" >> $GITHUB_OUTPUT
+
+      - name: Log In to GitHub Container Registry
+        uses: docker/login-action@4907a6ddec9925e35a0a9e82d7399ccc52663121 # v4.1.0
```
kin0992: **note**: latest available is 4.4.0

.github/workflows/_release-aws-mcp-server-deploy-dev-v1.yaml
```diff
@@ -23,37 +24,60 @@ jobs:
...
-          dockerfile_path: apps/mcpserver/Dockerfile
-          dockerfile_context: "."
-          docker_image_description: "DX MCP Server"
-          docker_image_authors: "DevEx"
-          docker_image_name: ${{ env.IMAGE_NAME }}
-          build_platforms: linux/arm64
-          push_to_registry: true
+      - name: Log In to Amazon ECR
+        id: ecr
+        uses: aws-actions/amazon-ecr-login@f2e9fc6c2b355c1890b65e6f6f0e2ac3e6e22f78 # v2.1.2
```
kin0992: **note**: latest available is 2.1.6

.github/workflows/release-azure-containerapp-v1.yaml
```diff
@@ -18,7 +18,7 @@ on:
         description: Docker image name, defaults to repository name
       docker_image_description:
         type: string
-        required: true
+        required: false
```
krusty93, line 21: question: why?

christian-calabrese: Because the Docker image description is only needed when building a new image. If the workflow is triggered with an existing image, that step is skipped, so the description isn’t required.

.github/workflows/release-azure-containerapp-v1.yaml
```diff
@@ -103,21 +109,40 @@ jobs:
...
-          set -euo
+          set -euo pipefail
 
           sha=$(git rev-parse --short "$GIT_SHA")
 
           echo "::notice::Latest commit: $sha"
-
           echo "short_sha=$sha" >> $GITHUB_OUTPUT
 
+      - name: Resolve Deployment Image
```
mamu0, line 128: [question](https://pullpo.io/cc?l=question) 
This workflow is also referenced by other repositories.
Are these changes backward compatible, meaning it will continue to work as expected for existing users without requiring updates to their configuration? Has this been tested?

christian-calabrese: You're right! Here are test runs:
- legacy way (without passing the already built image tag):  https://github.com/pagopa/dx/actions/runs/29922663977/job/88932395152
- new way (passing the tag): https://github.com/pagopa/dx/actions/runs/29920272735

mamu0: Great!
However, it's not clear to me whether this is actually backward compatible.
For example, the `image_tag` input was not required before, so we should verify whether all repositories using this workflow already have that input set; otherwise, they need to be notified.

If it helps, an alternative would be to create a new **v2** workflow and instruct the teams to update to the new configuration.

christian-calabrese: Yes this is backward compatible. Can you comment obscure parts where you think it may not be backward compatible?

mamu0: Got it, I misread one of the changes initially, but taking another look now it all looks good. Plus, since you've already tested it, I don't see any problems with this.

mamu0 (approved):

PR #1987, opened 2026-07-15
Add new GitHub App config for Terraform publish

mamu0:
This pull request introduces support for a dedicated GitHub App installation token specifically for Terraform publishing tasks. It updates the workflows, action inputs, and environment variables to use this new token, ensuring Terraform publishing is authenticated with the correct credentials. Additionally, the `nx-release` action and the `nx-terraform-plugin` package are updated to recognize and use the new token.

Resolves: CES-2218

### DIFF
--- a/.github/workflows/release-v2.yaml
+++ b/.github/workflows/release-v2.yaml
@@ -15,6 +15,12 @@ on:
       app-private-key:
         description: "GitHub App private key for authentication"
         required: true
+      app-tf-client-id:
+        description: "GitHub App Client ID for authentication for Terraform publish"
+        required: true
+      app-tf-private-key:
+        description: "GitHub App private key for authentication for Terraform publish"
+        required: true
 
 jobs:
   release:
@@ -32,6 +38,15 @@ jobs:
         with:
           client-id: ${{ secrets.app-client-id }}
           private-key: ${{ secrets.app-private-key }}
+          permission-contents: write
+          permission-pull-requests: write
+
+      - name: Generate GitHub App token for Terraform publish
+        id: generate-terraform-token
+        uses: actions/create-github-app-token@1b10c78c7865c340bc4f6099eb2f838309f1e8c3 # 3.1.1
+        with:
+          client-id: ${{ secrets.app-tf-client-id }}
+          private-key: ${{ secrets.app-tf-private-key }}
           permission-contents: write
           permission-pull-requests: write
 
@@ -81,3 +96,4 @@ jobs:
         with:
           app-slug: ${{ steps.generate-token.outputs.app-slug }}
           github-token: ${{ steps.generate-token.outputs.token }}
+          github-terraform-token: ${{ steps.generate-terraform-token.outputs.token }}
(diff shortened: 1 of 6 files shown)


.github/workflows/release-v2.yaml
```diff
@@ -15,6 +15,12 @@ on:
       app-private-key:
         description: "GitHub App private key for authentication"
         required: true
+      app-tf-client-id:
```
gunzip, line 18: **issue**: we're polluting this reusable workflow with inputs that are solely useful for us as DX team, as we operate on another repository. Can we try another design in order to keep this wf general purpose?

gunzip (changes requested):

actions/nx-release/action.yaml
```diff
@@ -8,6 +8,9 @@ inputs:
   app-slug:
     description: "GitHub App slug associated with github-token"
     required: true
+  github-terraform-token:
```
gunzip, line 11: ditto. public actions and workflow shouldn't be aware that this token is used for 'terraform publish' (which is a deceptive phrase anyway, since users may interpret it as 'iac publish')

gunzip (changes requested):

.github/workflows/release-v2.yaml
mamu0: The cleanest alternative is to avoid creating a new GitHub App and directly install the existing one from this organization into pagopa-dx, preventing any duplication.

christian-calabrese: Alternatively (more complex) we could make the existing inputs accept both a simple string or a json (something like <org_name1>=token1, <org_name2>=token2)

actions/nx-release/action.yaml
```diff
@@ -8,6 +8,9 @@ inputs:
...
+    description: "Pre-generated GitHub App installation token for Terraform publish"
+    required: true
```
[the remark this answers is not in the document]
mamu0: done [41ad0ff](https://github.com/pagopa/dx/pull/1987/commits/41ad0fff9eba5b7ce4239a9006acd2f59d50671d)

.github/workflows/release-v2.yaml
gunzip: yes, better! example's not a json, but a comma separated list which is ok anyway.

mamu0: [thought](https://pullpo.io/cc?l=thought) 
Keep in mind that this increases complexity.
Using the same two variables (`app-client-id` and `app-private-key`) without JSON means we must keep both lists in the exact same order and, regardless of the format, split them manually (e.g., via `bash`).
Additionally, the token retrieval step would need to run as a matrix, and, currently, this complex setup would only serve our specific dual-organization case (where one publishes Terraform modules).

This might confuse standard users, as we would still need to document this multi-app behavior for everyone.

To sum up, I don't oppose @christian-calabrese 's suggestion, but removing an input this way creates a more fragile system that is harder to maintain.

gunzip: Fair point, but we still need to think of a better design, as hiding exceptions inside public interfaces is unacceptable.

mamu0: We are trying the first option, which is to use the same pagopa GitHub App for pagopa-dx as well. If this works as expected, we can close this PR.

PR #1992, opened 2026-07-16
Configure alternative owner for Terraform resources

mamu0:
Introduce functionality to specify an alternative owner for Terraform-managed resources, enhancing resource management and ownership clarity.

Resolves: CES-2218

### DIFF
--- a/packages/nx-terraform-plugin/src/adapters/github/octokit.ts
+++ b/packages/nx-terraform-plugin/src/adapters/github/octokit.ts
@@ -1,9 +1,46 @@
 // Provides the concrete Octokit-backed GitHub repository adapter for publish flows.
 
+import { createAppAuth } from "@octokit/auth-app";
 import { Octokit } from "octokit";
 
-export const getGitHubToken = (): string | undefined =>
-  process.env.GH_TOKEN ?? process.env.GITHUB_TOKEN;
+export interface GitHubAppCredentials {
+  clientId: string;
+  privateKey: string;
+}
+
+export const createGitHubAppToken = async (
+  owner: string,
+  credentials: GitHubAppCredentials,
+): Promise<string> => {
+  const appOctokit = new Octokit({
+    auth: {
+      appId: credentials.clientId,
+      privateKey: credentials.privateKey,
+    },
+    authStrategy: createAppAuth,
+  });
+  const installation = await appOctokit.rest.apps.getOrgInstallation({
+    org: owner,
+  });
+  const auth = createAppAuth({
+    appId: credentials.clientId,
+    installationId: installation.data.id,
+    privateKey: credentials.privateKey,
+  });
+  const authentication = await auth({
+    permissions: {
+      contents: "write",
+    },
+    type: "installation",
+  });
+
+  return authentication.token;
+};
+
+export const revokeGitHubAppToken = async (token: string): Promise<void> => {
+  const octokit = new Octokit({ auth: token });
+  await octokit.rest.apps.revokeInstallationAccessToken();
+};
 
 const getAuthenticatedUserLogin = async (
   octokit: Octokit,
@@ -23,9 +60,10 @@ const getAuthenticatedUserLogin = async (
 export const ensureGitHubRepository = async (
   owner: string,
   repo: string,
+  token: string,
 ): Promise<void> => {
   const octokit = new Octokit({
-    auth: getGitHubToken(),
+    auth: token,
   });
 
   try {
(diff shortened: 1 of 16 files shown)


christian-calabrese (approved):

packages/nx-terraform-plugin/src/adapters/github/octokit.ts
```diff
@@ -1,9 +1,46 @@
...
+export interface GitHubAppCredentials {
+  clientId: string;
+  privateKey: string;
+}
+
+export const createGitHubAppToken = async (
+  owner: string,
+  credentials: GitHubAppCredentials,
+): Promise<string> => {
+  const appOctokit = new Octokit({
```
gunzip, line 15: suggestion: a better design would be to accept an instance of Octokit as argument. in this way, if you have to call this method multiple times, you avoid to re-create the instance every time and you can share it with other methods that need to call github APIs

packages/nx-terraform-plugin/src/adapters/github/publisher.ts
```diff
@@ -156,26 +168,35 @@ export const publishToGithub = async (
...
 
-  if (publishError !== undefined) {
-    throw publishError;
+  let revokeError: unknown;
+  if (input.githubAppCredentials !== undefined) {
+    try {
+      // Installation tokens are temporary, but revoking them minimizes their
+      // usable lifetime even when publishing failed or was skipped.
+      await revokeGitHubAppToken(token);
+    } catch (error) {
```
gunzip, line 191: suggestion: just let `revokeGitHubAppToken` throw here

packages/nx-terraform-plugin/src/adapters/github/octokit.ts
```diff
@@ -1,9 +1,46 @@
...
 import { Octokit } from "octokit";
 
-export const getGitHubToken = (): string | undefined =>
-  process.env.GH_TOKEN ?? process.env.GITHUB_TOKEN;
+export interface GitHubAppCredentials {
+  clientId: string;
+  privateKey: string;
+}
+
+export const createGitHubAppToken = async (
```
gunzip, line 11: suggestion: add a comment with the purpose of this method

packages/nx-terraform-plugin/src/adapters/github/octokit.ts
```diff
@@ -1,9 +1,46 @@
...
+      contents: "write",
+    },
+    type: "installation",
+  });
+
+  return authentication.token;
+};
+
+export const revokeGitHubAppToken = async (token: string): Promise<void> => {
+  const octokit = new Octokit({ auth: token });
```
gunzip, line 41: suggestion: this is the third instance of a github client (another one is being created at line 40). is creating a new instance necessary? see comment above about sharing octokit instances

packages/nx-terraform-plugin/src/executors/publish/schema.ts
```diff
@@ -2,18 +2,45 @@ import { z } from "zod/v4";
...
+    .min(1)
+    .transform((privateKey) => privateKey.replaceAll("\\n", "\n")),
+});
+
+export const githubTokenEnvironmentSchema = z
+  .object({
+    GH_TOKEN: z.string().min(1).optional(),
+    GITHUB_TOKEN: z.string().min(1).optional(),
+  })
+  .transform((environment, context) => {
```
gunzip, line 18: suggestion: use a zod xor across two different shapes:

https://zod.dev/api?id=exclusive-unions-xor#exclusive-unions-xor

packages/nx-terraform-plugin/src/executors/publish/publish.ts
```diff
@@ -31,6 +33,41 @@ const runExecutor: PromiseExecutor<NxReleasePublishExecutorInput> = async (
   }
 
   const validatedOptions = parseResult.data;
+  let githubAppCredentials:
+    | undefined
+    | { clientId: string; privateKey: string };
+  let githubToken: string;
+
+  if (validatedOptions.useGitHubAppAuthentication) {
```
gunzip, line 41: suggestion: we can have one zod schema for validatedOptions that embed an environmentParseResultfield that is either githubAppEnvironmentSchema or githubTokenEnvironmentSchema according to the value of useGitHubAppAuthentication. Same for githubAppCredentials and githubToken. You can describe the whole structure with only one zod schema. In this way you can only call safeParse once without manually handling conditionals.

PR #2002, opened 2026-07-20
Add DX cli support to multi-tenant environments

christian-calabrese:
## Summary

- Allow `dx add environment` to accept tenant-qualified names such as `ced-prod` and `cgn-dev`.
- Derive Azure short environment codes from the lifecycle suffix while keeping generated folders and GitHub environments tenant-qualified.
- Ensure repository-specific bootstrapper OIDC federated credentials and GitHub environment secrets are configured even when Azure bootstrap resources already exist.
- Document the multi-tenant workflow in the website and update the DX CLI companion skill for future agents.

Resolves CES-1753


christian-calabrese:
image

apps/cli/src/domain/environment.ts
```diff
@@ -7,18 +7,34 @@ import {
...
+const environmentNameSchema = z
+  .string()
+  .trim()
+  .toLowerCase()
+  // Keep the lifecycle suffix explicit so tenant-qualified names such as
+  // `ced-prod` can still reuse the same Azure short code as `prod`.
+  .regex(
+    /^(?:[a-z0-9]+(?:-[a-z0-9]+)*-)?(?:dev|prod|uat)$/,
+    "Environment name must be dev, prod, uat, or a tenant-qualified name ending with -dev, -prod, or -uat",
+  );
```
kin0992, line 40: suggestion: since it looks that we have a strong pattern here (`<something>-<env>` or just `<env>`), I would investigate the use of `z.templateLiterals`(https://zod.dev/api?id=template-literals).

I would do something like this (do not accept as it is, this is just a test)
```typescript
const envSchema = z.enum(["dev", "prod", "uat"]);

export const environmentNameSchema = z
  .union([
    envSchema,
    z.templateLiteral([z.string().lowercase(), z.literal("-"), envSchema]),
  ])
  .transform((val) => val.trim().toLowerCase()); // THIS MAY BE REDUNDANT

// Inferred type will strictly be:
// "dev" | "prod" | "uat" | `${string}-dev` | `${string}-prod` | `${string}-uat`
export type EnvironmentName = z.infer<typeof environmentNameSchema>;
```

apps/cli/src/adapters/azure/cloud-account-service.ts
```diff
@@ -827,6 +889,36 @@ export class AzureCloudAccountService implements CloudAccountService {
     return `/subscriptions/${subscriptionId}/providers/Microsoft.Authorization/roleDefinitions/${roleDefinitionId}`;
   }
 
+  async #getBootstrapperIdentity({
+    cloudAccountId,
+    identityName,
+    msiClient,
+    resourceGroupName,
+  }: {
+    cloudAccountId: string;
```
kin0992: suggestion: 
```suggestion
    cloudAccountId: CloudAccount["id"];
```

apps/cli/src/adapters/azure/cloud-account-service.ts
```diff
@@ -133,12 +133,74 @@ export class AzureCloudAccountService implements CloudAccountService {
...
+    assert.ok(
+      isAzureLocation(cloudAccount.defaultLocation),
+      "The default location of the cloud account is not a valid Azure location",
+    );
+
+    const short = {
+      env: getEnvironmentShort(name),
+      location: locationShort[cloudAccount.defaultLocation],
+    };
+    const resourceGroupName = `${prefix}-${short.env}-${short.location}-common-rg-01`;
```
kin0992: **note**: to avoid typos and type errors, you could type the ResourceGroupName as already done for the BootstrapperIdentityName.

Right now a string `aString` would be accepted, but the resource group has some format requirements.

apps/cli/src/domain/environment.ts
christian-calabrese: Addressed in bdd8dcbb: the environment name schema now uses `z.templateLiteral` for tenant-qualified names while keeping trim/lowercase normalization before validation.

apps/cli/src/adapters/azure/cloud-account-service.ts
christian-calabrese: Addressed in bdd8dcbb: `cloudAccountId` now uses `CloudAccount["id"]` in the touched bootstrap helper signatures.

apps/cli/src/adapters/azure/cloud-account-service.ts
christian-calabrese: Addressed in bdd8dcbb: bootstrap resource group names now use a dedicated template-literal type matching the generated naming pattern.

apps/cli/src/adapters/commander/commands/add.ts
```diff
@@ -419,11 +419,10 @@ export const makeAddCommand = (
       new Command("environment")
         .description("Add a new deployment environment")
         .addOption(
-          new Option("--name <name>", "Environment name").choices([
-            "dev",
-            "uat",
-            "prod",
-          ]),
+          new Option(
```
krusty93, line 422: **thought**: I'm not sure removing the choice is the right thing. Potentially, users can use everything they want and our framework would probably break. Maybe would it better to add another option like `add environment [--prefix cgn] --name prod`?

kin0992: I agree, I would prefer to keep the choice and handle it differently. Can we reuse the already existing `prefix` flag?

christian-calabrese: I’d avoid reusing `prefix` as the tenant identifier because it only drives Azure resource naming, while the tenant must also disambiguate repo-level surfaces such as folders, workflows, GitHub environments, and Terraform state.

The current implementation does not accept arbitrary values unchecked: Commander no longer enforces choices, but the domain schema still only accepts `dev|uat|prod` or `<tenant>-<env>`.

That said, if we want to preserve the `--name dev|uat|prod` contract, I think the cleaner alternative is adding a first-class `--tenant cgn` option and deriving the logical environment as `cgn-prod`, rather than overloading `--prefix`.

mamu0: What if we used a format similar to what the CLI already uses for the GitHub App?
We wouldn't add any new inputs, it would remain just `dev/uat/prod`, but the CLI could display a prompt like "Do you want to add a tenant? (y/n)" and, if confirmed, ask you to enter it.

krusty93: > The current implementation does not accept arbitrary values unchecked: Commander no longer enforces choices, but the domain schema still only accepts dev|uat|prod or <tenant>-<env>.

How did this information come back to the user/agent? Should we update the companion skill too?

btw @mamu0 proposed an interesting option

kin0992: **note**: to maximise agentic experience, we should not ask for confirmations, but instead pass a flag

kin0992 (approved):

PR #2004, opened 2026-07-21
Fix Terraform plan output truncation in CI

mamu0:
This PR updates the Infrastructure Review workflow to run Terraform-related Nx tasks with `--outputStyle=stream-without-prefixes`.
In our CI context, Nx grouped/static output was causing Terraform plan logs to appear truncated for large environments.

With this change:

- full Terraform plan output is preserved in step logs;
- noisy per-line project prefixes are removed;
- no custom plan truncation or GitHub Step Summary fallback is needed in the dx-tasks plugin.

Resolves: CES-2197

### DIFF
--- a/packages/dx-tasks/src/terraform/plan.ts
+++ b/packages/dx-tasks/src/terraform/plan.ts
@@ -1,5 +1,6 @@
 /** This module runs Terraform plans for dx-tasks without external process helpers. */
 
+import fs from "node:fs/promises";
 import util from "node:util";
 import * as z from "zod/mini";
 
@@ -60,6 +61,51 @@ export interface TerraformPlanReport {
 
 const noPlanOutputMessage = "No plan output available.";
 const terraformPlanReportSeparator = "\n\n";
+
+// CI step logs can truncate output for large plans; this limit keeps each plan's
+// console output well within safe bounds while leaving room for task runner overhead.
+const PLAN_CONSOLE_MAX_CHARS = 50_000;
+
+// Appends the full Terraform plan output to a summary file as a collapsible
+// <details> block. No-ops when summaryFilePath is not provided.
+// Errors are swallowed since this is best-effort; the report artifact and console log remain available.
+export const appendPlanOutputToSummary = async (
+  summaryFilePath: string | undefined,
+  modulePath: string,
+  strippedOutput: string,
+): Promise<void> => {
+  if (!summaryFilePath) {
+    return;
+  }
+
+  const markdown = `### Terraform Plan: \`${modulePath}\`\n\n<details><summary>Show Plan</summary>\n\n\`\`\`\n${strippedOutput}\n\`\`\`\n\n</details>\n\n`;
+
+  try {
+    await fs.appendFile(summaryFilePath, markdown, "utf8");
+  } catch (e) {
+    console.warn("Failed to write Terraform plan output summary", e);
+  }
+};
+
+// Truncates plan output to maxChars for console logging. When the output exceeds the limit,
+// replaces it with the summary line and a notice pointing users to the output summary file.
+export const truncateForConsoleLog = (
+  output: string,
+  summaryLine: string | undefined,
+  maxChars: number,
+  shouldTruncate: boolean,
+  hasSummaryFile: boolean,
+): string => {
+  if (!shouldTruncate || output.length <= maxChars) {
+    return output;
+  }
+
+  const truncationTarget = hasSummaryFile
+    ? "See the plan output summary for the full output."
+    : "See the plan report artifacts for the full output.";
+
+  return `${summaryLine ?? noPlanOutputMessage}\n\n[Plan output truncated. ${truncationTarget}]`;
+};
 
 const renderTerraformPlanReports = (
   reports: readonly TerraformPlanReport[],
@@ -232,6 +278,8 @@ const executeTerraformPlan = async (
   env: Record<string, string>,
   verbose: boolean,
   report: boolean,
+  summaryFilePath: string | undefined,
+  shouldTruncate: boolean,
   context: TaskRunContext,
 ) => {
   const result = await runCommand("terraform", ["plan"], modulePath, env);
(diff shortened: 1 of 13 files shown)


krusty93:
question: do you have the link to the run showed in the screenshot?

mamu0:
> question: do you have the link to the run showed in the screenshot?

@krusty93  You can take a look [here](https://github.com/pagopa/dx/actions/runs/29817938740/job/88593598688#step:11:159)

krusty93:
> > question: do you have the link to the run showed in the screenshot?
> 
> @krusty93 You can take a look [here](https://github.com/pagopa/dx/actions/runs/29817938740/job/88593598688#step:11:159)

Does the summary have any similar limit? Perhaps it can be solved by disabling NX groups? I sincerely don't know

mamu0:
> Does the summary have any similar limit? Perhaps it can be solved by disabling NX groups? I sincerely don't know

@krusty93 What do you mean by summary limit?
This PR addresses the issue where the terraform plan done in `validate_v2.yaml` was being truncated in the step logs. This happens because each step has a maximum limit of 65k characters, and large plans were exceeding this threshold.
For example [this](https://github.com/pagopa/dx/actions/runs/29814866002/job/88583802334#step:11:880) is an execution without the PR fix.

packages/dx-tasks/src/terraform/plan.ts
```diff
@@ -61,6 +62,51 @@ export interface TerraformPlanReport {
...
+// console output well within safe bounds while leaving room for task runner overhead.
+const PLAN_CONSOLE_MAX_CHARS = 50_000;
+
+// Appends the full Terraform plan output to a summary file as a collapsible
+// <details> block. No-ops when summaryFilePath is not provided.
+// Errors are swallowed since this is best-effort; the report artifact and console log remain available.
+export const appendPlanOutputToSummary = async (
+  summaryFilePath: string | undefined,
+  modulePath: string,
+  strippedOutput: string,
```
kin0992: **nitpick**: I would call this argument just `output` or in general I would not prepend the `stripped`prefix.

the function scope is to append something to a file, so it can be a stripped output or a simple string

kin0992: for the same reason, I would not accept an undefined value for `summaryFilePath`

in general this function aims to write some content to a file. so its scope should be just that; if the file does not exist, the `appendFile` will fail.

so, if the summaryFilePath may be undefined for some Action runs, then I would handle it somewhere else.
The caller should handle it in my opinion

```typescript
// some logic before
if (summaryFilePath) {
// filePath provided, then append to a file
    appendPlanOutputToSummary(file, module, output);
} else {
    // do something else, if needed
    // or do nothing
}
```

Please note that the code provided is just an example

packages/dx-tasks/src/terraform/plan.ts
```diff
@@ -61,6 +62,51 @@ export interface TerraformPlanReport {
...
+export const truncateForConsoleLog = (
+  output: string,
+  summaryLine: string | undefined,
+  maxChars: number,
+  shouldTruncate: boolean,
+  hasSummaryFile: boolean,
+): string => {
+  if (!shouldTruncate || output.length <= maxChars) {
+    return output;
+  }
```
kin0992: **note**: same as a previous comment

I would not handle the logic for truncating here. I would assume this function just returns a truncated output (as the function's name suggests)

packages/dx-tasks/src/terraform/plan.ts
```diff
@@ -232,6 +278,8 @@ const executeTerraformPlan = async (
   env: Record<string, string>,
   verbose: boolean,
   report: boolean,
+  summaryFilePath: string | undefined,
+  shouldTruncate: boolean,
```
kin0992: **suggestion**: I would rename the `shouldTruncate` in something different

If you call it `shouldTruncate`, I would expect to have some code like
```typescript
if (shouldTruncate)
    truncateOutput();
```
and the `truncateOutput` would just return the truncated output.
But looking at the implementation, the output is going to be truncated not only when the `shouldTruncate` condition is `true`, so naming is misleading

packages/dx-tasks/src/terraform/plan.ts
mamu0: thanks for the suggestion, i've done this here [13d674c](https://github.com/pagopa/dx/pull/2004/commits/13d674c1ce13ce24a920680df92e5d5daab2d0b2)

packages/dx-tasks/src/terraform/plan.ts
mamu0: done here [13d674c](https://github.com/pagopa/dx/pull/2004/commits/13d674c1ce13ce24a920680df92e5d5daab2d0b2)

packages/dx-tasks/src/terraform/plan.ts
mamu0: thanks for the suggestion, i've renamed that here [13d674c](https://github.com/pagopa/dx/pull/2004/commits/13d674c1ce13ce24a920680df92e5d5daab2d0b2)

krusty93:
> > Does the summary have any similar limit? Perhaps it can be solved by disabling NX groups? I sincerely don't know
> 
> @krusty93 What do you mean by summary limit? This PR addresses the issue where the terraform plan done in `validate_v2.yaml` was being truncated in the step logs. This happens because each step has a maximum limit of 65k characters, and large plans were exceeding this threshold. For example [this](https://github.com/pagopa/dx/actions/runs/29814866002/job/88583802334#step:11:880) is an execution without the PR fix.

I mean the [summary](https://github.com/pagopa/dx/actions/runs/29907654887#summary-88882888777) where this PR is moving the output, perhaps it does have a similar limit as the PR comments?

About the example run you linked, it seems very short to hit a limit. We saw several times runs with bigger outputs there were not truncated. For example, [this one is way longer](https://github.com/pagopa/io-infra/actions/runs/29905950107/job/88877843773?pr=1964) than the example (227k characters vs 65k). Something is hinting the limit is imposed by the NX sub-groups instead

krusty93 (changes requested):
Something is not right

https://github.com/pagopa/dx/pull/2004#issuecomment-5047534067

mamu0:
> I mean the [summary](https://github.com/pagopa/dx/actions/runs/29907654887#summary-88882888777) where this PR is moving the output, perhaps it does have a similar limit as the PR comments?
> 
> About the example run you linked, it seems very short to hit a limit. We saw several times runs with bigger outputs there were not truncated. For example, [this one is way longer](https://github.com/pagopa/io-infra/actions/runs/29905950107/job/88877843773?pr=1964) than the example (227k characters vs 65k). Something is hinting the limit is imposed by the NX sub-groups instead

@krusty93 Great point, thanks!
In this commit [87bea3f](https://github.com/pagopa/dx/pull/2004/commits/87bea3fd109e293c7b30343652d1867d1641d09c) I've removed the `GITHUB_STEP_SUMMARY` workaround and the 50k truncation logic from the Terraform task, so we now emit the full plan output again.
I also switched nx affected to `--outputStyle=stream` in `validate-v2.yaml` to bypass Nx grouped/static output handling in CI, which looks like the likely source of the truncation behavior we were seeing.

Check [this](https://github.com/pagopa/dx/actions/runs/29933171337/job/88967910417?pr=2004#step:11:1298) as result example

krusty93 (approved):

PR #2008, opened 2026-07-23
Fix Docker release publishing

christian-calabrese:
Fixes the Docker release path used by Nx Release.

- authenticate GHCR publishing with the workflow token and required Packages permission
- use Buildx annotations compatible with single-platform images
- add retry handling for transient Azure CLI installation failures in the self-hosted runner image
- support durable release versions for Docker-only Nx projects
- remove temporary verbose output from Nx Release commands

Validation:
- focused Docker plugin tests
- Nx release and publisher dry-runs for `self-hosted-runner`
- workflow and Dockerfile diagnostics


christian-calabrese:
Successful run
https://github.com/pagopa/dx/actions/runs/29991131733

packages/nx-dx-docker-plugin/src/executors/release-publish/release-publish.ts
```diff
@@ -1,138 +1,116 @@
...
     );
-    return { success: false };
+    if (!parseResult.success) {
+      throw new Error(`Could not read a version from ${packageJsonPath}.`, {
+        cause: parseResult.error,
+      });
+    }
+
+    return {
+      sourcePath: `${projectRoot}/package.json`,
```
kin0992, line 53: **question**: isn't this `packageJsonPath` instead?

packages/nx-dx-docker-plugin/src/executors/release-publish/release-publish.ts
```diff
@@ -1,138 +1,116 @@
...
-    return { success: false };
   }
 
-  const fullImageRef = readFileSync(versionFilePath, "utf8").trim();
-  const { imageBase, version } = splitImageReference(fullImageRef);
-  const aliasTags = computeReleaseTags(projectName, version).filter(
-    (tag) => tag !== version,
-  );
+  return {
+    sourcePath: `${projectRoot}/project.json`,
```
kin0992, line 74: same as previous comment

packages/nx-dx-docker-plugin/src/executors/release-publish/release-publish.ts
```diff
@@ -1,138 +1,116 @@
...
+    version: parseResult.data.metadata.version,
+  };
+};
+
+export const releasePublishExecutor = async (
```
kin0992, line 79: **question**: what is the goal of this function?

I see it is used in test files only and it is just exported, never called anywhere. Who is gonna use this function? When?

packages/nx-dx-docker-plugin/src/release/version-actions.ts
```diff
@@ -0,0 +1,99 @@
...
+
+/**
+ * Implements Nx Release for container-only projects that have no package
+ * manifest. The version lives in project.json metadata.version so it can be
+ * updated by nx release and consumed later by the publishing executor.
+ */
+export default class DockerProjectVersionActions extends VersionActions {
+  validManifestFilenames = ["project.json"];
+
+  async readCurrentVersionFromRegistry(): Promise<null> {
```
kin0992, line 45: **question**: what is the scope of this function? why is it useful?

packages/nx-dx-docker-plugin/src/release/version-actions.ts
```diff
@@ -0,0 +1,99 @@
...
+      .passthrough(),
+  })
+  .passthrough();
+
+/**
+ * Implements Nx Release for container-only projects that have no package
+ * manifest. The version lives in project.json metadata.version so it can be
+ * updated by nx release and consumed later by the publishing executor.
+ */
+export default class DockerProjectVersionActions extends VersionActions {
```
kin0992, line 41: **thought**: I see a lot of functions, under this class, that are not implemented (or just returns `null`. 
it is not clear to me what's going on here

packages/nx-dx-docker-plugin/src/executors/release-publish/release-publish.ts
christian-calabrese: `packageJsonPath` is the absolute path used to read the manifest. This value is intentionally workspace-relative because it is only used in log and error messages; that avoids exposing runner-specific absolute paths. The package manifest is therefore read from `packageJsonPath`, while the message reports `apps/.../package.json`.

packages/nx-dx-docker-plugin/src/executors/release-publish/release-publish.ts
christian-calabrese: Same rationale here: `projectJsonPath` is the absolute path used for I/O, while the returned path is workspace-relative and intended only for diagnostics.

packages/nx-dx-docker-plugin/src/executors/release-publish/release-publish.ts
christian-calabrese: This is the Nx executor entry point. Nx resolves it dynamically through `executors.json` when `nx release publish` invokes the inferred `nx-release-publish` target. It is exported so Nx can load it, and tests import it directly to validate the executor behavior without running Nx end to end.

packages/nx-dx-docker-plugin/src/release/version-actions.ts
christian-calabrese: This is the custom `VersionActions` implementation used only by Docker-only projects, which do not have a `package.json`. Its scope is to make `project.json.metadata.version` the durable version source for Nx Release: versioning updates that field and the publish executor reads it later, including when version and publish run on different CI runners.

packages/nx-dx-docker-plugin/src/release/version-actions.ts
christian-calabrese: `VersionActions` is an abstract Nx Release contract, so these methods must be implemented even when a capability does not apply. For Docker-only projects, `metadata.version` is the only version source: there is no package registry lookup, no dependency version manifest, and no dependency versions to rewrite. The methods returning `null` or `[]` are therefore intentional no-ops; only reading and updating `project.json.metadata.version` are relevant to this project type.

actions/nx-release/action.yaml
```diff
@@ -156,6 +156,14 @@ runs:
         # Create or update PR (script writes to GITHUB_OUTPUT)
         RELEASE_TAGS="$RELEASE_TAGS" node "$ACTION_PATH/scripts/dist/manage-version-pr.js"
 
+    - name: Log in to GitHub Container Registry
+      if: steps.release-mode.outputs.mode == 'publish' || steps.release-mode.outputs.mode == 'publish-all'
+      uses: docker/login-action@4907a6ddec9925e35a0a9e82d7399ccc52663121 # v4.1.0
```
mamu0: [suggestion(non-blocking)](https://pullpo.io/cc?l=suggestion&d=non-blocking) 

```suggestion
      uses: docker/login-action@06fb636fac595d6fb4b28a5dfcb21a6f5091859c # v4.5.0
```

kin0992 (approved):
you need to rebase and fix conflicts

mamu0 (approved):

PR #2023, opened 2026-07-29
Prevent legacy Terraform scanner references

kin0992:
## Summary
- Add a pre-commit guard that rejects new tfsec references.
- Keep the guard itself excluded from the scan because it contains the detection pattern.

## Validation
- ShellCheck passed.
- The guard correctly failed on a temporary legacy annotation and passed after cleanup.

Depends on #2022.

### DIFF
--- a/infra/scripts/check-legacy-scanner-reference.sh
+++ b/infra/scripts/check-legacy-scanner-reference.sh
@@ -3,7 +3,7 @@
 set -euo pipefail
 
 if matches=$(git grep -n -i -e 'tfsec' -- . ':(exclude)infra/scripts/check-legacy-scanner-reference.sh'); then
-  echo "Legacy Terraform scanner references were found:"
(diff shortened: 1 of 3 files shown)


infra/scripts/check-legacy-scanner-reference.sh
```diff
@@ -0,0 +1,9 @@
+#!/usr/bin/env bash
+
+set -euo pipefail
+
+if matches=$(git grep -n -i -e 'tfsec' -- . ':(exclude)infra/scripts/check-legacy-scanner-reference.sh'); then
+  echo "Legacy Terraform scanner references were found:"
```
lucacavallaro: suggestion: Clarify that the output below contains the matching references.

```suggestion
  echo "Deprecated tfsec references detected. Please use Trivy instead. Matches found:"
```

lucacavallaro (approved):

PR #2026, opened 2026-07-29
Deduplicate findings and separate governance from cost

mamu0:
## What
Unifies findings from AZQR, Azure Advisor and the live-scan custom analyzers on a `resourceId + recommendationId` key, and separates governance/diagnostic observations from actual billable waste.

## Why
The same waste (e.g. an unattached disk) could be reported once per source, inflating the report and making the FinOps signal noisy. Governance findings (missing tags, wrong location, ...) were also tagged `category: "cost"`, suggesting a "saving" that doesn't exist and risking false dedup matches.

## How
- `Finding` gains an optional `recommendationId`, normalised onto a shared vocabulary: `orphan.<resourceType>` for orphaned resources (any source),  `advisor.<recommendationTypeId>`, `azqr.<recommendationId>`.
- New `azure/finding-dedup.ts`: `foldDuplicateFinding` collapses an incoming AZQR finding onto an already-collected one for the same resource, keeping max severity + the only available monetary estimate, and annotating the corroborating source (`Also flagged by azqr as an orphaned resource.`). Applied during `ingestAzqrReport`. Findings without a `recommendationId` (all custom sentence-level findings today) never participate - no guesswork dedup.
- New `azure/finding-category.ts`: custom analyzer sentences about missing tags, unpreferred location, unsupported resource type, or a failed detail lookup are now `operationalExcellence` instead of `cost`.
- The Azure Advisor Cost query (`Category eq 'Cost'`) is unchanged - it remains the only source of `estimatedMonthlySavings`.

## Testing
`pnpm nx run-many -t test typecheck lint build format:check -p @pagopa/dx-savemoney @pagopa/dx-cli` - all green, 186 tests in savemoney (+26).

Or local usage:

```bash
pnpm nx build @pagopa/dx-cli --skip-nx-cache
azqr scan --subscription-id 0000 --mask=false --json
node ./apps/cli/bin/index.js savemoney --azqr-report ./azqr_action_plan_xxxx_yy_zz_T123456.json --format lint > savemoney_result.md
```

Example result:
<kbd>
  image
</kbd>

Resolves: CES-2193

### DIFF
--- a/.nx/version-plans/version-plan-1785249132426.md
+++ b/.nx/version-plans/version-plan-1785249132426.md
@@ -1,0 +1,22 @@
+---
+'@pagopa/dx-savemoney': patch
+---
+
+Deduplicate findings across sources and separate governance from cost.
+
+Findings are now unified on a `resourceId + recommendationId` key: the new
+optional `Finding.recommendationId` normalises each source's native identifier
+onto a shared vocabulary (`orphan.<resourceType>` for orphaned resources,
+`advisor.<recommendationTypeId>` for Azure Advisor, `azqr.<recommendationId>`
+for the remaining AZQR rows). When an AZQR row collides with a finding already
+reported by the live scan or by Azure Advisor for the same resource, the row is
+dropped and the surviving finding keeps the highest severity and the only
+available monetary estimate, gaining a note that records the corroborating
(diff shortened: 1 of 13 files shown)


gunzip (approved):
very good!

PR #2029, opened 2026-07-31
Upgrade TypeScript

lucacavallaro:


packages/azure-tracing/package.json
```diff
@@ -102,13 +102,13 @@
     "eslint": "catalog:",
     "redis": "^5.12.1",
     "testcontainers": "^12.0.4",
-    "tsup": "catalog:",
+    "tsdown": "catalog:",
```
kin0992, line 105: **issue**: I think this is out of scope.

The PR should handle only the TS upgrade, not changing the bundler (which should be in a separate PR)

pnpm-workspace.yaml
```diff
@@ -56,7 +56,7 @@ catalog:
   tslib: ^2.8.1
   tsup: ^8.5.1
   tsx: ^4.23.1
-  typescript: ~5.9.3
+  typescript: ~6.0.3
```
kin0992, line 59: **question**: why not moving directly to version 7?

.nx/version-plans/version-plan-1785514869108.md
```diff
@@ -0,0 +1,5 @@
+---
+'@pagopa/azure-tracing': patch
+---
+
+Build the package with TypeScript 6-compatible tooling without changing its public API.
```
kin0992, line 5: **question**: is this true? is `tsup` not working with version 6?

kin0992: **note**: it looks to me that `tsup` continues to work with TS 6: https://github.com/pagopa/dx/actions/runs/30647189843/job/91211379188?pr=2029#step:9:980

packages/azure-tracing/package.json
lucacavallaro: I will check that, If I remember correctly, `tsup` does not work with latest typescript version due to compiler changes.

pnpm-workspace.yaml
lucacavallaro: @kin0992 version 7 is a complete rewrite in Go, that can break our toolchain and IDE-integration. Version `6.x` is a "migration release" meant to prepare the codebase to be ported on version 7 (which shares the same config file and same deprecations).

I prefer to do a gradual upgrade, in order to avoid breaking our tools.

PR #2048, opened 2026-08-05
Fix mise permissions in runner image

christian-calabrese:
## Summary

- grant the runner user write access to standard XDG tool directories without changing ownership elsewhere
- pre-install a pinned image baseline containing AWS CLI, Azure CLI, and zip, with uv as the Azure CLI backend dependency
- keep Node.js, Python, Go, and Terraform managed by their official workflow setup actions
- disable setup-action-owned runtimes in mise by default for every DX runner workflow
- resolve Node.js, pnpm, Python, and Terraform versions from their idiomatic project files
- configure Node.js once through `actions/setup-node` and cache the selected package manager after Corepack is enabled
- pin a known-good official runner image digest and reject images with an empty startup script or JavaScript Actions runtime
- remove redundant image stages and the dedicated APT package layer
- document the scoped writable directories and add a patch version plan

## Why

`Setup repository tools` failed because mise backends could not write to their expected directories. The image now grants narrowly scoped ownership instead of recursively changing the complete user-local tree.

The image baseline includes only common cloud tooling. The image also sets `MISE_DISABLE_TOOLS=go,node,pnpm,python,terraform`, preventing workflows from accidentally duplicating or shadowing runtimes installed by setup actions. Idiomatic version files keep local mise and workflow setup actions on the same runtime versions.

Local testing also exposed an incomplete upstream amd64 runner image whose `run.sh` and internal Node 24 binary were empty. Build-time smoke checks prevent publishing an image that can register with GitHub but cannot execute JavaScript actions.

## Validation

- built the self-hosted runner image through Nx
- executed AWS CLI, Azure CLI, uv, and zip directly from the image PATH
- confirmed Node.js, Python, and Terraform are absent from the image baseline
- confirmed the image policy skips Node.js, pnpm, Python, Go, and Terraform during repository mise installation
- confirmed local mise resolves Node.js from `.node-version`, pnpm from `package.json`, Python from `.python-version`, and Terraform from `.terraform-version`
- confirmed the Node.js setup action configures the requested runtime once and resolves the pnpm cache after Corepack is enabled
- completed the full Static Review flow on the candidate runner image during local end-to-end testing

Part of CES-2228


containers/self-hosted-runner/Dockerfile
```diff
@@ -11,55 +11,59 @@ ARG MISE_SHA256=961b1fcc78830e861ab887abd19d9b961478bcf252e37881fdd61c81388308d4
...
 ENV MISE_DATA_DIR=/home/runner/.local/share/mise
-ENV PATH="${MISE_DATA_DIR}/bin:${PATH}"
+ENV PATH="${MISE_DATA_DIR}/shims:${MISE_DATA_DIR}/bin:${PATH}"
 
 USER root
 RUN printf 'Acquire::Retries "5";\nAcquire::http::Timeout "30";\nAcquire::https::Timeout "30";\n' \
     > /etc/apt/apt.conf.d/80network-retries && \
     apt-get update && apt-get install -y --no-install-recommends \
-    zip wget xz-utils && \
+    zip wget && \
```
krusty93: question: Do we have this entire stage only to install zip and wget?

containers/self-hosted-runner/Dockerfile
```diff
@@ -11,55 +11,59 @@ ARG MISE_SHA256=961b1fcc78830e861ab887abd19d9b961478bcf252e37881fdd61c81388308d4
...
     apt-get clean && rm -rf /var/lib/apt/lists/*
 
 # === Stage 2: Install mise ===
 FROM base AS mise
 ARG MISE_VERSION
 ARG MISE_SHA256
-RUN mkdir -p "${MISE_DATA_DIR}/bin"
+RUN install -d -o runner -g runner -m 0755 \
```
krusty93, line 15: question: is there documentation for these paths?

containers/self-hosted-runner/Dockerfile
```diff
@@ -11,55 +11,59 @@ ARG MISE_SHA256=961b1fcc78830e861ab887abd19d9b961478bcf252e37881fdd61c81388308d4
...
 USER runner
 
+RUN cd /home/runner && \
+    mise trust mise.toml && \
+    mise install && \
+    openssl version && \
+    mise --version && \
+    mise exec -- aws --version && \
+    mise exec -- az version && \
+    mise exec -- node --version && \
```
krusty93: issue: Nod, python and Terraform should be installed via setup actions

christian-calabrese: Removed Node.js, Python, and Terraform from the image baseline. Static Review now uses MISE_DISABLE_TOOLS so CI skips Node.js, Python, Go, and Terraform while local mise installs the full toolchain; the workflow setup actions own those runtimes. Added setup-python and made node-setup bootstrap Node.js through actions/setup-node when the runner has no global Node.js.

containers/self-hosted-runner/Dockerfile
christian-calabrese: Documented every writable path and its consumer in the runner README. The Dockerfile now also explains why ownership is intentionally scoped to these directories.

containers/self-hosted-runner/Dockerfile
christian-calabrese: Collapsed the redundant base stage into the mise stage and removed the unused wget package. zip remains a runtime dependency because repository deployment and artifact actions invoke it directly.

mise.toml
```diff
@@ -11,7 +11,7 @@ jq = "1.8"
 node = { version = "24.14", postinstall = "corepack enable" }
 "npm:nx" = "22.7"
 pre-commit = "4.6"
-python = { version = "3.14.6", postinstall = "python -m ensurepip --upgrade" }
+python = { version = "{{ read_file(path='.python-version') | trim }}", postinstall = "python -m ensurepip --upgrade" }
```
krusty93: question: why the approach for python is different?

mise.ci.toml
lucacavallaro, line 1: **Issue:** This information should be hardcoded in the pipeline or in the runner image, because it will have the same content for all monorepos.

In general, we should introduce new configuration file to the monorepo skeleton only when we expect that there is a behavior that should be customized by our consumers. In this case, the behavior will be the same for all.

lucacavallaro (changes requested):

mise.toml
christian-calabrese: Already fixed!

mise.ci.toml
christian-calabrese: Implemented as an env variable in the dockerfile. This can be overridden inside a workflow if needed

.github/actions/node-setup/action.yaml
```diff
@@ -19,8 +19,24 @@ runs:
...
+      run: |
+        if ! command -v node >/dev/null || ! command -v corepack >/dev/null; then
+          echo "bootstrap=true" >> "$GITHUB_OUTPUT"
+        fi
+
+    - name: Bootstrap Node.js
+      if: ${{ steps.detect-node.outputs.bootstrap == 'true' }}
+      uses: actions/setup-node@48b55a011bda9f5d6aeb4c2d9c7362e8dae4041e # v6.4.0
+      with:
+        node-version-file: .node-version
```
lucacavallaro: **Suggestion:** `actions/setup-node` does more that just installing node, so it must always run.

I suggest to remove the `detect-node` and keep only the `Bootstrap Node.js` step

.github/workflows/validate-v2.yaml
```diff
@@ -98,6 +98,11 @@ jobs:
       - name: Setup Node.js
         uses: pagopa/dx/.github/actions/node-setup@main
 
+      - name: Setup Python
+        uses: actions/setup-python@5fda3b95a4ea91299a34e894583c3862153e4b97 # v7.0.0
+        with:
+          python-version-file: ".python-version"
```
lucacavallaro: **Thought:** we should find a way to install only the tool needed by the repository. 

For instance, `Go` and `Python` are used only by `pagopa/dx`. Other projects (like SelfCare) use `Java`, other requires only `Node.js`.

I think it is an issue that can be addressed later, before opening this workflow to others.

containers/self-hosted-runner/Dockerfile
```diff
@@ -11,55 +11,64 @@ ARG MISE_SHA256=961b1fcc78830e861ab887abd19d9b961478bcf252e37881fdd61c81388308d4
...
+        /home/runner/.local/bin \
+        /home/runner/.local/share/aube \
+        /home/runner/.local/share/gh \
+        /home/runner/.local/state/mise \
+        /home/runner/.config/mise \
+        /home/runner/.cache/mise \
+        /home/runner/.cache/node \
+        /home/runner/.cache/rosetta \
+        /home/runner/.cache/sigstore-rust \
+        /home/runner/.cache/uv
```
lucacavallaro: **Question:** can you please explain this change? Why it is needed?

Keep in mind that we don't know what tools `mise` will install, but there there are specific tool listed (`gh`, `uv`)

containers/self-hosted-runner/README.md
```diff
@@ -34,10 +44,48 @@ The Docker image includes an entrypoint script (`entrypoint.sh`) that is respons
...
+
+This prevents repository `mise.toml` files from duplicating or shadowing the
+runtimes installed by workflow setup actions. The policy is part of the image
+so every workflow using the DX runner receives the same safe default. A
+workflow that does not use those setup actions can explicitly opt back in:
+
+```yaml
+env:
+  MISE_DISABLE_TOOLS: ""
+```
```
lucacavallaro: **Nitpick:** we don't want/expect that users will customize this behavior because it may lead to bad practices (like installing node without properly configure the package manager cache)

So, I would remove this paragraph

lucacavallaro (changes requested):

.github/actions/node-setup/action.yaml
christian-calabrese: Done. `actions/setup-node` now always runs, so Node.js is configured consistently and the action can apply all of its setup behavior before Corepack and package-manager caching.

containers/self-hosted-runner/README.md
christian-calabrese: Done. I removed the override guidance from the README so the image policy remains the single expected behavior and runtimes continue to be configured through the setup actions.

.github/workflows/validate-v2.yaml
christian-calabrese: Agreed. I removed the unnecessary Python setup from Static Review. Go remains for the current DX validation targets; selecting setup actions from each repository’s actual requirements should be addressed separately before making this workflow generally available.

containers/self-hosted-runner/Dockerfile
christian-calabrese: The original list covered the paths observed for the current mise backends, but it could not support tools we do not know in advance. I replaced the tool-specific directories with the standard XDG roots (`~/.local/share`, `~/.local/state`, `~/.config`, and `~/.cache`, plus `~/.local/bin`). This keeps ownership scoped outside the rest of the home while allowing arbitrary repository mise backends to write to their conventional locations.

.github/actions/node-setup/action.yaml
```diff
@@ -19,8 +19,15 @@ runs:
   using: "composite"
 
   steps:
+    - name: Bootstrap Node.js
+      uses: actions/setup-node@48b55a011bda9f5d6aeb4c2d9c7362e8dae4041e # v6.4.0
+      with:
+        node-version-file: .node-version
```
lucacavallaro: **Issue:** `actions/setup-node` is called to times inside this action

I think this step should be removed

containers/self-hosted-runner/README.md
```diff
@@ -34,10 +45,39 @@ The Docker image includes an entrypoint script (`entrypoint.sh`) that is respons
...
 ## Testing
 
 The image includes basic tests to verify the installation of key tools:
 
-- Node.js (`node --version`)
-- npm (`npm --version`)
 - mise (`mise --version`)
+- AWS CLI (`aws --version`)
+- Azure CLI (`az version`)
+- uv (`uv --version`)
```
lucacavallaro: **Question:** why is `uv` pre-installed?

lucacavallaro (changes requested):

.github/actions/node-setup/action.yaml
lucacavallaro: sorry I inverted the action names by mistake, the one to be removed was `Bootstrap Node.js` and not the other (the un-named one)

because the former had the appropriate config for the pm as well, so please remove `Bootstrap Node.js` and keep the original `setup-node`, in the same position.

containers/self-hosted-runner/README.md
christian-calabrese: `uv` is not an Azure CLI runtime dependency; it is required by mise’s `pipx:azure-cli` backend to install Azure CLI. Mise does not bootstrap `uv` or `pipx` itself, and an isolated installation fails when neither is available. Installing Azure CLI through Microsoft’s OS packages would avoid `uv`, but would move it outside the mise-managed image baseline; the Conda backend currently provides only Azure CLI 2.74.0, below the pinned 2.87.0.

.github/actions/node-setup/action.yaml
christian-calabrese: Thanks for clarifying. Done: the `Bootstrap Node.js` step has been removed and the original `actions/setup-node` invocation remains as the only Node setup step. Package-manager cache resolution now runs after Corepack because the built-in pnpm/yarn cache handling invokes the package manager before Corepack exposes its shims on this runner.

.github/actions/node-setup/action.yaml
christian-calabrese: Fixed: `actions/setup-node` is now invoked only once. The duplicate bootstrap step was removed.

.nx/version-plans/version-plan-1785931489400.md
```diff
@@ -0,0 +1,7 @@
+---
+self-hosted-runner: patch
+---
+
+Make mise user directories writable and reject incomplete upstream runner
```
krusty93: question:
> reject incomplete upstream runner images

What does it mean?

.nx/version-plans/version-plan-1785931489400.md
```diff
@@ -0,0 +1,7 @@
...
+images. Pre-install a common cloud and archive tool baseline through mise while
```
krusty93:
```suggestion
images. Install `aws-cli`, `az-cli` and `zip` through mise.
```

.nx/version-plans/version-plan-1785931489400.md
```diff
@@ -0,0 +1,7 @@
+---
```
krusty93, line 1: nitpick: I'm not sure this file should be formatted this way, it makes revision harder

.nx/version-plans/version-plan-1785931489400.md
```diff
@@ -0,0 +1,7 @@
...
+self-hosted-runner: patch
+---
+
+Make mise user directories writable and reject incomplete upstream runner
+images. Pre-install a common cloud and archive tool baseline through mise while
+leaving language runtimes and Terraform to workflow setup actions by default.
```
krusty93:
```suggestion
```

containers/self-hosted-runner/Dockerfile
```diff
@@ -1,8 +1,8 @@
 
-# === Stage 1: Use base Runner and install tools ===
-ARG RUNNER_DIGEST
+# === Stage 1: Install runner dependencies and mise ===
+ARG RUNNER_DIGEST=sha256:3f06bf70688f9d314884d5bde3e7403eaea257d9a5120a635f4b7788d3411bf2
```
krusty93: thought: Setting the runner digest as `ARG` is not meaningful, nor used. To me it can be removed and simplified.

2nd thought: we should also import [this workflow](https://github.com/pagopa/github-self-hosted-runner-azure/blob/main/.github/workflows/check-release.yml) (made by us btw)

containers/self-hosted-runner/mise.toml
```diff
@@ -0,0 +1,9 @@
+[tools]
```
krusty93, line 1: question: why those are set at this level only? the repo needs aws and az cli too - we had them in repository's devcontainer

containers/self-hosted-runner/Dockerfile
```diff
@@ -11,55 +11,65 @@ ARG MISE_SHA256=961b1fcc78830e861ab887abd19d9b961478bcf252e37881fdd61c81388308d4
...
+    mise trust mise.toml; \
+    mise install; \
+    mise ls --current
+
+RUN set -eux; \
+    cd /home/runner; \
+    openssl version; \
+    mise --version; \
+    corepack --version; \
+    mise exec -- aws --version; \
```
krusty93, line 59: question: are those needed as `ls --current` runs before?

containers/self-hosted-runner/Dockerfile
```diff
@@ -11,55 +11,65 @@ ARG MISE_SHA256=961b1fcc78830e861ab887abd19d9b961478bcf252e37881fdd61c81388308d4
...
+    mise exec -- az version; \
+    mise exec -- uv --version; \
+    mise exec -- zip -v | head -n 2
+
+RUN set -eux; \
```
krusty93, line 64: question: what's the goal here?

containers/self-hosted-runner/Dockerfile
christian-calabrese: It is used in the very next line

`FROM ghcr.io/actions/actions-runner${RUNNER_DIGEST:+@${RUNNER_DIGEST}} AS mise`

But i agree we can integrate that workflow to keep it updated

containers/self-hosted-runner/mise.toml
christian-calabrese: They are already present in the repository `mise.toml` file. This is a way to pre-install in the docker image some common packages that are always required.

.nx/version-plans/version-plan-1785931489400.md
christian-calabrese: Good point. I kept the required Nx front matter, but moved the description to a single line to make the diff easier to review.

containers/self-hosted-runner/Dockerfile
christian-calabrese: This checks the contract we rely on from the upstream Actions runner image: `run.sh` must be present to start the runner, and the embedded Node.js runtime must be available for JavaScript actions. Failing during the image build is preferable to discovering an incomplete or incompatible base image when a job starts.

.nx/version-plans/version-plan-1785931489400.md
christian-calabrese: Removed. The version plan now only describes the tools added to the image.

.nx/version-plans/version-plan-1785931489400.md
christian-calabrese: Agreed, this is clearer. I updated the version plan accordingly.

.nx/version-plans/version-plan-1785931489400.md
christian-calabrese: It referred to the build-time checks for the upstream runner files (`run.sh` and the embedded Node.js runtime), but that is an implementation detail rather than a useful release note. I removed it from the version plan.

containers/self-hosted-runner/Dockerfile
christian-calabrese: Yes. `mise ls --current` confirms that mise selected the expected versions, while these commands verify that the installed binaries are actually executable and that their runtime dependencies are intact. This makes the image build fail before publishing a broken tool installation.

containers/self-hosted-runner/Dockerfile
christian-calabrese: I revisited this and simplified it. The Dockerfile now uses an explicit `ghcr.io/actions/actions-runner:2.335.0@sha256:...` reference, so I removed the `RUNNER_DIGEST` argument and label. I did not import the linked workflow because image releases are now handled by Nx through our Docker plugin; I removed the legacy self-hosted runner release workflow instead. Renovate already supports version-plus-digest Docker references, and a local dry-run correctly detected the `2.335.0` → `2.336.0` update, including the new digest and target branch. The Docker build currently rejects `2.336.0` because its `run.sh` and embedded Node.js runtime are empty, which is also why having Renovate open a validated PR is safer than silently building `latest`.

containers/self-hosted-runner/Dockerfile
krusty93: Thanks, as I expected. to me it's ok to leave it here as it's already been written, but as the base image provides that runtime I suspect this check could be removed

containers/self-hosted-runner/mise.toml
krusty93: so why duplicate here aws and az cli? Asking because we must keep versions in sync

christian-calabrese: This mise.toml file in the docker container has been introduced after [this question](https://github.com/pagopa/dx/pull/2048#discussion_r3727317954) by @lucacavallaro 

Keeping the packages in both files allows to:
- pre-install basic tools in the docker image to reduce install time in CI/CD workflows
- keep local mise installation compliant with all the required tools

Do you think we should add more or remove any?

krusty93: What are the effects to other repos if we update the version here only? Keep in mind az cli gets a new version one or two times per month

Also, we use few of [these extensions](https://learn.microsoft.com/en-us/cli/azure/azure-cli-extensions-list?view=azure-cli-latest). I think they must be added here too

christian-calabrese: If other repos require a newer version of the az cli, they can create a `mise.toml `file specifying an updated (or latest) version

However, we can always install the latest versions so that the latest available is built into the docker image every time we rebuild it

krusty93: I prefer this approach. I think ICs will use brew to always have the latest versions of these tools

lucacavallaro: @krusty93 But we aren't configuring `mise` for local development, only for CI.

We need to have `az-cli` and `aws-cli` stable and predictable for CI, so why should we align with our consumers' usage on their machines?

krusty93: TLDR: We took the decision to remove az and aws clis from here because restoring cache is faster than we thought (8s)

containers/self-hosted-runner/README.md
```diff
@@ -34,10 +46,39 @@ The Docker image includes an entrypoint script (`entrypoint.sh`) that is respons
...
+so every workflow using the DX runner receives the same safe default.
+
 ## Testing
 
 The image includes basic tests to verify the installation of key tools:
 
-- Node.js (`node --version`)
-- npm (`npm --version`)
 - mise (`mise --version`)
+- AWS CLI (`aws --version`)
```
krusty93:
```suggestion
```

containers/self-hosted-runner/README.md
```diff
@@ -34,10 +46,39 @@ The Docker image includes an entrypoint script (`entrypoint.sh`) that is respons
...
+- Azure CLI (`az version`)
```
krusty93:
```suggestion
```

lucacavallaro (approved):

krusty93 (approved):

PR #2050, opened 2026-08-05
Fix module.register() deprecation warning in CLI

mamu0:
The `dx` CLI preload module (apps/cli/src/adapters/azure-monitor/instrumentation.ts) used `module.register()` from node:module to install the import-in-the-middle ESM hook before any HTTP client libraries are loaded. Since Node.js marked this API as deprecated (DEP0205), every CLI invocation printed:

```
  (node:XXXXX) [DEP0205] DeprecationWarning: `module.register()` is deprecated.
  Use `module.registerHooks()` instead.
```

import-in-the-middle@3.3.2 ships a synchronous alternative based on `module.registerHooks()` (register-hooks.mjs), which avoids the deprecated API and the message-channel round-trip entirely.
It's only supported on Node.js versions where a related core bug (nodejs/node#59929) is fixed: `>=22.22.3, >=24.11.1, >=25.1.0, >=26.0.0`.
Since apps/cli declares `engines.node: >=22.0.0`, a wider range, this change adds a dual path:

- Use the synchronous `registerHooks()` API when `supportsSyncHooks()` reports the running Node.js correctly supports it.
- Otherwise fall back to the existing `module.register()` + `createAddHookMessageChannel`/`waitForAllMessagesAcknowledged` approach, so older supported Node.js versions keep working exactly as before (with the same deprecation warning, since Node itself hasn't removed the old API).

Also adds an ambient module declaration for `import-in-the-middle/register-hooks.mjs`, since the package ships
`register-hooks.d.ts` rather than `register-hooks.d.mts`, which TypeScript's `node16`/`nodenext` module resolution doesn't associate with the `.mjs` file.

Verified manually on Node.js 26 (sync path, no warning) and by temporarily forcing the fallback branch (old behavior, warning still present as expected on unsupported Node versions).

Resolves: CES-2232

### DIFF
--- a/apps/cli/src/adapters/azure-monitor/import-in-the-middle-register-hooks.d.ts
+++ b/apps/cli/src/adapters/azure-monitor/import-in-the-middle-register-hooks.d.ts
@@ -1,16 +1,11 @@
 /**
  * Ambient module declaration for `import-in-the-middle/register-hooks.mjs`.
  *
- * The package ships `register-hooks.d.ts` (not `.d.mts`), which TypeScript's
- * `node16`/`nodenext` module resolution does not associate with the `.mjs`
- * implementation file, so the subpath import would otherwise be untyped.
+ * The package ships `register-hooks.d.ts`, not `.d.mts`, so TS's node16/nodenext
+ * resolution can't match it to the `.mjs` file. Re-exporting from the real
+ * declaration keeps this in sync with upstream instead of hand-copying types.
  */
-declare module "import-in-the-middle/register-hooks.mjs" {
-  export type RegisterHooksOptions = {
-    exclude?: (RegExp | string)[];
-    include?: (RegExp | string)[];
-  };
 
-  export function register(options?: RegisterHooksOptions): void;
-  export function supportsSyncHooks(): boolean;
+declare module "import-in-the-middle/register-hooks.mjs" {
+  export * from "import-in-the-middle/register-hooks.d.ts";
 }
--- a/apps/cli/src/adapters/azure-monitor/import-in-the-middle-register-hooks.d.ts
+++ b/apps/cli/src/adapters/azure-monitor/import-in-the-middle-register-hooks.d.ts
@@ -1,0 +1,16 @@
+/**
+ * Ambient module declaration for `import-in-the-middle/register-hooks.mjs`.
+ *
+ * The package ships `register-hooks.d.ts` (not `.d.mts`), which TypeScript's
+ * `node16`/`nodenext` module resolution does not associate with the `.mjs`
+ * implementation file, so the subpath import would otherwise be untyped.
+ */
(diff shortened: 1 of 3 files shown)


gunzip:
related issue upstream: https://github.com/open-telemetry/opentelemetry-js/issues/4933#issuecomment-4593525039

apps/cli/src/adapters/azure-monitor/import-in-the-middle-register-hooks.d.ts
```diff
@@ -0,0 +1,16 @@
...
+ */
+declare module "import-in-the-middle/register-hooks.mjs" {
+  export type RegisterHooksOptions = {
+    exclude?: (RegExp | string)[];
+    include?: (RegExp | string)[];
+  };
+
+  export function register(options?: RegisterHooksOptions): void;
+  export function supportsSyncHooks(): boolean;
+}
```
lucacavallaro, line 11: **question:** why does this declaration file is needed?

lucacavallaro (approved):

apps/cli/src/adapters/azure-monitor/import-in-the-middle-register-hooks.d.ts
gunzip: isn't the reason pointed by the comment?

mamu0: `import-in-the-middle` ships `register-hooks.d.ts` (not `.d.mts`) for its `register-hooks.mjs` file, and TS's `node16`/`nodenext` resolution doesn't match declaration files across extensions, so the subpath import ends up untyped (TS7016). This ambient declaration is the minimal fix instead of using `any`/`@ts-expect-error`. It's a known upstream packaging gap (one of the commenters in the linked OTel issue hit the same "missing types" problem).

lucacavallaro: Can the types be imported from the `d.ts` file (using `import type` ...) instead of being re-declared?

My main concern here is that our static checks won't catch any future API changes in `import-in-the-middle`, because we are overriding its typings. This will lead to runtime errors for our consumers.

mamu0: Good catch, you're right, we can.
Updated to `export * from "import-in-the-middle/register-hooks.d.ts"` instead of restating the signatures ([b6aab47](https://github.com/pagopa/dx/pull/2050/commits/b6aab47e0cac3e04c539f6a611e657ac28e4c66c)). This re-exports the real upstream types, so if their API changes, typechecking will fail instead of silently drifting. Verified this by temporarily breaking the API usage locally, the build correctly errors.

lucacavallaro (approved):

PR #2054, opened 2026-08-06
Move integration and e2e tf tests to Nx Terraform Plugin

lucacavallaro:
- Move Terraform integration and e2e tests into dedicated Nx target configurations
- Keep the default `tf-test` target scoped to unit and contract tests
- Remove redundant module scripts and workspace `targetOverrides`
- Add plugin contract coverage, the generated bundle, and a patch version plan

### DIFF
--- a/packages/nx-terraform-plugin/src/project.ts
+++ b/packages/nx-terraform-plugin/src/project.ts
@@ -103,6 +103,31 @@ const getPublishTarget = (
   }
 };
 
+const getTestTarget = (initTargetName: string): TargetConfiguration => ({
+  cache: true,
+  command: `terraform test`,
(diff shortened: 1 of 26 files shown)


christian-calabrese (approved):

packages/nx-terraform-plugin/src/project.ts
```diff
@@ -103,6 +103,31 @@ const getPublishTarget = (
   }
 };
 
+const getTestTarget = (initTargetName: string): TargetConfiguration => ({
```
gunzip, line 106: suggestion: add expected behavior (ie. file names) and docs to plugin README

PR #2065, opened 2026-08-10
Use $/ self-repository syntax for internal action references

kin0992:
## Why

This repo (`pagopa/dx`) referenced its own actions and reusable workflows in two inconsistent ways:
- The full external form `pagopa/dx/...@main`, pinned to the mutable `main` branch.
- The workspace-relative `./...` form, which requires a prior checkout to resolve.

GitHub's new `$/` self-repository syntax resolves same-repo `uses:` references to the exact commit currently running, with no checkout required (Actions runner >= 2.336.0).

**Important caveat (found during review):** `$/` always resolves against the **top-level caller's repository**, even through nested composite actions and reusable workflows. `pagopa/dx` is a shared workflow/action provider consumed directly by many other pagopa repositories (`io-wallet`, `io-infra`, `io-services-cms`, `io-sign`, etc.) via `uses: pagopa/dx/...@main`. So `$/` is only safe inside files that can **never** be a `uses:` target from another repository.

## Benefits

- Same-repo-only internal automation (cron jobs, subrepo pushes, e2e tests, drift detection, changesets release trigger, etc.) no longer depends on tracking the mutable `main` branch and stays in sync with the exact commit/SHA a caller is pinned to.
- Removes the remote-fetch overhead of resolving "external" references to our own repo for those internal-only files.
- Drops an unnecessary checkout step in one integration test workflow.
- Does **not** break cross-repo consumption of `pagopa/dx`'s public reusable workflows and actions.


kin0992:
I also double checked with [zizmor](https://github.com/zizmorcore/zizmor) locally to verify the unpinned issues are gone (and they are)

.github/instructions/github-workflows.instructions.md
```diff
@@ -21,7 +21,9 @@ Always prefer explicit inputs over implicit environment variables for workflow a
 
 Maintain a clear distinction between internal and external dependencies.
 
-- **Internal DX Actions**: It is acceptable to use branch names like `main` or semantic tags for internal PagoPA DX actions during development or when they are part of the same lifecycle as the caller.
+- **Internal-only actions/workflows (same repository, never called externally)**: Reference sibling actions and reusable workflows with the self-repository `$/` syntax (e.g. `uses: $/actions/csp-login`, `uses: $/.github/workflows/static_analysis.yaml`), without a `@ref` suffix. `$/` always resolves to the exact commit currently running, so it never needs to track `main` and automatically stays in sync even when the caller is pinned to a tag or full commit SHA. Requires Actions runner >= 2.336.0.
```
krusty93: suggestion: `Requires Actions runner >= 2.336.0.` can be removed as we're already running it

.github/workflows/_validate-docker-e2e-cosmos-networkaccess.yaml
krusty93, line 26: question: Is it possible to remove this checkout action?

krusty93 (approved):

.github/workflows/_validate-docker-e2e-cosmos-networkaccess.yaml
kin0992: This checkout isn't related to the `$/` self-repo syntax change — it's the primary checkout of this repo, needed so `docker-build-push` has access to `infra/modules/azure_cosmos_account/tests/apps/network_access/Dockerfile` as its build context (that action doesn't do its own checkout). It's different from the redundant secondary `dx/` checkout removed elsewhere in this PR, so it should stay.

